PR #44, opened 2023-01-18
DOCS DOCS DOCS DOCS

Gankra:


Gankra:
(the build failure is fixed by #43)

README.md
```diff
@@ -3,59 +3,332 @@
...
+
+That's gonna do a whole bunch of stuff you might not have expected (but I left off the --execute flag from `cargo-release` so you won't actually break anything if you really did just copy paste that 😇). Read the rest of the docs to know what it does!
+
+
+
+
+
+# Installation
+
+## Build From Source With Cargo
```
ashleygwilliams, line 62: do we want to also show the installers you created?

README.md
```diff
@@ -3,59 +3,332 @@
...
+cargo dist generate-ci github
+```
+
+See the next section ("Usage (CI)") for how the github workflow is triggered and what it does.
+
+
+
+## Configuring Installers
+
+If you would like to generate (currently bad) installer scripts, you can pass `--installer` flags
```
ashleygwilliams, line 115: maybe not "currently bad" but something.. nicer than that?

README.md
```diff
@@ -3,59 +3,332 @@
...
+to either `init` or `generate-ci`:
+
+```sh
+cargo dist init --ci=github --installer=github-shell --installer=github-powershell
+```
+
+This will result in `installer.sh` and `installer.ps1` being generated which fetch from a Github Release™️ and copy the binaries to `HOME/.cargo/bin/` on the assumption that this is on your path. The scripts are currently brittle and won't properly tell cargo about the installation. As such they are currently only really appropriate for bootstrapping CI scripts. This will be improved in the future.
```
ashleygwilliams, line 122: we may want to try to caveat this for folks who may not care if cargo knows about the installation e.g. people who build rust tools for other ecosystems..

README.md
```diff
@@ -3,59 +3,332 @@
...
+* x86_64-unknown-linux-gnu
+* x86_64-apple-darwin
+
+(In The future arm64 counterparts and linux-musl will probably join this, but unfortunately we currently don't support cross-compilation.)
+
+If you would like to manually specify the targets, you can do this with `--target=...` which can be passed any number of times. If this flag is passed then the defaults will be cleared. 
+
+Other commands like `cargo dist build` (bare `cargo dist`) will always default to only using the current host target, and may need more manual target specification. This is handled automatically if you're using dist's generated CI scripts.
+
+**cargo-dist does not currently support specifying additional targets based on different `--features` or anything else, this will change in the future.**
```
ashleygwilliams, line 140: maybe link to the issue and encourage folks to comment if they have ideas?

ashleygwilliams (approved):
some small comments but in general, i dig it. i do think we should move this out of the readme asap and into a doc because it is long and i'm worried people may not get to the great content at the bottom :)

README.md
Gankra: I don't want to advise those right now, because they put cargo in a pseudo-corrupt state where `cargo dist` runs but `cargo uninstall cargo-dist` says "there's no such thing" which is uhhh bad for anything but CI.

README.md
Gankra: Hmm could you elaborate?

Gankra: thinking about this more, it's not clear if they're useful for that *either* because then `.cargo/bin` is unlikely to be on the PATH.

README.md
Gankra: Changed my mind and added them, and actually demoted the two options that Literally Won't Work Because It's Not Published On Crates.io.

README.md
ashleygwilliams: oh dang right, hrm. yeah ok. you're right here.

PR #54, opened 2023-01-27
provide manual instructions for releasing

### LINKED ISSUE #51: provide manual instructions in addition to cargo-release
i get a gnarly error cargo-installing cargo-release, to be on the safe side i think including manual docs would be a good call

Gankra:
fixes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,25 @@ cargo release 0.1.0
 
 That's gonna do a whole bunch of stuff you might not have expected (but I left off the --execute flag from `cargo-release` so you won't actually break anything if you really did just copy paste that 😇). Read the rest of the docs to know what it does!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,25 @@ cargo release 0.1.0
 
 That's gonna do a whole bunch of stuff you might not have expected (but I left off the --execute flag from `cargo-release` so you won't actually break anything if you really did just copy paste that 😇). Read the rest of the docs to know what it does!
 
+Or if you don't want to use cargo-release, you can emulate its behaviour manually:
```
ashleygwilliams, line 59: can we show it this way first and then be like- hey cargo-release does some of this if you'd like to use it!

PR #56, opened 2023-01-31
Full github changelog support

Gankra:
...not properly tested


cargo-dist/src/cli.rs
```diff
@@ -84,6 +85,15 @@ pub enum Commands {
     Manifest(ManifestArgs),
 }
 
+#[derive(Args)]
+pub struct ReleaseNotesArgs {
+    /// Get release notes for a specific version.
+    ///
+    /// Otherwise the app's current version will be used.
+    #[clap(long)]
+    pub version: Version,
```
ashleygwilliams, line 94: i wonder if we will get people who don't want to use semver.... (this is not a change request, just random thought)

cargo-dist/src/lib.rs
```diff
@@ -629,23 +643,35 @@ fn gather_work(cfg: &Config) -> Result<DistGraph> {
                         .filter(|s| !s.contains("windows"))
                         .cloned()
                         .collect::<Vec<_>>();
+                    let app_version = format!("v{version}");
+                    let hint = format!("# WARNING: this installer is experimental\ncurl --proto '=https' --tlsv1.2 -L -sSf {repo_url}/releases/download/{app_version}/installer.sh | sh");
```
ashleygwilliams, line 647: warning is solid

cargo-dist/src/lib.rs
```diff
@@ -666,6 +692,129 @@ fn gather_work(cfg: &Config) -> Result<DistGraph> {
...
+        }
+
+        if !installers.is_empty() {
+            changelog_body.push_str("## Install\n\n");
+            for installer in installers {
+                let install_hint;
+                let description;
+
+                match &installer.bundle {
+                    BundleStyle::Installer(InstallerImpl::GithubShell(info)) => {
```
ashleygwilliams, line 752: this function is Very Large, it may be valuable to break it up into smaller chunks

cargo-dist/src/lib.rs
```diff
@@ -1462,6 +1615,16 @@ fn compute_package_info(
     workspace_root: &Utf8Path,
     package: &PackageMetadata,
 ) -> Result<PackageInfo> {
+    // Is there a better way to get the path to sniff?
+    // Should we spider more than just package_root and workspace_root?
+    // Should we more carefully prevent grabbing LICENSES from both dirs?
+    // Should we not spider the workspace root for README since Cargo has a proper field for this?
```
ashleygwilliams, line 1621: i would say we should probably use what we can from cargo- 

cargo also has the license information... we could check if it matches? unclear if that's valuable or not-

cargo-dist/tests/snapshots/cli_tests__manifest.snap
```diff
@@ -127,17 +127,21 @@ stdout:
...
           "target_triples": [
             "x86_64-pc-windows-msvc"
           ],
-          "install_hint": "irm 'https://github.com/axodotdev/cargo-dist/releases/download/v1.0.0-FAKEVERSION/installer.ps1' | iex"
+          "install_hint": "# WARNING: this installer is experimental\nirm 'https://github.com/axodotdev/cargo-dist/releases/download/v1.0.0-FAKEVERSION/installer.ps1' | iex",
+          "description": "Install prebuilt binaries via powershell script"
         }
-      ]
+      ],
+      "changelog_title": "CHANGELOG-CENSORED",
```
ashleygwilliams, line 143: lol

ashleygwilliams (commented):
i think this looks good. i would have an easier time reviewing if that big function was broken up into smaller ones :)

cargo-dist/src/lib.rs
Gankra: we do use what we can but license-file is one value and most repos have two (license-file is an escape-hatch for "my license is something weird and custom SPDX has no idea about")

cargo-dist/src/cli.rs
Gankra: within cargo itself you have no option but to use Semver

PR #112, opened 2023-02-10
Remove reference to Rover CLI when installer downloads musl binary

shadaj:


ashleygwilliams (approved):

ashleygwilliams:
thank you so much! will merge when CI passes

Issue #111, opened 2023-02-08
[RFC] multi-tenant release workflow


Gankra:
The core of cargo-dist supports the notion of workspaces with multiple binaries -- it will find them all and create individual "Releases" for each one ([see "concepts" for important details on what that means](https://github.com/axodotdev/cargo-dist/#concepts), but it's basically "here's a new version of an application" which contains all the various artifacts like platform-specific binaries).

However as I prototyped out more features I increasingly avoided trying to think about how that semantically works and support for this is consequently very broken.

To try to create a coherent vision/implementation of these multi-tenant workspaces, I would like to introduce a new concept at the top of the hierarchy: The *Announcement*. Whereas before all the Releases were in some sense top-level and independent, they will now be logically part of one unified Announcement, which contains the Releases. This will not change the structure of dist-manifest.json -- the entire object is already The Announcement, I'm just formalizing the notion.

# Things That Should Be Unified

The importance of formalizing this notion is that it's reified in things like Github Releases, since we're currently containing one Github Release containing many cargo-dist Releases. I think this is fine and desirable, but it means some values need to be centralized/shared:

* Every Release needs to agree on the repository url and git-tag is being used for The Announcement to create correct artifact-download URLs
  * currently they all use their own `repository` value and assume the tag is `v{version}` where `version` is the one set in their own Cargo.toml. If any two packages disagree on these values and they're part of the same Announcement one of them is going to get broken URLs!
* A Github Release has only one title and body, so even if there's individual changelogs/tables for each binary they need to be stitched together into some unified form
  * #69 
  * I might still emit "partial" release notes for each binary in the dist-manifest.json, in case another tool wants its own custom stitching
* ci-kind (e.g. "github ci") really needs to be central
* toolchain versions ("use this version of cargo-dist", "use this version of rustc") basically have to be central
* compiler flags varying across packages are a headache, you really want to build the workspace in one shot [for consistency](https://crates.io/crates/cargo-hakari) 
  * Cargo nods to this with [the fact that \[profile\] can only be configured at the root of the workspace](https://doc.rust-lang.org/cargo/reference/profiles.html)
* target-triples being unified is *simpler* but this one will probably have to be weakened eventually
  * does Cargo.toml already have a way to say a binary is platform-specific? I've never seen it in the reference!
  
Notably anything that we agree *must* be centralized should only be sourced from `[workspace.metadata.dist]` and not `[package.metadata.dist]` (maybe we'll still parse them just to emit a warning).

[...]

Issue #103, opened 2023-02-04
Detect binary on init


CAD97:
Obviously most of the complexity handled by cargo-dist is around building artifacts for a binary release, and as such it's not really needed for library crates.

However, cargo-dist is so convenient that I just went to apply it to a library crate before realizing my error. It would be nice if cargo-dist would detect that a crate/workspace has no binary target(s) and show a warning/error, probably pointing to cargo-release and/or [cargo-bins/release-pr](https://github.com/cargo-bins/release-pr) as covering the library part of releases.

Issue #97, opened 2023-02-02
Precompute flags and whatnot for build_cargo_target in tasks.rs


Gankra:
Ideally we should know all the settings for the build before going to do it!

The most egregious exception to this is the RUSTFLAGS computing logic, but it might also be nice to have the task graph be able to report for the full CLI invocation before going to actually run it!

Issue #93, opened 2023-02-02
Allow to select which binaries should be build & published


DCNick3:
I have a workspace (https://github.com/DCNick3/shin) with a binary that is used mostly for write-once experiments (called `junk`), so I don't want it in my builds & releases.

I am already excluding it from `cargo-release` with `release = false` in `[package.metadata.release]` in `Cargo.toml`. It would be nice to have such an option in `cargo-dist` too

Gankra:
Is #79 an acceptable solution for your usecase?

DCNick3:
No. I currently have some git dependencies (because stuff is not merged yet), so I can't really publish to crates.io

A hint would be nice though

kujeger:
There's `default-members`in `[workspace]` which determines which members/packages are built or run by default when using e.g. `cargo build` -- perhaps cargo-dist could use the same behaviour?

PR #115, opened 2023-02-11
Massively rework task computation/execution while introducing configs

Gankra:
This commit on its own is still missing some major pieces, will iterate on the PR.

TODO:

* [x] reimplement changelog code (currently commented out)
* [x] properly compute `announce_tag` (currently dummied)
* [x] rework installer code to use `announce_tag` and other newly centrally computed values
  * [x] make the installers know there are multiple binaries in zips (pass a list in) 
* [x] reimplement the ci script to use the new `--artifacts` flags and global `announce_*` fields
  * [x] make it actually work (push to CI 1000 times) 
  * [x] make generate-ci (and similar commands) ignore the fact that there's no coherent announcement-tag!
* [x] reimplement `init` to check/initialize the new configs
* [x] docs
* [x] test test test test
* [x] add new `announce_*` fields to dist-manifest.json
* [x] properly test multi-tenant stuff on something like [evil-workspace](https://github.com/Gankra/evil-workspace)
   * [x] now do it again now that you've done a bunch of "fixes" 

FIXES:

* fixes https://github.com/axodotdev/cargo-dist/issues/28
* fixes https://github.com/axodotdev/cargo-dist/issues/34
* fixes https://github.com/axodotdev/cargo-dist/issues/58
* fixes https://github.com/axodotdev/cargo-dist/issues/59
* fixes https://github.com/axodotdev/cargo-dist/issues/61
* fixes https://github.com/axodotdev/cargo-dist/issues/63
* fixes https://github.com/axodotdev/cargo-dist/issues/68
* fixes https://github.com/axodotdev/cargo-dist/issues/69
* fixes https://github.com/axodotdev/cargo-dist/issues/79
* fixes https://github.com/axodotdev/cargo-dist/issues/91
* fixes https://github.com/axodotdev/cargo-dist/issues/93
* fixes https://github.com/axodotdev/cargo-dist/issues/97
* fixes https://github.com/axodotdev/cargo-dist/issues/103
* fixes https://github.com/axodotdev/cargo-dist/issues/111 
* fixes https://github.com/axodotdev/cargo-dist/issues/123


Gankra:
This is a good chunk of #111 with drive-by fixes of like 10 other issues just because it's all tangled up in configs and proper data flow.

cargo-dist/src/tasks.rs
```diff
@@ -49,50 +92,221 @@ pub const CPU_ARM64: &str = "arm64";
...
 /// Contents of METADATA_DIST in Cargo.toml files
-#[derive(Deserialize, Debug)]
-pub struct DistMetadata {}
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub struct DistMetadata {
+    /// The intended version of cargo-dist to build with. (normal Cargo SemVer syntax)
+    ///
+    /// When generating full tasks graphs (such as CI scripts) we will pick this version.
+    ///
+    /// FIXME: Should we produce a warning if running locally with a different version? In theory
```
niklaswimmer, line 111: I think warning would be a good idea, for two reasons:

1. When someone runs `cargo-dist` locally, then most likely they want to test what the CI will do. But if the CI uses a different version then the user locally, they may see different behavior which could cause confusion. A heads up that the local build may not produce the same result because versions differ would solve that.
2. This version here essentially functions as a lock file, ensuring CI builds are reproducible. The problem is that users have to remember to bump this version when a new `cargo-dist` release is published. We could therefore use that opportunity to inform the user that their CI is still using an older version and that they can update.

Gankra: I largely agree, it's mostly a question of working out the right UX for it

Gankra:
Just the CI to rewrite now... (and a fuckton of testing/docs)

Gankra:
IT WORKS!!!!

Cargo.toml
```diff
@@ -8,9 +8,21 @@ members = [
...
+installers = ["github-shell", "github-powershell"]
+# Target platforms to build apps for (Rust target-triple syntax)
+targets = ["x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows-msvc"]
+
+
+# The profile that 'cargo dist' will build with
 [profile.dist]
 inherits = "release"
-debug = true
-split-debuginfo = "packed"
```
niklaswimmer, line 15: Is this intended to change in the new version?

cargo-dist/src/ci.yml
```diff
@@ -0,0 +1,110 @@
...
+# choose what to do with partial successes and avoids spamming
+# anyone with notifications before the release is actually ready.
+name: Release
+
+permissions:
+  contents: write
+
+# This task will run whenever you push a git tag that looks like
+# a version number. We just look for `v` followed by at least one number
+# and then whatever. so `v1`, `v1.0.0`, and `v1.0.0-prerelease` all work.
```
niklaswimmer, line 22: Those docs seem to be out of date, considering that the CI runs on every Git tag now (because the version may be prefixed with an application identifier).

On another thought, maybe we should try to just use a better regex that can catch those, so we do not run on *every* tag?

cargo-dist/src/tasks.rs
```diff
@@ -49,50 +92,221 @@ pub const CPU_ARM64: &str = "arm64";
...
+    /// Can be "off", "packed", or "unpacked".
+    ///
+    /// If "packed" then we expect a pdb/dsym/dwp artifact.
+    pub split_debuginfo: Option<String>,
+}
 
 /// Global config for commands
 #[derive(Debug)]
 pub struct Config {
     /// Whether we'll actually run builds (if false we'll still generate installers)
```
niklaswimmer, line 249: Just a quick note that those docs are now outdated.

niklaswimmer (commented):
Very nice! :rocket: 

I only marked some places where I noticed that the docs are out of date, so that they are not forgotten. Codewise, well I did not have the pleasure to actually write much Rust yet so I think I am not really in the position to note anything there. I looked through it and it seems good - it probably has something to say that I was able to understand it all without much experience xD

I also noticed that there are a bunch of new fixmes and todos in the codebase now. I think we should create new issues for those and not block this PR on them.

I hope it is ok if I leave my thoughts, just want to help out where I can. If desired, I can take another look when the docs are updated :)

Gankra:
Added a rework of installer.ps1 that passes down a table of supported targets/artifacts that the script just defers to, rather than having the script build up a bunch of pieces in a more adhoc way. This makes it now correctly handle zips with multiple binaries, and sets up a decent framework for work like #125.

I'm not... *super* happy with the introduction of ExecutableZipFragment and make_executable_zip_for_variant, they are super clearly hacks. The problem is that instead of always building a maximal graph and then selecting down a subset, I currently rely on just building the minimal graph for correctness. This creates a problem for installers which essentially *do* want to know about the Full Graph to get the list of fetchable executable-zips.

The hacky compromise is to just have the installers temporarily construct those artifacts and throw them away after pull out the bits they want.

I'm also unhappy with all the places I am implicitly spreading the concept that a binary is always copied to the root of an executable-zip, but this PR is ridiculous enough as it is.

I now need to apply a similar treatment to the shell script, which thankfully already has the values fed down, but still needs a rewrite of the script itself (this is my weakest area, I suck at bash, and bash sucks at Structures).

Cargo.toml
Gankra: Yeah I'm backing off this stuff for now because there are weird unreliable interactions with people's existing profiles. Will be revisited later. :/

cargo-dist/src/ci.yml
Gankra: Thanks for the reminder. I think it should be `*-?v[0-9]+*` (the filter format isn't regexes, it's [weird pseudo-globs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet))

cargo-dist/src/tasks.rs
Gankra: good catch

Gankra:
...and THAT's the last major piece!

time to test test test doc doc doc

Issue #125, opened 2023-02-14
Allow specifying executable bundle style


localcc:
This is useful when the app only has one executable, so there's no need to pack it into an archive, or you want to choose another bundle option, e.g. `tar.xz` for windows.

This could be accomplished by adding an `--exe-bundle-style` cli argument to the tool with different bundle style options.

niklaswimmer:
Just wanted to note that this is related to #17

I like the option of no bundling at all though, would be useful for simple tools.

Gankra:
definitely a dupe of #17, where I'll write some more notes

PR #126, opened 2023-02-14
Make exe bundle style configurable

localcc:
This solves the issue #125 and allows exe bundle style to be configurable by user.

### DIFF
--- a/cargo-dist/src/ci.rs
+++ b/cargo-dist/src/ci.rs
@@ -8,7 +8,7 @@ use camino::Utf8PathBuf;
 use miette::{IntoDiagnostic, WrapErr};
 use tracing::warn;
 
-use crate::InstallerStyle;
+use crate::{BundleStyle, CompressionImpl, InstallerStyle};
 
 const GITHUB_CI_TRIGGER: &str = r###"# CI that:
 #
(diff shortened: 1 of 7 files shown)


Gankra:
I'm afraid this is going to massively conflict with #115, so I'm not gonna be able to review/accept this before that lands. :(

niklaswimmer:
Unfortunately this change seems to break installers, because they are hardcoded to the archive extension (based on the platform). I think we would need to pass the extension into the installer generation as well, and using platform specific values only as a fallback if no extension was explicitly set.

localcc:
I'm not really seeing how this breaks installers as the option is only processed in the executables loop, which has a match clause saying that an installer variant is unreachable

niklaswimmer:
What I meant by this is that the installer script itself won't work (under some circumstances). The script will always try to download a tar archive if it is executed on a Unix host and a zip when on Windows, even if cargo-dist generates zips for all platforms.

Gankra:
yeah I need to finish my refactor which computes way more information centrally and feeds it down to the installers. at that point we can then start spoon-feeding the installers more precise details like per-platform-extensions

Gankra:
The latest commit in #115 now has the WIP refactor of the installers wired up, with a bunch of the end-user configurability disabled (i.e. no more specifying version), as the schema for URLs/paths is in flux and can't yet be guaranteed across versions reliably. The installers now basically just plug together exactly the values fed into them.

However they are missing one large thing that needs to be handled for #111 -- if a package has two binaries defined, that means we'll make a zip with two binaries (whose names won't match the app name or zip name). As such, we need installer.rs to feed down a list of binaries. Similar logic could be used to feed down a map of target-triple => zip-extension/exe-extension, presumably?

Gankra:
sorry about the bad timing of my refactor! gonna close this for now.

PR #138, opened 2023-02-16
try to use no-prerelease release notes if available

### LINKED ISSUE #129: release note parsing: prerelease sugar
If you have release notes for 0.0.3 and publish 0.0.3-prerelease, we should pick up the 0.0.3 notes on the assumption that they're applicable. This is pretty straight forward, if the first key lookup fails with parse_changelog we just do a second one with the prerelease suffix removed.

It's really annoying having to bump the title in the changelog when cutting prereleases!

Gankra:
fixes #129

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -1397,7 +1397,25 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
 
         // Try to lookup this version in the changelog
         let version_string = format!("{}", announcing_version);
-        let Some(release_notes) = changelogs.get(&*version_string) else {
+        let release_notes = if let Some(release_notes) = changelogs.get(&*version_string) {
+            release_notes
+        } else if !announcing_version.pre.is_empty() {
+            // If we're a prerelease, try looking for this version without the prerelease suffix
+            // on the basis that we shouldn't be backfilling prereleases after the real release,
+            // so those are probably also our own WIP release notes.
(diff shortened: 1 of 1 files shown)


ashleygwilliams (approved):
this seems good to me behavior wise and i trust that the code does what it says it does. i do think it's worth making it explicit in docs that we do this and if you want the prerelease notes to be diff (at CF for example we would put QA instructions in the release notes) all you ahve to do is *have* pre-release notes and we'll prefer those to the non pre release ones

cargo-dist/src/tasks.rs
```diff
@@ -1397,7 +1397,25 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
 
         // Try to lookup this version in the changelog
         let version_string = format!("{}", announcing_version);
-        let Some(release_notes) = changelogs.get(&*version_string) else {
+        let release_notes = if let Some(release_notes) = changelogs.get(&*version_string) {
+            release_notes
+        } else if !announcing_version.pre.is_empty() {
```
ashleygwilliams, line 1402: ok actually after staring at this code i think this check should go *after* looking up if there are pre-release specific notes, if that makes sense?

Gankra: it does -- first we check the entire version-string blindly (1.0.0-prerelease) and then IF that fails AND we notice there was a suffix THEN we take that branch (checking bare 1.0.0).

PR #140, opened 2023-02-16
purge github- prefix from installers

Gankra:


README.md
```diff
@@ -157,9 +157,11 @@ cargo dist init
 If you want to just (re)generate the ci scripts, you can do:
 
 ```sh
-cargo dist generate-ci github
+cargo dist generate-ci
 ```
 
+(Note that for this precise invocation you should have `ci = ["github"]` set in your Cargo.toml)
```
ashleygwilliams, line 163: if folks don't have ci = github in their cargo.toml what happens

README.md
```diff
@@ -170,7 +172,7 @@ If you would like to generate (still under development) installer scripts, you c
 to either `init` or `generate-ci`:
 
 ```sh
-cargo dist init --ci=github --installer=github-shell --installer=github-powershell
+cargo dist init --ci=github --installer=shell --installer=powershell
```
ashleygwilliams, line 175: does clap let you do --installer=shell,powershell to mean the same as this

README.md
```diff
@@ -389,10 +389,10 @@ cargo-dist is self-hosting, so just follow the usual [usage instructions](#usage
 The CI is (re)generated with:
 
 ```
-cargo dist generate-ci github --installer=github-shell --installer=github-powershell
+cargo dist generate-ci
 ```
 
-**Including the installers is very important, as all the CI scripts cargo-dist generates for *other* projects will bootstrap dist with those installers.**
+**NOTE: if you want to update the version of cargo-dist used in the CI, you need to update cargo-dist-version in the root Cargo.toml! I keep forgetting this and should add a feature to help this this! Yeah! Looking at you, Future Me Who Still Hasn't Done This!**
```
ashleygwilliams, line 395: this note feels like it might be worth filing an issue for tackling in the next release and/or documenting more explicitly

cargo-dist/src/installer.sh
```diff
@@ -6,7 +6,7 @@
 # except according to those terms.
 
 # This is just a little script that can be downloaded from the internet to
-# install $APP_NAME. It downloads the tarball from GitHub releases,
+# install $APP_NAME. It downloads the tarball from ARTIFACT_DOWNLOAD_URL,
```
ashleygwilliams, line 9: chef's kiss

cargo-dist/tests/snapshots/cli_tests__manifest.snap
```diff
@@ -35,10 +35,10 @@ stdout:
           "description": "Install prebuilt binaries via powershell script"
         },
         {
-          "name": "cargo-dist-v1.0.0-FAKEVERSION-x86_64-pc-windows-msvc.zip",
```
ashleygwilliams, line 38: these snapshot diffs kill me (does --arifact all always do alphabetical?)

ashleygwilliams (approved):
lgtm, i have some comments but nothing blocking

cargo-dist/tests/snapshots/cli_tests__manifest.snap
Gankra: I keep adding more sorting to make this stuff more stable but I think in this case it's using the order in the Cargo.toml (where previously it was using the order in the --target flags I was passing).

README.md
Gankra: https://github.com/axodotdev/cargo-dist/issues/141

README.md
Gankra: Added more text basically saying "hmm maybe the cli flags for this are Bad now"

README.md
Gankra: (I also explicitly don't add short forms, again because stuff like `curl -LsSf` fills me with SADNESS even though it's nice that it's compact for installer hints)

Gankra: (fight me)

Issue #131, opened 2023-02-16
whitespace-clean outputs


Gankra:
the generated CI/installers contain a little bit of trailing/mixed whitespace and that's gonna annoy people. I thought i flushed it out but at least the CI still has some :(

Gankra:
maaaaybe fixed on tip, so prerelease-03 should have the fixes? (had a bit of a caching struggle so ended up just pushing)

Gankra:
This is still clearly an issue, I have no idea why running `cargo dist generate-ci` on my machine produces weird mixed line-endings. The inputs are all LF-only, afaict? Unless this is a fucked up "my own tools are hiding line-ending stuff from me" problem.

ashleygwilliams:
@Gankra what editor are you using? (i think it may be a "your tools are hiding stuff" situation)

Gankra:
I used vscode to change all the line endings to LF-only, and inspecting with sublime-text seems to confirm this worked, even downloading raw files from github... but maybe there's something I'm missing?

Gankra:
it's poooossible this is actually fixed on tip but some more boostraps need to settle things?

Gankra:
OH OH THIS WOULD BE SO EVIL

IT'S POSSIBLE THAT GITHUB CI HAS `git config --global core.autocrlf true` SET AND SO THE OFFICIAL BUILDS HAVE THE TEMPLATES REWRITTEN TO HAVE CRLF ENDINGS AAAAAAAA

Issue #74, opened 2023-02-01
cross-compilation support


Gankra:
This is punted on in the current version, pending some evaluation of the options. There are three options I know:

* [cross-rs](https://github.com/cross-rs/cross) (established, based on docker)
* [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) (new kid on the block, [has some blocking issues](https://github.com/rust-cross/cargo-zigbuild/blob/5de2f13fbc1d92a9d732f26af1ca15c0688e2472/README.md#caveats))
* Do It Our Dang Selves With Rustup (please god no)

This is a blocker for musl and aarch64 (arm64) support.

Note that [taiki-e/upload-rust-binary-action](https://github.com/taiki-e/upload-rust-binary-action) *does* handle this and in my experiments whatever it does Just Works (great work from taiki-e as always).

Gankra:
One of my big concerns about cross-compilation is to what extent certain Fancy Things cannot be cross-compiled.

Apple certainly does not want you cross-compiling from linux to macos. I'm not sure if support for code signing (#21) or certain installers is necessarily non-crossable. Certainly some are *harder* to cross. Certainly there are... Ways Around The Rules. But I don't want to get people in trouble or have them ship busted things!

reubenmiller:
Just some of my recent experience/pain regarding cross compiling for Linux arm* and x86_64 targets for an open source project that I'm involved with.

Basically we have to build for linux environments on arm >= v6, arm64 and x86_64, for both gnu and musl releases. We were using [cargo-cross](https://github.com/cross-rs/cross) however still had some issues to build for all the targets and the output was dependant on your build machine.

Then I saw a great blog post about [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) and was amazed with how effort it was to cross compile for all of our targets, especially since the other developers have a mix between Windows (using WSL2) and MacOS (M1/arm64 and amd64) setups.

Below details some of the details with the projects:

**Experience cargo-cross**
* Slow builds as building on MacOS M1 (arm64/aarch64) and the current default is to use amd64 images by default (though there are unofficial arm64 available...there is also a "zigbuild" variant, from what I found digging around)
* Building for musl on some targets just didn't work (mostly because of our project's dependency to the common [ring](https://github.com/briansmith/ring)...the number of times that dependency has caused me build headaches is just astounding)

**Experience with cargo-zigbuild**
* Builds are quick
* Does not rely on docker
[...]

PR #143, opened 2023-02-18
feat: Add aarch64-apple-darwin support

### LINKED ISSUE #133: cross compile aarch64-apple-darwin by default
because this is made dead simple by apple (we need only add the target), i think we should include compiling to aarch64-apple-darwin by default, ahead of any larger cross compilation story (which will require more configuration and tooling)

this plus #82/#100  will bring us to target parity with a lot of folks looking to migrate to using cargo-dist

frol:
*Resolves #133.* cc @ashleygwilliams @Gankra 

cargo-dist is now compiled for and also can be used to package other packages for aarch64-apple-darwin (M1/M2 chips)

I have tested this patch in a personal fork, here is the release: https://github.com/frol/cargo-dist/releases. Also, I have [manually patched the CI scripts](https://github.com/FroVolod/near-social/commit/814dd57812ff3b62d70736fd5693512fc43038b1) for my near-social CLI, and it [worked](https://github.com/FroVolod/near-social/releases/tag/v0.2.2) like a charm (see that it now has aarch64-apple-darwin binary artifact), so I went ahead with creating this patch.

P.S. The `release.yml` file in main branch has a mix of CRLF and LF line endings. My guess is that git checks out `ci.yml` template with CRLF on Windows and then cargo-dist code generates some multiline strings with `\n` (`push_github_artifacts_matrix_entry`)


frol:
@Gankra I was also trying to understand the reasoning behind `distribute_targets_to_runners`. Having parallel compilation (through having a matrix group per build) is faster, and easier to reason about and debug, so there must be something behind this complication; what is that?

Gankra:
Doing a quick skim before I look at this properly tomorrow morning: this seems like a reasonable minimal approach, although I had been mentally planning for this to look more like:

* teach cargo-dist to detect that you had "extra" tools like rustup/cross/strip/lipo/whatever (once in gather work, similar to makefile configure steps)
* if we find rustup, handle getting the toolchain installed for the user (I don't think we even need to bother asking for something as unobtrusive as a toolchain)

This would leave the CI untouched while also making *local* (non-CI) macos builds work better as well. Although the other platforms would however need to pull up a proper "cross-compile" core (#74).

frol:
> this seems like a reasonable minimal approach

Yes, this was exactly the goal of this PR, and I feel it solves the immediate pain point (#133) well without introducing complexity or diverging in a direction that will be hard to solve when a full-scale solution for cross-compilation will be developed.

M1/M2 Macbooks are quite popular among developers these days, and I would either need to come up with some clever logic in the installer that would pull x86_64 binary when aarch64 artifact is missing, or I could just build my CLI for aarch64-apple-darwin; the latter turned out to be quite a straightforward thing to do.

Gankra:
I'm adding this as a tentative candidate for 0.0.3 which I'm going to cut in a couple days. If I find the bandwidth to do the Ideal solution I'll opt for that. Otherwise I'll land this to help people Get Stuff Done (well, assuming the implementation works).

My main concern is that the Ideal solution will restore the CI to its current state, so anyone who makes changes to their CI will have an extra round of otherwise-avoidable CI churn with this quick fix.

frol:
@Gankra Sorry, I don't see where you are heading. The changes to CI script are minimal and the current diff is weird since the original file had a mix of CRLF line endings and now it is completely CRLF (I feel it is a separate issue to address). [Change the diff view to skip the whitespace changes](https://github.com/axodotdev/cargo-dist/pull/143/files?diff=split&w=1) to see the actual change.

```
- dist-args: --artifacts=local --target=x86_64-apple-darwin
+ dist-args: --artifacts=local --target=aarch64-apple-darwin --target=x86_64-apple-darwin
```

This is what essentially got changed, and I feel it is unavoidable unless you suggest cargo-dist should guess all the supported platforms and cross-compile to everything supported by default.

Gankra:
Ugh I guess I didn't properly fix #131 :(

But yes there will be some churn on CI as we refine things. There's definitely going to be a huge churn for 0.0.3 already, and that cost is already accepted. The question is after 0.0.3, when we ostensibly introduce the Ideal Solution. 

* if we don't accept this change then when we introduce the Ideal Solution, any users who adopt it will get that one line diff you noted -- not as likely to conflict, and easy to fix if it does.
* if we do accept this change then users will essentially have to revert all the CI changes this PR makes, notably removing "targets" lines and the lines that configure rustup. this will be a lot more likely to conflict, and might mess up anyone who thought they could/should rely on "targets" being in the matrix.

It's not a big deal at all, I'm just noting the only real downside of accepting this PR.

(I have been pondering making the first task actually generating the matrix so there would be *no* churn even for the Ideal solution but that's a bigger change and makes it harder for users to configure anything...)

frol:
I believe the lines that setup rustup will inevitably get changed with the introduction of cross-compilation unless you want to go Zig-lang route and package cross-compilation toolkits for all the platforms or install Docker for users.

I may think of some way to avoid `targets` in the matrix if you don't want to give people a chance to depend on it, but so far I can only think of ugly hacks

Gankra:
We discussed this and 0.0.3 is too big already for us to be poking this feature too. I'm pushing this back to 0.0.4 which *optimistically* is "next week".

Gankra:
I implemented my version in https://github.com/axodotdev/cargo-dist/pull/172, which was indeed thankfully straight-forward with my refactor done :)

This prerelease has it: https://github.com/axodotdev/cargo-dist/releases/tag/v0.0.4-prerelease.2

Issue #145, opened 2023-02-19
Account for the 'Unreleased' section when parsing release notes


niklaswimmer:
#138 added support for using the no-prerelease release notes for prereleases. It would be nice if we account for the 'Unreleased' section as well when parsing, using it as another fallback if no no-prerelease entry is available.

Similar to #142 the title of the release notes should be adjusted to match the actually published version.

Gankra:
Just confirming this isn't a thing parse-changelog supports, and would require us to do some significant upstream work or doing some of our own markdown parsing?

niklaswimmer:
I have not tried it tbh, will do later. But I hope it does because their [header regex](https://docs.rs/parse-changelog/latest/parse_changelog/#versions) includes it.

niklaswimmer:
Can confirm, works like a charm. 

```rust
changelogs["Unreleased"].version == "Unreleased"
```

Gankra:
oh ha! I didn't scroll the regex far enough :)

PR #146, opened 2023-02-19
Append prerelease suffix to sugar'd release titles

### LINKED ISSUE #142: append a "this is a prerelease" suffix to sugar'd release title
#138 is nice but now it's really hard to see what prerelease it is. If we hit the sugar path we should add a suffix like "(prerelease04)" to make it clear!

niklaswimmer:
Fixes #142 by appending the prerelease suffix to sugar'd release titles.

An example: if you try to create a release for `1.0.0-pre1` and we detect a changelog entry of the form `## v1.0.0` the title of the release notes will end up as "v1.0.0-pre1".

Another example: you try `3.0.0-prerelease5`, we detect `## Version 3.0.0 (not yet released!)`, the title will end up "Version 3.0.0-prerelease5 (not yet released!)"

I am unsure whether we would want to strip the "(not yet released!)" part as well. When looking at the GitHub releases page they seem kinda confusing - a released version that says it is not released :exploding_head:

The problem is that it is not clear how we should detect such an indication. A simple solution would be to just strip everything after the first opening parenthesis. We could also try to remove only the parenthesis itself, leaving anything that comes after it intact, but this leaves us with the problem of deciding which parenthesizes to delete if there are multiple (or just yeet them all I guess).

On another note, it may actually be cleaner to use the 'Unreleased' header in the changelog together with #145 once that is implemented for achieving this painless and fast prereleasing.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -98,6 +98,8 @@ Release notes are now temporarily simplified for reliability:
 
 * If you are publishing `1.2.0-prerelease` and we don't find that in your RELEASES/CHANGELOG file, we will now also look for bare `1.2.0` (stripping the prerelease/build portions), on the assumption that these are the WIP release notes for the version you're prereleasing. This lets you iterate on a version without having to churn headings every time you want to cut a prerelease (we recommend including a parenthetical indicating the version is not yet released).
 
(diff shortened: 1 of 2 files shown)


Gankra:
Oh this is a nice approach! Happy to land with rustfmt fixed.

(I'm just going to change the parenthetical in the README to "(under development)")

PR #150, opened 2023-02-21
final default dist profile

### LINKED ISSUE #118: settle on default profile
At least on macos, even though we're generating split-debuginfo the binary is still stuffed with a bunch of debuginfo like function names that ideally would only be in the dSYM! I need to audit the flags we're picking for their actual platform-specific semantics and maybe add post-process steps like `strip`.

(Current profile is inherits="release", debug=true, and split-debuginfo="packed". iirc adding the builtin `strip` flag to this caused at least one platform to clobber its own split-debuginfo but I need to go back and take proper notes. If that persists we can just manually invoke strip.)

Gankra:
fixes #118


niklaswimmer:
Maybe leave a note somewhere in the code with a link to #118, since it now contains all the explanations and discussions of possible options?

Issue #141, opened 2023-02-16
More convenient "update cargo-dist" workflow


Gankra:
Because cargo-dist-version is baked into your Cargo.toml, you need to remember to update that field to actually get the CI to use a new version. Installing the new version locally on your machine is insufficent. I keep messing this up and forgetting the field!

Gankra:
oh this was implemented in #227

PR #157, opened 2023-02-23
emit an error if generate-ci disagrees with cargo-dist-version

Gankra:


Gankra:
this fixes the worst aspects of #141

Issue #85, opened 2023-02-02
Support for multiple Windows architectures in installer.ps1


AustinWise:
While x86_64 is the most popular architecture for Windows at this time, ARM64 is slowly gaining relevance and x86 remains useful for certain scenarios.

I propose adding the ability for the installer script to automatically pick the best executable from those available when installing. The `dist-manifest.json` file already contains all the information needed to pick the best version. Building and uploading the different version for Windows architectures is a separate issue.

I have some ideas on how to approach this based on writing a [similar script](https://github.com/AustinWise/smeagol/blob/master/site/install.ps1). There are [some details](https://github.com/AustinWise/smeagol/blob/master/PowershellNotes.md) about this issues I ran into.

### Level One: download artifacts that match the current native OS architecture

There are a number of different ways to figure out the current OS native architecture:

* The [`RuntimeInformation`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.runtimeinformation.osarchitecture). This is the easiest way to get the architecture. On the latest versions of Windows and Power Shell Core, it tells the truth. On earlier versions, it might return X64 on ARM64. Given that ARM64 Windows can run X64 in emulation, it's probably fine to use this. Added in .NET 4.7.1.
* [`Environment.Is64BitOperatingSystem`](https://learn.microsoft.com/en-us/dotnet/api/system.environment.is64bitoperatingsystem) - added .NET 4 (so available in PowerShell 5). This can be used as a fallback to `RuntimeInformation` to detect 32-bit x86 Windows running older operating systems.

There are some native APIs that are probably not worth the complexity of calling from PowerShell:
 
* [`IsWow64Process2`](https://learn.microsoft.com/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2). Returns accurate information about the system. Available starting in Windows 10 v1511.
* [`GetNativeSystemInfo`](https://learn.microsoft.com/windows/win32/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo)
  Added in Windows XP. Lies about the native system when running under x86 emulation on ARM64 and in ARM64EC processes.

### Level Two: downloading the best artifact, taking into account emulation.

If there is not an artifact that matches the current system's OS, it's possible an artifact of a different architecture may be runnable under emulation. For example, Windows 11 ARM64 systems can run X86, X64, and sometimes ARM32 executables.

[...]

Issue #17, opened 2022-10-24
Allow user to pick packaging scheme


Gankra:
Currently we just use `.zip` when targetting windows and `.tar.xz` everywhere else, but we should allow the user to configure this in the settings.

https://github.com/taiki-e/upload-rust-binary-action has some examples for inspiration.

Gankra:
Note that this information will need to be threaded through to installers and generate-ci. Currently those things slightly hardcode zip/tar extensions. Unclear if installers should get these choices baked in, or if they should be written to investigate the github release's contents and more intelligently detect what to do (this may be necessary if we want installer scripts to meaningfully support version-to-install-as-an-argument-to-the-script.

Gankra:
Now that #115 is implemented we're better postured to do this, some notes on Things You'll Need To Do:

* Config: 
  * Add field(s) to the [DistMetadata config](https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/tasks.rs#L106) (this adds it to Cargo.tomls)
  * Merge field(s) in [merge_workspace_config](https://github.com/axodotdev/cargo-dist/blob/7c77b2908359a03000717df5b767d26e44b99602/cargo-dist/src/tasks.rs#L211) so packages can override
  * ~~Add an equivalent CLI flag and merge it too~~ probably not appropriate here!
  * Read field(s) in [make_executable_zip_for_variant](https://github.com/axodotdev/cargo-dist/blob/7c77b2908359a03000717df5b767d26e44b99602/cargo-dist/src/tasks.rs#L966-L974)

* Build:
  * In priniciple the actual build steps already support all the formats, nothing to do! (test this...)

* Teach the installers to handle these formats
  * Clear the assertions [(1)](https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/installer.rs#L54) [(2)](https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/installer.rs#L136) from installer.rs that complain you haven't implemented the feature
  * Teach installer.sh to handle the new format (maybe only needs zip support?)
    * [Handle the value of _zip_ext here](https://github.com/axodotdev/cargo-dist/blob/7c77b2908359a03000717df5b767d26e44b99602/cargo-dist/src/installer.sh#L74)
  * The installer.ps1 to handle the new format (presumably it will choke on any tar.whatever?)
[...]

PR #168, opened 2023-02-26
PowerShell installer: use manifest to select files to download.

AustinWise:
This does not implement actually building something other than X64.

This checks for architects that matches the current operating system's architecture. It falls back to X64 if none are available.

Related issue: #85

### DIFF
--- a/cargo-dist/src/installer.ps1
+++ b/cargo-dist/src/installer.ps1
@@ -35,12 +35,58 @@ function Install-Binary($install_args) {
   $ErrorActionPreference = $old_erroractionpreference
 }
 
+function Get-TargetTriple() {
+  try {
+    # NOTE: this might return X64 on ARM64 Windows, which is OK since emulation is available.
+    # It works correctly starting in PowerShell Core 7.3 and Windows PowerShell in Win 11 22H2.
+    # Ideally this would just be
+    #   [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture
+    # but that gets a type from the wrong assembly on Windows PowerShell (i.e. not Core)
+    $a = [System.Reflection.Assembly]::LoadWithPartialName("System.Runtime.InteropServices.RuntimeInformation")
+    $t = $a.GetType("System.Runtime.InteropServices.RuntimeInformation")
+    $p = $t.GetProperty("OSArchitecture")
+    # Possible OSArchitecture Values: https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.architecture
+    # Rust supported platforms: https://doc.rust-lang.org/stable/rustc/platform-support.html
+    switch ($p.GetValue($null).ToString())
+    {
+      "X86" { return "i686-pc-windows-msvc" }
+      "X64" { return "x86_64-pc-windows-msvc" }
+      "Arm" { return "thumbv7a-pc-windows-msvc" }
+      "Arm64" { return "aarch64-pc-windows-msvc" }
+    }
+  } catch {
+    # The above was added in .NET 4.7.1, so Windows PowerShell in versions of Windows
+    # prior to Windows 10 v1709 may not have this API.
+    Write-Verbose "Get-TargetTriple: Exception when trying to determine OS architecture."
+    Write-Verbose $_
+  }
+
+  # This is available in .NET 4.0. We already checked for PS 5, which requires .NET 4.5.
+  Write-Verbose("Get-TargetTriple: falling back to Is64BitOperatingSystem.")
+  if ([System.Environment]::Is64BitOperatingSystem) {
+    return "x86_64-pc-windows-msvc"
+  } else {
+    return "i686-pc-windows-msvc"
+  }
+}
+
 function Download($download_url, $platforms) {
   # FIXME: make this something we lookup based on the current machine
-  $arch = "x86_64-pc-windows-msvc"
+  $arch = Get-TargetTriple
+
+  if (-not $platforms.ContainsKey($arch)) {
+    # X64 is well-supported, including in emulation on ARM64
+    Write-Verbose "$arch is not availablem falling back to X64"
+    $arch = "x86_64-pc-windows-msvc"
+  }
+
+  if (-not $platforms.ContainsKey($arch)) {
+    # should not be possible, as currently we always produce X64 binaries.
+    $platforms_json = ConvertTo-Json $platforms
+    throw "ERROR: could not find binaries for this platform. Last platform tried: $arch platform info: $platforms_json"
+  }
 
   # Lookup what we expect this platform to look like
-  # FIXME: produce a nice error if $arch isn't found (or do fallback guesses?)
   $info = $platforms[$arch]
   $zip_ext = $info["zip_ext"]
   $bin_names = $info["bins"]


AustinWise:
@Gankra What do you think of using the `dist-manfiest.json` to select which files to download and install?

With regards to supporting archives other than zip (#17), the script stops executing after download if it encounters something other than a .zip file. The download path is printed, so the user can find the downloaded file easily.

I'm leaving this as draft until I remove the part in `install.rs` that generates the hard-coded asset list.

Gankra:
Hey! Sorry for the delay, was focusing on the last bits of cutting 0.0.3.

> What do you think of using the dist-manfiest.json to select which files to download and install?

I am... *mixed* on it. In principle it's a good idea, but also in principle it's unnecessary and pushing complexity from the extremely maintainable/safe Rust code to the more hairy/dangerous shell scripts. Do you have a usecase in mind that makes it important/useful to do this kind of late-bound analysis?

AustinWise:
> Do you have a usecase in mind that makes it important/useful to do this kind of late-bound analysis?

Not really. There would certainly be fewer things that could go wrong if the list of available artifacts and assets is hard-coded into the script.

Do you still think it is a good idea to have the script detect the OS architecture? I can roll-back my changes to use dist-manifest.json and instead include the architecture in the platform info injected by cargo-dist.

Gankra:
yeah #85 would be great still! Sorry I didn't realize you had actually impl'd the target detection stuff, I'll review that part now!

Gankra:
Out of morbid curiousity I checked out what powershell on macos does (yeah it's a thing) and then hacked up a version that should work on all platforms:

```ps1
function Get-TargetTriple() {
  try {
    Write-Verbose "test"
    # NOTE: this might return X64 on ARM64 Windows, which is OK since emulation is available.
    # It works correctly starting in PowerShell Core 7.3 and Windows PowerShell in Win 11 22H2.
    # Ideally this would just be
    #   [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture
    # but that gets a type from the wrong assembly on Windows PowerShell (i.e. not Core)

    # Possible OSPlatform Values: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform?source=recommendations&view=net-7.0
    # Possible OSArchitecture Values: https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.architecture
    # Rust supported platforms: https://doc.rust-lang.org/stable/rustc/platform-support.html

    $a = [System.Reflection.Assembly]::LoadWithPartialName("System.Runtime.InteropServices.RuntimeInformation")
    $t = $a.GetType("System.Runtime.InteropServices.RuntimeInformation")
    $raw_arch = $t.GetProperty("OSArchitecture")
    $arch = $null
    switch ($raw_arch.GetValue($null).ToString())
    {
      "X86" { $arch = "i686" }
      "X64" { $arch = "x86_64" }
      "Arm" { $arch = "thumbv7a" }
      "Arm64" { $arch = "aarch64" }
    }
    Write-Verbose "arch is $arch"

    $windows_os = [System.Runtime.InteropServices.OSPlatform]::Create("WINDOWS")
    $mac_os = [System.Runtime.InteropServices.OSPlatform]::Create("OSX")
    $linux_os = [System.Runtime.InteropServices.OSPlatform]::Create("FREEBSD")
    $freebsd_os = [System.Runtime.InteropServices.OSPlatform]::Create("LINUX")
    $os = $null
    if ([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($windows_os)) {
      $os = "pc-windows-msvc"
    } elseif ([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($mac_os)) {
      $os = "apple-darwin"
    } elseif ([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($linux_os)) {
      $os = "unknown-linux-gnu"
    } elseif ([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($freebsd_os)) {
      $os = "unknown-freebsd"
    }
    Write-Verbose "os is $os"

    if ($os && $arch) {
      return "$arch-$os"
    } else {
      Write-Verbose "failed to get os and arch"
    }
  } catch {
    # The above was added in .NET 4.7.1, so Windows PowerShell in versions of Windows
    # prior to Windows 10 v1709 may not have this API.
    Write-Verbose "Get-TargetTriple: Exception when trying to determine OS architecture."
    Write-Verbose $_
  }

  # This is available in .NET 4.0. We already checked for PS 5, which requires .NET 4.5.
  Write-Verbose("Get-TargetTriple: falling back to Is64BitOperatingSystem.")
  if ([System.Environment]::Is64BitOperatingSystem) {
    return "x86_64-pc-windows-msvc"
  } else {
    return "i686-pc-windows-msvc"
  }
}
```

(probably this can be cleaned up)

```
PS /Users/ABeingessner/dev/tmp> ./tmp.ps1
VERBOSE: test
VERBOSE: System.Runtime.InteropServices.RuntimeInformation
VERBOSE: arch is x86_64
VERBOSE: os is apple-darwin
x86_64-apple-darwin
```

cargo-dist/src/installer.ps1
```diff
@@ -16,53 +16,135 @@ param (
...
-function Download($download_url, $platforms) {
-  # FIXME: make this something we lookup based on the current machine
-  $arch = "x86_64-pc-windows-msvc"
+function Get-TargetTriple() {
+  try {
+    # NOTE: this might return X64 on ARM64 Windows, which is OK since emulation is available.
+    # It works correctly starting in PowerShell Core 7.3 and Windows PowerShell in Win 11 22H2.
+    # Ideally this would just be
+    #   [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture
+    # but that gets a type from the wrong assembly on Windows PowerShell (i.e. not Core)
```
Gankra, line 43: could you elaborate on this? (is this a Known Issue with an article?)

AustinWise: I wrote down a little bit more about the problem I encountered here:

https://github.com/AustinWise/smeagol/blob/master/PowershellNotes.md#footnote-why-we-need-to-use-reflection-to-access-runtimeinformation

I'll do a little more research to see if there is a better solution. Reflection is a bit verbose.

Gankra: Oh wow, this is awesome, thanks! I'm willing to tolerate the weird reflection with this kind of documentation backing it.

Gankra:
Doing a bit more research into aarch64 windows, and I'm seeing a lot of notes of "eh fuck it, just use x64, arm64 windows has a rosetta-like emulation layer". Do you think we should add the same kind of installer fallback that was added in #165?

AustinWise:
> Out of morbid curiousity I checked out what powershell on macos does (yeah it's a thing) and then hacked up a version that should work on all platforms:

It is a fair question whether or not the PowerShell script should support non-Windows platforms, given that it can run on several platforms now. Espeially since it appears you could [set PowerShell as your logon shell](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pwsh?view=powershell-7.3#-login---l) on Unix-like systems!

One argument against that is you would have to implement all the logic to support Unix-like systems twice: once in the shell script and once in the PowerShell script. To avoid that maintenance burden, the PowerShell script could detect when it is being run on non-Windows:

```powershell
if ([System.Environment]::OSVersion.Platform -ne [System.PlatformID]::Win32NT) {
    throw "This script is only supported on Windows. Please run this instead:\n curl --proto '=https' --tlsv1.2 -LsSf https://github.com/{{REPO}}/releases/download/{{VERSION}}/{{CRATE}}-{{VERSION}-installer.sh | sh";
}
```

`cargo-dist` could fill in all the right values to make that *just work*. It could even run the `curl | sh` pipeline behind the scenes, but that might be excessively clever.

AustinWise:
> Doing a bit more research into arm64 windows, and I'm seeing a lot of notes of "eh fuck it, just use x64; arm64 windows has a rosetta-like emulation layer". Do you think we should add the same kind of installer fallback that was added in #165?

After fiddling with my Windows Dev Kit 2023 ARM64 I have come to a similar conclusion for Windows. It makes sense to try to download platform-specific binaries if they are available, but otherwise fallback to X64. That is the WIP state of this PR and I have updated #85 to reflect that this simple approach should be good enough.

AustinWise:
This is now much-simplified from the original PR and now ready for more review.

Gankra:
taking today off, will review on thursday, but i wanted to leave you some
loose notes on installer stuff:

* powershell installers will need to grow support for most of the linux-y
functionality anyway once we make zip style configurable. this doesn't need
to happen in your PR, i'm just keeping an eye out for the future. also i am
told that windows 10 now ships with tar out of the box which makes that
simple, although i need to look into how new that is.

* i'm going to refactor the shell/powershell code on the rust side to have
them share more code. this can happen before or after your pr, whichever
you prefer. i intentionally kept them separate in case they needed to
wildly diverge but i'm now seeing they are only converging more. this work
will also power future feature work on npm-based installers.

* i'm going to add some more general logic for "fallback" platforms so it's
easy for us to plug in more compatibility cases like x64 on arm64
(mac/windows), or x86 on x64 (windows only at this point..?)

cargo-dist/src/installer.ps1
```diff
@@ -35,12 +35,58 @@ function Install-Binary($install_args) {
...
 function Download($download_url, $platforms) {
   # FIXME: make this something we lookup based on the current machine
-  $arch = "x86_64-pc-windows-msvc"
+  $arch = Get-TargetTriple
+
+  if (-not $platforms.ContainsKey($arch)) {
+    # X64 is well-supported, including in emulation on ARM64
+    Write-Verbose "$arch is not availablem falling back to X64"
+    $arch = "x86_64-pc-windows-msvc"
+  }
```
Gankra, line 81: I'm probably going to hoist this logic into the rust (as with rosetta fallback) but this is legit for now.

Gankra (approved):

Gankra:
Thanks so much! I'll get started on the installer refactor now so these things are fed more info.

cargo-dist/src/installer.ps1
AustinWise: Makes sense. You could even have cross-OS fallbacks, i.e. running Linux binaries on FreeBSD system that have emulation enabled.

AustinWise:
Thanks for merging. I really like the direction your are taking the install script by embedding the data about assets into them. This totally sidesteps of the thorny issue of having to parse JSON in `dist-manifest.json` in a cross-platform way.

PR #172, opened 2023-02-28
Use rustup to install the appropriate macOS target when doing mac->mac crosses

### LINKED ISSUE #133: cross compile aarch64-apple-darwin by default

Gankra:
Currently semi-WIP, I need to actually dig out my macbook to test this lol

fixes #133

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ ci = ["github"]
 # The installers to generate for each app
 installers = ["shell", "powershell"]
 # Target platforms to build apps for (Rust target-triple syntax)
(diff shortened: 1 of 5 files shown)


Gankra:
Ah shoot, that last commit ("remove vestigial target hardcode") was a mistake in 0.0.3's refactor. 0.0.3 is fine since it's ~unobservable but it means we'll need to do an extra prerelease to bootstrap the fix into CI.

Issue #175, opened 2023-03-01
No way to keep Oranda on the last stable version (or stable link old versions)


Gankra:
I just cut 0.0.4-prerelease.2 and now https://opensource.axo.dev/cargo-dist/ is recommending/showing that over 0.0.3

This is... not what we want to happen, I think! This is more of an Oranda bug but I wanted to note it here in case we hack around it locally.

SaraVieira:
Right now I don't allow you to override the version in the oranda config but this may actually be a super good use case for that

ashleygwilliams:
i'd rather do something less manual... we should have some notion of stability channels and only promote new releases as stable. currently oranda is very much a "single version" tool, but now that we are doing the changelog stuff, that's changing and i think it will continue to move towards being multi-version mutli-channel and having a way to make a release as latest would be useful!

SaraVieira:
we can easily have a notion of stability by using the same API we are using for the changelog and fetching the first release that is not a pre release, if this is a good idea I can definitely integrate it in the changelog PR

ashleygwilliams:
you know that i'd recommend making it two PRs haha but yeah sounds good

SaraVieira:
Ill make a PR to that PR 

![image](https://user-images.githubusercontent.com/1051509/222193046-11b4a008-eeea-4d7d-9f30-201ab22a7943.png)

Gankra:
As a temporary workaround I've pinned the commit oranda looks at: https://github.com/axodotdev/cargo-dist/commit/29d4e3c891fc0aa77417958e79de21a635292387

So https://opensource.axo.dev/cargo-dist/ will show 0.0.3 and no longer update until we have a proper solution.

PR #176, opened 2023-03-01
docs: format --execute flag in docs

domoritz:


Gankra:
thanks!

Gankra:
(note that our hosted docs are currently frozen while we work on #175)

PR #206, opened 2023-03-23
Add `Swatinem/rust-cache@v2` to generated github CI

WaffleLapkin:
Since `cargo-dist` operates on binary crates, it seems to be reasonable to assume that dependency caching would be efficient — binary crates generally have quite a few dependencies which can be cached.

On the other hand caching is made less important/efficient by the fact that releases are generally a lot less frequent than, for example, per-commit/PR CI checks.

Thus I'm not entirely sure this has a lot of value, but I think it still should make releases a bit faster.

(P.S. sorry for not discussing in issues first, I didn't notice contributing instructing at first; anyway this change is small enough that I don't mind if my "work" is wasted here and this PR is not accepted)

### DIFF
--- a/cargo-dist/src/ci.yml
+++ b/cargo-dist/src/ci.yml
@@ -53,6 +53,8 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install Rust
         run: {{{{INSTALL_RUST}}}}
+      - name: Cache Dependencies
+        uses: Swatinem/rust-cache@v2
       - name: Install cargo-dist
         run: {{{{INSTALL_DIST_SH}}}}
       - id: create-release


Gankra:
oh god i just noticed how old this PR is, jesus sorry. i've been on a really deep yakshave and neglecting this repo (finally back).

WaffleLapkin:
> do you know much about the time/scope of the caching? like does this put you in a global github cache, or would it only cache across runs of the release ci itself?

As far as I can tell the cache is local to the repository (or maybe even a concrete job in the CI? it is not very clear...) and is keyed by rustc version, hashes of `Cargo.lock`s/`Cargo.toml`s, and some [other information](https://github.com/Swatinem/rust-cache#cache-details) (although there is some mechanism to re-use old cache after changing `Cargo.lock` to support partial updates).

Only *dependencies* are cached which is especially efficient with binary crates with a committed `Cargo.lock`. Caching workspace crates is inefficient (since they change often, I assume) and as such is not done.

Wrt time, it looks like caching is limited to 10 MB and once this threshold is reached oldest cache entries are evicted ([src](https://github.com/Swatinem/rust-cache#cache-limits-and-control)).

Does that answer your question?

panekj:
Cache artefacts are limited to repository and branches. Cache is only available to own branch and default branch. In PRs that's extended to base branch as well.
So cache in CI runs on `main` will be available only from `main`, cache in CI runs on `feat/cool-feature` will be available from `main` and `feat/cool-feature`.
In PR with branch `feat/cool-feature`, cache will be available from `main`, `feat/cool-feature` and `base-branch` from which `feat/cool-feature` was created. 
[Documentation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) 

> Wrt time, it looks like caching is limited to 10 MB

That sounds extremely low and also a typo, 10GB*

Gankra:
I'm really torn on this -- I don't think this gains us much *today* but it might in a future where we're supporting "i want to have an --all-features build and a --no-default features build" or something. That said if we just run those on the same machine (we already group arm mac and intel mac builds onto the same machine) then the cache isn't needed either...

I think I'm gonna opt to close this, but I'm keeping it in my back pocket as a Good Suggestion For Future Problems.

PR #224, opened 2023-04-20
fix(npm-installer): Update binary.js to respect suppressLog passed from run.js

frol:
Without this change, the argument passed from `run.js` is ignored:

https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/templates/npm/run.js#L4

FYI, `suppressLog` is a partially inverted in `binary-install` package: https://github.com/EverlastingBugstopper/binary-install/pull/23, but it is still better to pass `true` to avoid "... is already installed, skipping installation." message being printed on every CLI run.

### DIFF
--- a/cargo-dist/templates/npm/binary.js
+++ b/cargo-dist/templates/npm/binary.js
@@ -95,10 +95,9 @@ const getBinary = () => {
   return binary;
 };
 
-const install = () => {
+const install = (suppressLogs) => {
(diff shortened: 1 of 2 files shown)


Gankra:
oh geez i thought i merged this

Gankra (approved):

PR #228, opened 2023-04-25
feat: support keywords in npm installer

### LINKED ISSUE #220: pipe axoproject "keywords" into npm installer
Depends on https://github.com/axodotdev/axoproject/issues/16

The code already has the logic for outputting this field wired up, but it feeds in an dummy empty array. You need to pipe the field from axoproject to the npm installer, which will involve several steps/types because I pedantically have a few "stages" in cargo-dist where data must be explicitly forwarded (to force things to compute stuff more centrally).

* [get the axoproject value here](https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/tasks.rs#L963)
* [then handoff to the installer here](https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/tasks.rs#L1417)
* [final destination here where it should automatically get fed into package.json](https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/installer.rs#L252)

shadows-withal:
Closes #220. I hope I caught all the necessary places keywords needed to be shoehorned in? It looks right to me.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -613,10 +613,12 @@ pub struct Release {
     pub app_authors: Vec<String>,
     /// The license of the app
     pub app_license: Option<String>,
-    /// The license of the app
+    /// The URL to the app's source repository
     pub app_repository_url: Option<String>,
-    /// The license of the app
+    /// The URL to the app's homepage
     pub app_homepage_url: Option<String>,
+    /// A list of the app's keywords
+    pub app_keywords: Option<Vec<String>>,
     /// The version of the app
     pub version: Version,
     /// The unique id of the release (e.g. "my-app-v1.0.0")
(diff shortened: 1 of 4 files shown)


cargo-dist/src/tasks.rs
```diff
@@ -613,10 +613,12 @@ pub struct Release {
...
     /// The license of the app
     pub app_license: Option<String>,
-    /// The license of the app
+    /// The URL to the app's source repository
     pub app_repository_url: Option<String>,
-    /// The license of the app
+    /// The URL to the app's homepage
     pub app_homepage_url: Option<String>,
+    /// A list of the app's keywords
+    pub app_keywords: Option<Vec<String>>,
```
shadows-withal, line 621: I don't even think adding this to the `DistGraph` was necessary, but I figured it'd be good future-proofing in case we mess around with stuff outside of npm

Gankra: Yep this is perfect.

Gankra (approved):
perfect!

PR #229, opened 2023-04-27
feat: factor out fs calls into axoasset

### LINKED ISSUE #221: factor out more fs code to axoasset
cargo-dist predates [axoasset](https://github.com/axodotdev/axoasset), so we still have a crapload of fs code that probably "should" be in axoasset (or removed in favour of existing axoasset code).

some things that come to mind...

# code to copy/tar/zip files/dirs

https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/lib.rs#L445-L595

(Not sure if we want the ZipStyle/CompressionImpl stuff in axoasset, maybe it exposes just 4 entrypoints for the 4 choices?)

# code that loads a tomledit document

https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/tasks.rs#L2151-L2165

# code that creates/writes files based on templates

https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/installer.rs#L23-L30

https://github.com/axodotdev/cargo-dist/blob/4daf3da315032d73c2dcc263412d55c27cc1525d/cargo-dist/src/installer.rs#L102-L109

shadows-withal:
Culmination of the work that started with the 0.2.0 release of `axoasset`, which is to remove straight fs i/o calls from cargo-dist and put them in axoasset instead. Omits the templating things for now until we have a consensus on that.

Closes #221.

Needs testing, I haven't had enough time for that yet.

### DIFF
--- a/cargo-dist/Cargo.toml
+++ b/cargo-dist/Cargo.toml
@@ -36,10 +36,6 @@ thiserror = "1.0.35"
 tracing = { version = "0.1.36", features = ["log"] }
 serde = { version = "1.0.144", features = ["derive"] }
 cargo_metadata = "0.15.0"
-tar = "0.4.38"
-zip = "0.6.3"
-flate2 = "1.0.24"
-xz2 = "0.1.7"
 guppy = "0.15.0"
 camino = "1.1.1"
 semver = "1.0.14"
@@ -48,6 +44,7 @@ parse-changelog = { version = "0.5.3", default-features = false }
 newline-converter = "0.2.2"
 axoproject = { version = "0.3.0", default-features = false, features = ["cargo-projects"] }
 dialoguer = "0.10.4"
+axoasset = "0.2.0"
 
 [dev-dependencies]
 insta = { version = "1.26.0", features = ["filters"] }
--- a/cargo-dist/src/installer.rs
+++ b/cargo-dist/src/installer.rs
@@ -2,9 +2,10 @@
 //!
 //! In the future this might get split up into submodules.
 
-use std::{fs::File, io::BufWriter};
+use std::fs::File;
 
-use camino::Utf8Path;
+use axoasset::{AxoassetError, LocalAsset};
+use camino::{Utf8Path, Utf8PathBuf};
 use miette::{Context, IntoDiagnostic};
 use newline_converter::{dos2unix, unix2dos};
 use serde_json::json;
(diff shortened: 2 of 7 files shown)


cargo-dist/Cargo.toml
```diff
@@ -48,6 +44,7 @@ parse-changelog = { version = "0.5.3", default-features = false }
 newline-converter = "0.2.2"
 axoproject = { version = "0.3.0", default-features = false, features = ["cargo-projects"] }
 dialoguer = "0.10.4"
+axoasset = "0.2.0"
```
Gankra, line 47: In my ideal world we have some pretty fine-grained feature flags on axoasset to toggle the various compression formats on, to avoid all our projects getting bloated with unused dependencies.

cargo-dist/src/installer.rs
```diff
@@ -20,20 +21,25 @@ pub(crate) fn generate_install_sh_script(
     info: &InstallerInfo,
 ) -> Result<(), miette::Report> {
     let installer_file = &info.dest_path;
-    let mut file = File::create(installer_file)
+    // Unwrapping here because the path is basically guaranteed to have a parent at this point
+    let dest_dir = installer_file.parent().unwrap();
+    // Potential FIXME: Have an axoasset method that _just_ creates a file, without writing
+    // anything (think `File::create`)
+    LocalAsset::write_new("", installer_file.as_str(), dest_dir.as_str())
```
Gankra, line 28: hmm interesting, we ultimately buffer up the entire result before writing to the file anyway (to do newline cleanups), and presumably the more transactional you can make "create a file, write its contents to completion" the better, right? So maybe we should just move the file creation to the end?

Might make it harder to test this code later though...?

Gankra: Also I forget, what's the significance of making us pass the file path and the dir path?

Or wait, I think this is maybe just incorrect code, and you should be getting `installer_file.filename()` or something?

cargo-dist/src/installer.rs
```diff
@@ -106,7 +112,11 @@ fn write_install_sh_script<W: std::io::Write>(
         .replace("{{ARTIFACT_DOWNLOAD_URL}}", base_url)
         .replace("{{PLATFORM_INFO}}", &entries);
 
-    f.write_all(dos2unix(&install_script).as_bytes())?;
+    LocalAsset::write_new(
+        &dos2unix(&install_script),
+        installer_file.as_str(),
+        dest_dir.as_str(),
+    )?;
```
Gankra, line 119: ok wait yeah now this really doesn't make sense, we're making the file twice..?

Gankra (changes requested):
the installer code seems a bit messed up but i'm liking the changes elsewhere

cargo-dist/src/installer.rs
shadows-withal: It is incorrect, good catch! I'll move the write calls to the end of the function, I think when testing those functions, we might want to structure them somewhat differently anyways, so I don't think that's necessarily a huge blocker.

shadows-withal: As for why we pass filename and directory name separately, I have no idea, that's what the signature already looked like when I came upon it. We could probably squash those arguments, though.

cargo-dist/Cargo.toml
shadows-withal: That sounds like a good idea!

Gankra:
Closing in favour of #295, thanks again for doing this all!

(Some fs calls will remain, but they're in components that are getting rewritten anyway with #283 and https://github.com/axodotdev/axoproject/pull/26)

PR #230, opened 2023-04-28
Cli cleanup

Gankra:
* adds flags to `init`
  * --yes auto-accepts all default/recommended values, making init non-interactive
  * --no-generate-ci disables automatically running generate-ci at the end
* removes the alias from `cargo dist` to `cargo dist build`, making bare `cargo dist` just a help page
* adds a bunch of short aliases for common flags


cargo-dist/src/cli.rs
```diff
@@ -101,18 +100,18 @@ pub struct Cli {
...
 #[derive(Subcommand)]
 pub enum Commands {
     /// Build artifacts
     #[clap(disable_version_flag = true)]
     Build(BuildArgs),
-    /// Initialize default settings in your Cargo.toml
+    /// Setup or update cargo-dist
+    ///
+    /// This will interactively guide you through the process of selecting configuration options
+    /// and will also automatically run 'cargo dist generate-ci' afterwards as necessary. It will
```
ashleygwilliams, line 113: interesting choice to auto run it! (not sure how i feel but it's not immediately negative heh). we may want to emphasize that some how (is there out put that says it? if so that's likely enough)

ashleygwilliams: related q: can running the generate-ci be destructive at all?

cargo-dist/src/lib.rs
```diff
@@ -597,10 +597,15 @@ fn really_zip_dir(src_path: &Utf8Path, dest_path: &Utf8Path) -> Result<()> {
 
 /// Arguments for `cargo dist init` ([`do_init`][])
 #[derive(Debug)]
-pub struct InitArgs {}
+pub struct InitArgs {
```
ashleygwilliams, line 600: for lib version of this i would either drop it or i would turn this into some sort of config struct that we expect folks to hand us- i think having interaction in this part is a bit odd

ashleygwilliams: (and move the interaction to a different place) otherwise it is very confusing to read tbh, sorta a "uncanny valley" space

ashleygwilliams (changes requested):
i think in general, as discussed in channel, the strict bin/lib split is really pressed by this PR. the duplication is confusing given that the terminal interaction is still in the lib but the clap stuff is not. i would:

- maintain the split by putting all init stuff including interaction in a mod that only the bin uses and let the lib take a config struct that has the same value interface but not the interaction
- give up on the split and get rid of the duplication

i probably vaguely prefer the first one due to sunk cost and the chance it's useful in the future (honesty likely tbh)

cargo-dist/src/cli.rs
Gankra: also note that this behaviour was "totally" non-destructive in the original design of `init` where it refused to run if you were already init

cargo-dist/src/cli.rs
```diff
@@ -133,18 +132,20 @@ pub enum Commands {
...
-    pub version: Version,
+    /// If you want to know what running your cargo-dist CI will produce,
+    /// this is the command for you! It should run the exact same logic and do some
+    /// basic integrity checks.
+    ///
+    /// This is roughly an alias for:
+    ///
+    ///     cargo dist manifest --artifacts=all --no-local-paths
+    #[clap(disable_version_flag = true)]
+    Status(StatusArgs),
```
ashleygwilliams, line 145: i definitely would prefer some more akin to plan as i think status is confusing (since it's telling you what *will* happen in the future, not what is currently happening)

Gankra: hmmm

Gankra: gonna merge this for now and we can rename before release if we want

Issue #200, opened 2023-03-09
all items in release table say "tarball", even for zip files


zkat:
I would expect the `.zip` to say `zip file` or similar, here.

![image](https://user-images.githubusercontent.com/17535/224153723-de64c8f8-2932-4af5-98e6-f00e11196d6a.png)

Gankra:
Yeah internally cargo-dist considers tarballs and zips to be "the same thing", and tarball felt like the most familiar term for end users. This table is a mess and mostly just exists as a proof of concept, and should be completely reworked (#70). I didn't worry too hard about it because this oranda is the Good Version of this stuff, but I wanted to have something for people Not Using Oranda (which especially now is "everyone").

Gankra:
This format changed a long time ago so this isn't a thing.

Issue #120, opened 2023-02-12
Support signing binaries


rgreinho:
As a developer I would like to have the ability to sign the binaries I release.

This could be done using [sigstore](https://github.com/sigstore) as they provide a fantastic keyless signing ecosystem (can use GitHub OIDC), as well as rust bindings [sigstore-rs](https://github.com/sigstore/sigstore-rs).

Gankra:
possible dupe of #21

passcod:
Registering interest in this implementation as we're going to add support for checking sigstore in binstall in the near-but-not-defined future, so would be annoying if we were incompatible by accident

Gankra:
With #243 done this is the next logical step for integrity/checksum/signing stuff but i'm definitely in "i have no idea what i'm doing" territory with sigstore

dunxen:
> With #243 done this is the next logical step for integrity/checksum/signing stuff but i'm definitely in "i have no idea what i'm doing" territory with sigstore

Although no releases yet, we started testing with `cargo-dist` before #243 went in. Unfortunately ended up modifying the generated workflow which I guess defeats the purpose of a feature of `cargo-dist`, but we ended up with something working with the `cosign`  [GH Action](https://github.com/marketplace/actions/cosign-installer). We decided generating one checksum file and signing that was easier for us. Although not with GitHub OIDC as that's not available for plain blobs with cosign yet, just container images. So we had to generate a keypair using the approach [here](https://docs.sigstore.dev/cosign/git_support/#key-generation-and-management).

We definitely want to rework our workflow to align with what y'all decide to do as we'd like less moving parts (and not modifying generated workflows, although was just a hack for PoC). Happy to provide feedback too, and thanks for an awesome tool! :)

Our frankenflow: https://github.com/lndk-org/lndk/blob/5f0ed07eb3b669d6dec3edb1813b50f18e78bc6e/.github/workflows/release.yml#L134

passcod:
FTR The design we're thinking of with binstall is to use the cosign (sign-blob and) verify-blob workflows (but in Rust, not through the CLI go tools) directly on the packages / archives / distributed artifacts, not on generated checksum files, as cosign [already does the hashing internally](https://github.com/sigstore/cosign/blob/main/pkg/cosign/pivkey/pivkey.go#L265) and because that lets us verify from the files we're already downloading, rather than having to fetch an extra one. But awaiting more advice / consensus / free time before implementing :)

dunxen:
[...]

PR #243, opened 2023-05-09
add basic support for shasums 

Gankra:
We now default generate .sha256 checksum files (one for each archive). Example:

image

A new `checksum = "..."` config has been added, accepting 3 values:

* "sha256" (default)
* "sha512"
* "false" (disable checksumming)

All checksums are implemented with the sha2 crate and match sha256sum and sha512sum's hashes. The resulting file is just a lowercase hex string of the hash (no filename included, since each file gets its own .sha256 file). I've defaulted to sha256 since that seems to be the default other people use, and if you're really serious about hardening this stuff you probably want actual signing/pki backing this up (like #120).

I've also cleaned up the Github Release table output a bit as part of this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,6 +121,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "axoasset"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


Gankra:
Oh I guess this drive-by fixes #200 by just not describing the artifact anymore

PR #250, opened 2023-05-10
Fallback to the  Unreleased section in changelogs

### LINKED ISSUE #145: Account for the 'Unreleased' section when parsing release notes

Gankra:
parse_changelog supports a special section title of 'Unreleased', which we can default to if we're publishing a prerelease and fail to find any entry for it.

Fixes #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+placeholder
+
 # Version 0.0.7 (2023-05-09)
 
 This is just a quick little release that makes the npm package tarballs we can generate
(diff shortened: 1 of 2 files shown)


Gankra:
bubbled to the top of my stack because i just cut 0.1.0-prerelease.1 without a changelog entry which of course broke tests that expect us to always have a changelog entry for cargo-dist. This should make it more reliable/easy to always have a heading (could even make cargo-release auto-add it ([example from rust-minidump](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-stackwalk/Cargo.toml#L39-L42)).

Gankra:
[oh boy github ci is really busted huh](https://www.githubstatus.com/incidents/pr3498h3qkfy)

PR #275, opened 2023-07-17
feat(toolchain): deprecate rust-toolchain-version in favor of rust-toolchain.toml

### LINKED ISSUE #205: default to the version specified in `rust-toolchain.toml` when `rust-toolchain-version` not provided
I use cargo's `rust-toolchain.toml` to pin my rust version to a specific stable release.
This enforces that developers and CI all run the same version of rust.
It also allows me to set CI to fail if any rustc or clippy warnings occur without violating the "not rocket science" rule.

I think this should be the standard way to set what rust version to use, with the `rust-toolchain-version` acting as override and automatically setup when the user hasnt created a `rust-toolchain.toml`

I think the following logic would work:
* If a `rust-toolchain.toml` specifies a `channel` (the toolchain version) then:
   * `cargo dist init` should not set a `rust-toolchain-version`
* If a `rust-toolchain.toml` specifies a `channel` (the toolchain version) and cargo dist's `rust-toolchain-version` is unset then:
   * `cargo dist generate-ci` should not include the `name: Install cargo-dist` section. The `rust-toolchain.toml` will cause rustup to automatically install the specified toolchain the first time `cargo` is invoked in a directory containing `rust-toolchain.toml` or one of its subdirectories)

Gankra:
Deprecation reason: [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) is a more standard/universal mechanism for pinning toolchain versions for reproducibility. Teams without dedicated release engineers will likely benefit from unpinning their toolchain and letting the underlying CI vendor silently update them to 'some recent stable toolchain', as they will get updates/improvements and are unlikely to have regressions.

If you delete the key, generate-ci won't explicitly setup a toolchain, so whatever's on the machine will be used (with things like rust-toolchain.toml behaving as normal). Before being deprecated the default was to 
ustup update stable, but this is no longer the case.

We will warn if the key is set, recommending rust-toolchain.toml

----

This also adds system_info.cargo_version_line to dist-manifest.json, giviing back some reproducibility info to folks using unpinned toolchains.

fixes #205

### DIFF
--- a/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
+++ b/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
@@ -61,6 +61,17 @@ expression: json_schema
       "items": {
         "$ref": "#/definitions/Release"
       }
+    },
+    "system_info": {
+      "description": "Info about the toolchain used to build this announcement",
+      "anyOf": [
+        {
+          "$ref": "#/definitions/SystemInfo"
+        },
+        {
+          "type": "null"
+        }
+      ]
     }
   },
   "definitions": {
@@ -322,6 +333,19 @@ expression: json_schema
           "items": {
             "type": "string"
           }
+        }
+      }
+    },
+    "SystemInfo": {
+      "description": "Info about the system/toolchain used to build this announcement.\n\nNote that this is info from the machine that generated this file, which *ideally* should be similar to the machines that built all the artifacts, but we can't guarantee that.\n\ndist-manifest.json is by default generated at the start of the build process, and typically on a linux machine because that's usually the fastest/cheapest part of CI infra.",
+      "type": "object",
+      "properties": {
+        "cargo_version_line": {
+          "description": "The version of Cargo used (first line of cargo -vV)\n\nNote that this is the version used on the machine that generated this file, which presumably should be the same version used on all the machines that built all the artifacts, but maybe not! It's more likely to be correct if rust-toolchain.toml is used with a specific pinned version.",
+          "type": [
+            "string",
+            "null"
+          ]
         }
       }
     }
--- a/book/src/config.md
+++ b/book/src/config.md
@@ -67,9 +67,11 @@ If you delete the key, generate-ci will just use the version of cargo-dist that'
 
 ### rust-toolchain-version
 
-> since 0.0.3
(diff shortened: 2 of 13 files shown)


cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
```diff
@@ -324,6 +335,19 @@ expression: json_schema
           }
         }
       }
+    },
+    "SystemInfo": {
+      "description": "Info about the system/toolchain used to build this announcement.\n\nNote that this is info from the machine that generated this file, which *ideally* should be similar to the machines that built all the artifacts, but we can't guarantee that.\n\ndist-manifest.json is by default generated at the start of the build process, and typically on a linux machine because that's usually the fastest/cheapest part of CI infra.",
+      "type": "object",
+      "properties": {
+        "cargo_version_line": {
+          "description": "The version of Cargo used (first line of cargo -vV)\n\nNote that this is the version used on the machine that generated this file, which presumably should be the same version used on all the machines that built all the artifacts, but maybe not! It's more likely to be correct if rust-toolchain.toml is used with a specific pinned version.",
```
ashleygwilliams, line 344: "It's more likely" -> if rust-toolchain.toml is present and uses a specific version it will always be correct, right? the only way it wouldn't be correct is if rust-toolchain.toml wasn't present or if it was but specified something like "latest"?

ashleygwilliams (commented):
do we think that we should fully deprecate on the off-chance people who do not use rustup want a way to pin?

cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
Gankra: per https://rust-lang.github.io/rustup/overrides.html, several things have precedence over rust-toolchain.toml (like RUSTUP_TOOLCHAIN and `rustup override`), and I can't *guarantee* you haven't messed up your build system by using one of these in addition to rust-toolchain.toml.

Gankra:
> do we think that we should fully deprecate on the off-chance people who do not use rustup want a way to pin?

The way we use this field is to invoke rustup, so the feature was always dependent on it. The new behaviour when the field is removed essentially prevents cargo-dist from getting in the way of someone using their own system like nix or whatever to setup the toolchain.

PR #276, opened 2023-07-17
feat(fault-isolate): prefer independent tasks and don't fail-fast

Gankra:
[feat(fault-isolate): add merge-tasks config and default it to false](https://github.com/axodotdev/cargo-dist/commit/10f740b24f22094f49858ce98b4d9e76d6c9fbe8)

This config specifies whether we should try to merge otherwise-parallelizable tasks onto the same machine, sacrificing latency and fault-isolation for more the sake of minor effeciency gains.

For example, if you build for x64 macos and arm64 macos, by default we will generate ci which builds those independently on separate logical machines. With this enabled we will build both of those platforms together on the same machine, making it take twice as long as any other build and making it impossible for only one of them to succeed.

The default is false. Before 0.1.0 it was always true and couldn't be changed, making releases annoyingly slow (and technically less fault-isolated). This config was added to allow you to restore the old behaviour, if you really want.

----

[feat(fault-isolate): add fail-fast config and default it false](https://github.com/axodotdev/cargo-dist/commit/0291fd28f7db1334ea601ef92060486ef23be080)

When building a release you might discover that an obscure platform's build is broken. When this happens you have two options: give up on the release entirely (fail-fast = true), or keep trying to build all the other platforms anyway (fail-fast = false).

cargo-dist was designed around the 'keep trying' approach, as we create a draft Release
and upload results to it over time, undrafting the release only if all tasks succeeded.
The idea is that even if a platform fails to build, you can decide that's acceptable
and manually undraft the release with some missing platforms.

(Note that the dist-manifest.json is produced before anything else, and so it will assume
that all tasks succeeded when listing out supported platforms/artifacts. This may make
you sad if you do this kind of undrafting and also trust the dist-manifest to be correct.)

Prior to 0.1.0 we didn't set the correct flags in our CI scripts to do this, but now we do.
This flag was introduced to allow you to restore the old behaviour if you prefer.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -250,6 +250,20 @@ pub struct DistMetadata {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "precise-builds")]
     pub precise_builds: Option<bool>,
+
+    /// Whether we should try to merge otherwise-parallelizable tasks onto the same machine,
+    /// sacrificing latency and fault-isolation for more the sake of minor effeciency gains.
+    ///
+    /// (defaults to false)
+    ///
+    /// For example, if you build for x64 macos and arm64 macos, by default we will generate ci
+    /// which builds those independently on separate logical machines. With this enabled we will
+    /// build both of those platforms together on the same machine, making it take twice as long
+    /// as any other build and making it impossible for only one of them to succeed.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "merge-tasks")]
+    pub merge_tasks: Option<bool>,
 }
 
 impl DistMetadata {
@@ -268,6 +282,7 @@ impl DistMetadata {
         // * rust_toolchain_version
         // * ci
         // * precise_builds
+        // * merge_tasks
 
         if self.installers.is_none() {
             self.installers = workspace_config.installers.clone();
@@ -391,6 +406,8 @@ pub struct DistGraph {
     pub dist_dir: Utf8PathBuf,
     /// Whether to bother using --package instead of --workspace when building apps
     pub precise_builds: bool,
+    /// Whether to try to merge otherwise-parallelizable tasks the same machine
+    pub merge_tasks: bool,
     /// The desired cargo-dist version for handling this project
     pub desired_cargo_dist_version: Option<Version>,
     /// The desired rust toolchain for handling this project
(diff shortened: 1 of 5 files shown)


Gankra:
nb i haven't properly tested that i actually understand how fail-fast works in github, and the continue-on-error flag is not clearly documented so I don't know if it's actually required to get the behaviour i want: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

Gankra:
# Test 1: Happy Path

✅ Normal releases still work fine
✅ Mac builds are concurrent/independent now

* workflow: https://github.com/Gankra/axolotlsay/actions/runs/5587603794
* release: https://github.com/Gankra/axolotlsay/releases/tag/v0.3.19


-------


# Test 2: Break arm64 macos

I intentionally sabotaged the build on aarch64-apple-darwin. 

✅ Only arm64 macos failed
✅ All other tasks in the matrix succeeded
✅ The release was left as a draft with all other assets succesfully uploaded.
✅ The Release workflow was marked as an overall failure
✅ Workflows that were watching for Release success skipped themselves as status != success. 
✅ I edited a mention of the arm64 macos build from the github release body, and undrafted it without any issues

⚠️ The arm64 macos build took the longest, so the result is inconclusive (it couldn't cancel anything else in the matrix).

* commit: https://github.com/Gankra/axolotlsay/commit/f155525ae99b4fc95017f0e430a64cf15f453421
* workflow: https://github.com/Gankra/axolotlsay/actions/runs/5587952494
* release: https://github.com/Gankra/axolotlsay/releases/tag/v0.3.20


-------



# Test 3: Break x64 linux

Changed test 2 to break the usually-fastest build: x64 linux.

✅ Same results as before, but this time linux finished (failed) first, so we can be confident everything worked properly.

* commit: https://github.com/Gankra/axolotlsay/commit/fda7301629b06a38bf961c7420c108f435778041
* workflow: https://github.com/Gankra/axolotlsay/actions/runs/5588261130
* release: https://github.com/Gankra/axolotlsay/releases/tag/v0.3.21

cargo-dist/src/tasks.rs
```diff
@@ -927,12 +973,18 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             if package_config.cargo_dist_version.is_some() {
                 warn!("package.metadata.dist.cargo-dist-version is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package.manifest_path);
             }
-            if package_config.cargo_dist_version.is_some() {
+            if package_config.rust_toolchain_version.is_some() {
```
shadows-withal, line 976: nice catch!

shadows-withal (approved):

PR #277, opened 2023-07-18
feat(precise-builds): add a precise-builds config and default it false

Gankra:
Setting it to true tells us to Build only the required packages, and individually (default: false)

By default when we need to build anything in your workspace, we build your entire workspace with --workspace. This setting tells cargo-dist to instead build each app individually.

On balance, the Rust experts we've consulted with find building with --workspace to be a safer/better default, as it provides some of the benefits of a more manual [workspace-hack][], without the user needing to be aware that this is a thing.

TL;DR: cargo prefers building one copy of each dependency in a build, so if two apps in your workspace depend on e.g. serde with different features, building with --workspace, will build serde once with the features unioned together. However if you build each package individually it will more precisely build two copies of serde with different feature sets.

The downside of using --workspace is that if your workspace has lots of example/test crates, or if you release only parts of your workspace at a time, we build a lot of gunk that's not needed, and potentially bloat up your app with unnecessary features.

If that downside is big enough for you, this setting is a good idea.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -223,6 +223,33 @@ pub struct DistMetadata {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub checksum: Option<ChecksumStyle>,
+
+    /// Build only the required packages, and individually (since 0.1.0) (default: false)
+    ///
+    /// By default when we need to build anything in your workspace, we build your entire workspace
+    /// with --workspace. This setting tells cargo-dist to instead build each app individually.
+    ///
+    /// On balance, the Rust experts we've consulted with find building with --workspace to
+    /// be a safer/better default, as it provides some of the benefits of a more manual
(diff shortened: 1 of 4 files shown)


Gankra:
Seems to behave correctly on the examples I've tried

cargo-dist/src/tasks.rs
```diff
@@ -1690,19 +1738,46 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+                    let bin = self.binary(bin_idx);
+                    builds_by_pkg_spec
+                        .entry(bin.pkg_spec.clone())
+                        .or_insert(vec![])
+                        .push(bin_idx);
+                }
+                for (pkg_spec, expected_binaries) in builds_by_pkg_spec {
+                    builds.push(BuildStep::Cargo(CargoBuildStep {
+                        target_triple: target.clone(),
+                        // Just build the whole workspace for now
```
shadows-withal, line 1754: This comment needs to be updated

Issue #41, opened 2023-01-12
better installer behaviour


Gankra:
Various improvements:

* cargo home
  * [x] respect CARGO_HOME
  * [ ] Update `~/.cargo/.crates.toml` to register the binary?
* other install dirs
  * [x] #235 
  * [ ] Look for other big likely-on-PATH dirs?
* better installs
  * [ ] #314 
  * [ ] Run an "install" subcommand of the binary itself?

Gankra:
Discussing this with Ashley we concluded that it might be for the best if we do like [apollo's rover does](https://github.com/apollographql/rover/tree/main/installers/binstall) and try to build out machinery for CLIs to have an `install` subcommand (which would also theoretically support self-update and uninstalling and whatnot).

That way we can have more hairy logic in rust and not bash scripts (yuck!). This may involve us building out a reusable library for everyone to use for this. It's possible this library already exists in some of the [binstalk crates in cargo-binstall](https://github.com/cargo-bins/cargo-binstall/tree/main/crates), or is aching to be extracted from them.

Gankra:
Relevant issues rustup had a while back 

* https://github.com/rust-lang/rustup/pull/2387
* https://github.com/rust-lang/rustup/issues/2106

jyn514:
It would be nice if this also respected CARGO_HOME instead of hardcoding ~/.cargo.

mistydemeo:
We should update the milestone again.

Gankra:
This issue doesn't really make much sense anymore / needs to be revived as several smaller issues.

PR #284, opened 2023-07-21
feat(install-path): add install-path config

### LINKED ISSUE #235: install-path customization
As noted in #41, the shell/powershell installers unconditionally install into cargo home, but we should let the user configure that.

This should possibly support some amount of templating, as seen in [the config for cargo-binstall](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md).

Theoretical usecases with strawman syntax:

* `install-path = "{cargo}"` (use cargo's global install dir)
* `install-path = "{npm}"` (use npm's global install dir?)
* `install-path = "~/.my-app/"` (install to a hardcoded dir in the user's home)
* `install-path = "{xdg}/my-app/"` (install to whatever the hell xdg says you're supposed to do)

Gankra:
fixes #235 and the most important part of #41

Also fixes the way we handle parse errors or dist-metadata (before we would just replace corrupt values with Default, which, is confusing).

Possible values include:

* "CARGO_HOME": install to the user's cargo bins (default)
* "~/some/subdir": install to the given subdir of the user's home dir
* "$ENV_VAR/some/subdir": install to the given subdir of the given env var

Examples: "~/.my-app", "$MY_APP/", "$MY_APP/bin/"

If the resulting path doesn't exist, a directory will be created at its location.

All 3 of these settings will error out if required env vars aren't set:

* "CARGO_HOME": first tries $CARGO_HOME, then $HOME
* "~/subdir": tries $HOME
* "$ENV_VAR/subdir": tries $ENV_VAR

(Note that $HOME/subdir is technically valid syntax, but won't behave the way you want,
because in powershell $HOME isn't a normal env var. Use `~/subdir` and we'll handle these
details for you.)


shadows-withal (commented):
only looking over the new templates, and i won't pretend to understand most of the stuff that goes on in the shell script, but it looks good to me, i have only nits as of this point

PR #292, opened 2023-07-28
chore: introduce integration tests for installers

Gankra:
* fetches axolotolsay (can be configured to any repo)
* applies an overlay to the Cargo.toml
* runs cargo dist build -aglobal on it (builds installer scripts)
* runs shellcheck on the shell scripts (if available)
* runs PSScriptAnalyzer on the powershell scripts (if available)
* insta snapshots both scripts

The OVERRIDE_CARGO_BIN_EXE_cargo-dist variable also lets you inject a different binary for cargo-dist, which will let us run these tests for shippable binaries (infra not yet implemented)


Gankra:
note that this *does not* test that running the scripts behaves as desired, I haven't yet figured out how I want that to work.

Gankra:
TODO: install shellcheck and PSScriptAnalyzer on our infra before running tests

PR #295, opened 2023-07-31
chore: use axoasset's tar/zip support

Gankra:


Gankra:
depends on https://github.com/axodotdev/axoasset/pull/61

mistydemeo (approved):

PR #298, opened 2023-08-01
fix: strip .git from repo URL before appending

### LINKED ISSUE #263: Installer and download links don't work when repository is specified with .git
On github, the canonical clone url for repository ends with `.git` (`https://github.com/user/repo.git`), while the canonical url for the web interface doesn't have `.git` (`https://github.com/user/repo`).

However, you still can clone repo with url without the `.git` and go to the web interface by url that has git.

This creases some confusion between those two, and with cargo book saying that "The repository field should be a URL to the source repository for your package.", it's not clear which url should be used. There are about 5.5k crates on crates.io that specify the repository url with `.git` and around 88k without.

However, you can't use `.git` url to create release urls, so specifying such a url makes `cargo-dist` create invalid urls in the downloads table and in the installers.

mistydemeo:
It's legal for a repository URL to end with .git, but cargo-dist currently assumes it's a web URL rather than a git clone URL before working with it. This makes sure we strip the .git part of the URL before tacking on the `/release` path so we generate a valid URL.

I considered putting this in `axoproject` but it felt wrong to be messing with the repository URL from `Cargo.toml` over there.

Fixes #263.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,26 +87,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.72"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b13c32d80ecc7ab747b80c3784bce54ee8a7a0cc4fbda9bf4cda2cf6fe90854"
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Gankra:
Not sure if you noticed but this is also an annoyance in oranda (https://github.com/axodotdev/oranda/pull/563), so I think in fact it would be reasonable to have `repository` normalization logic in axoproject (even if only as a method you can call on `PackageInfo`).

Gankra (approved):
(Approving regardless so you can make the final call on what to do)

mistydemeo:
Good call, opening a PR to add a method.

PR #300, opened 2023-08-02
chore: add misty's project to the gallery

Gankra:
This also reworks the gallery code to properly have a global lock between all Repos since we cd into the repo before running a test, which is global mutable state. I did this by requiring you to define a global lock for Tools that must be passed by-ref into each TestContextLock::new call, and using that as THE global lock (which is fine, Tools getting initialized once for each Repo was jank anyway).

### DIFF
--- a/cargo-dist/tests/gallery/dist.rs
+++ b/cargo-dist/tests/gallery/dist.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::sync::Mutex;
 
 use axoasset::LocalAsset;
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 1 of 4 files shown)


cargo-dist/tests/snapshots/akaikatana_basic-installers.snap
```diff
@@ -0,0 +1,984 @@
...
+    local _bins
+    local _zip_ext
+    local _artifact_name
+
+    # Lookup what to download/unpack based on platform
+    case "$_arch" in 
+        "aarch64-apple-darwin")
+            _artifact_name="akaikatana-repack-aarch64-apple-darwin.tar.xz"
+            _zip_ext=".tar.xz"
+            _bins="akextract akmetadata akrepack"
```
Gankra, line 133: multi-bins! (one day we'll handle spaces in binary names... probably a lot easier with jinja)

cargo-dist/tests/snapshots/akaikatana_basic-installers.snap
```diff
@@ -0,0 +1,984 @@
...
+  $old_erroractionpreference = $ErrorActionPreference
+  $ErrorActionPreference = 'stop'
+
+  Initialize-Environment
+
+  # Platform info injected by cargo-dist
+  $platforms = @{
+    "x86_64-pc-windows-msvc" = @{
+      "artifact_name" = "akaikatana-repack-x86_64-pc-windows-msvc.zip"
+      "bins" = "akextract.exe", "akmetadata.exe", "akrepack.exe"
```
Gankra, line 759: multi-bins!

mistydemeo (approved):

cargo-dist/tests/gallery/dist.rs
```diff
@@ -18,15 +19,30 @@ const ENV_RUIN_ME: &str = "RUIN_MY_COMPUTER_WITH_INSTALLERS";
...
+pub static AKAIKATANA_REPACK: TestContextLock<Tools> = TestContextLock::new(
+    &TOOLS,
+    &Repo {
+        repo_owner: "mistydemeo",
+        repo_name: "akaikatana-repack",
+        commit_sha: "9516f77ab81b7833e0d66de766ecf802e056f91f",
+        app_name: "akaikatana-repack",
+        bins: &["akextract", "akmetadata", "akrepack"],
+    },
+);
```
Gankra, line 45: this is all you need to do to add a new repo

cargo-dist/tests/integration-tests.rs
```diff
@@ -77,6 +77,27 @@ scope = "@axodotdev"
...
+installers = ["shell", "powershell"]
+targets = ["x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows-msvc", "aarch64-apple-darwin"]
+
+"#))?;
+        results.check_all(ctx, ".cargo/bin/")?;
+
+        Ok(())
+    })
+}
+
```
Gankra, line 100: And this is all you need to do to use that repo in a test (note that even though this is the vanilla `workspace.metadata.dist` we still have to overwrite it to make `cargo-dist-version` be the currently running one to avoid integrity checks). Potentially this should be automated in `patch_cargo_toml` because it's tedious for every test to do...

Issue #308, opened 2023-08-09
installer: homebrew


Gankra:
It would be nice for cargo-dist to be able to generate a homebrew formula for your package, similar to the npm installer, which:

* detects your platform (constructs a target triple)
* fetches your prebuilt binaries
* unpacks them as appropriate for a homebrew package

Initial discussion is that this requires us to:

* generate a ruby script ("the formula")
* help the user define a tap for their formulas (git repo containing some text files)
   * easiest might be adding a `formulas/` dir to the root of their actual project

I expect that much of the logic here will be shared with the "shell" installer. Hopefully a lot of the places that need to be changed can be found by just adding a new variant to this enum and following the compiler errors.

https://github.com/axodotdev/cargo-dist/blob/7a8ce388b220a1e6deca44f423fa9536a9f4f645/cargo-dist/src/cli.rs#L249-L258

Gankra:
If you want to publish a GUI, you want cask (#309)

Gankra:
done in #318, #339, and #353 and available in 0.2.0-prerelease.6 (0.2.0 out this week) 

🎊

PR #318, opened 2023-08-11
feat: add Homebrew formula installer support

mistydemeo:
This starts the work of adding a Homebrew installer. At this point, it generates a usable formula file that installs binaries and doc files, but not other optional features we might want. It also doesn't handle the work of uploading it to a tap yet.

Sample generated formula:

```ruby
class AkaikatanaRepack < Formula
  if Hardware::CPU.type == :arm
    url "https://github.com/mistydemeo/akaikatana-repack/releases/download/v0.2.0/akaikatana-repack-aarch64-apple-darwin.tar.xz"
    sha256 ""
  else
    url "https://github.com/mistydemeo/akaikatana-repack/releases/download/v0.2.0/akaikatana-repack-x86_64-apple-darwin.tar.xz"
    sha256 ""
  end
   license "GPL-2.0-or-later" 

  def install
    bin.install "akextract",  "akmetadata",  "akrepack" 
  end
end
```

I tested it via `brew install`, and it works just fine; it installs all of the defined binaries, plus the README and license files.

refs #308

### DIFF
--- a/cargo-dist/Cargo.toml
+++ b/cargo-dist/Cargo.toml
@@ -36,6 +36,7 @@ thiserror = "1.0.35"
 tracing = { version = "0.1.36", features = ["log"] }
 serde = { version = "1.0.144", features = ["derive"] }
 cargo_metadata = "0.15.0"
+cruet = "0.13.3"
 guppy = "0.15.0"
 camino = "1.1.1"
 semver = "1.0.14"
--- a/cargo-dist/src/backend/installer/homebrew.rs
+++ b/cargo-dist/src/backend/installer/homebrew.rs
@@ -1,0 +1,43 @@
+//! Code for generating installer.sh
+
+use axoasset::LocalAsset;
+use serde::Serialize;
+
+use super::InstallerInfo;
+use crate::{
+    backend::templates::{Templates, TEMPLATE_INSTALLER_RB},
+    errors::DistResult,
+    installer::ExecutableZipFragment,
+};
+
+/// Info about a Homebrew formula
+#[derive(Debug, Clone, Serialize)]
+pub struct HomebrewInstallerInfo {
+    /// The application's name
+    pub name: String,
+    /// Formula class name
+    pub formula_class: String,
+    /// The application's license, in SPDX format
+    pub license: Option<String>,
+    /// The URL to the application's homepage
+    pub homepage: Option<String>,
+    /// A brief description of the application
+    pub desc: Option<String>,
+    /// AMD64 artifact
+    pub x86_64: Option<ExecutableZipFragment>,
+    /// ARM64 artifact
+    pub arm64: Option<ExecutableZipFragment>,
+    /// List of binaries to install
+    pub binaries: Vec<String>,
+    /// Generic installer info
+    pub inner: InstallerInfo,
+}
+
+pub(crate) fn write_homebrew_formula(
+    templates: &Templates,
+    info: &HomebrewInstallerInfo,
+) -> DistResult<()> {
+    let script = templates.render_file_to_clean_string(TEMPLATE_INSTALLER_RB, info)?;
+    LocalAsset::write_new(&script, &info.inner.dest_path)?;
+    Ok(())
+}
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -52,6 +52,7 @@ use std::process::Command;
 
 use axoproject::{PackageIdx, WorkspaceInfo};
 use camino::Utf8PathBuf;
+use cruet::to_class_case;
 use guppy::PackageId;
 use miette::{miette, Context, IntoDiagnostic};
 use semver::Version;
@@ -59,7 +60,10 @@ use tracing::{info, warn};
 
 use crate::{
     backend::{
-        installer::{npm::NpmInstallerInfo, ExecutableZipFragment, InstallerImpl, InstallerInfo},
+        installer::{
+            homebrew::HomebrewInstallerInfo, npm::NpmInstallerInfo, ExecutableZipFragment,
+            InstallerImpl, InstallerInfo,
+        },
         templates::Templates,
     },
     config::{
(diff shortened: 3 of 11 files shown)


cargo-dist/Cargo.toml
```diff
@@ -36,6 +36,7 @@ thiserror = "1.0.35"
 tracing = { version = "0.1.36", features = ["log"] }
 serde = { version = "1.0.144", features = ["derive"] }
 cargo_metadata = "0.15.0"
+cruet = "0.13.3"
```
mistydemeo, line 39: I didn't feel great about adding another dep, I just wanted to not have to think about the ClassCase conversion.

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Formula class name
+    pub formula_class: String,
+    /// The application's license, in SPDX format
+    pub license: Option<String>,
+    /// The URL to the application's homepage
+    pub homepage: Option<String>,
+    /// A brief description of the application
+    pub desc: Option<String>,
+    /// AMD64 artifact
+    pub x86_64: Option<ExecutableZipFragment>,
```
mistydemeo, line 27: I figured that since Homebrew does need to specifically query the x86_64 and ARM binaries, it made sense to promote them to first-class properties of the installer info struct.

cargo-dist/src/tasks.rs
```diff
@@ -1103,6 +1108,134 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            kind: ArtifactKind::Installer(InstallerImpl::Homebrew(HomebrewInstallerInfo {
+                arm64,
+                x86_64,
+                name: app_name,
+                formula_class: formula_name,
+                desc: app_desc,
+                license: app_license,
+                homepage: app_homepage_url,
+                // For now, assume all packages contain the same binaries
+                binaries: artifacts[0].binaries.clone(),
```
mistydemeo, line 1222: This doesn't feel right; the formula currently assumes the binary list for ARM and x86_64 will always be the same.

cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -0,0 +1,16 @@
+class {{ formula_class }} < Formula
+  {%- if desc %} desc "{{ desc }}" {% endif %}
+  {%- if homepage %} homepage "{{ homepage }}" {% endif %}
+  if Hardware::CPU.type == :arm
+    url "{{ inner.base_url }}/{{ arm64.id }}"
+    sha256 "{{ arm64.checksum }}"
```
mistydemeo, line 6: Looks like these aren't generated right now; I'll have to figure out how to fix that.

Gankra: We do generate sha256 checksums for archives (as files), what do you need the shasum to cover? just the binary?

cargo-dist/src/tasks.rs
```diff
@@ -1103,6 +1108,134 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            }
+
+            if do_rosetta_fallback && target == X64_MACOS {
+                // Copy the info but respecify it to be arm64 macos
+                let mut arm_fragment = fragment.clone();
+                arm_fragment.target_triples = vec![ARM64_MACOS.to_owned()];
+                artifacts.push(arm_fragment.clone());
+                arm64 = Some(arm_fragment);
+            }
+            artifacts.push(fragment);
```
Gankra, line 1191: I assume it's intentional that even though we're doing fallback we're not actually assigning the x64 binary to the `arm64` var because we want homebrew et al to understand what's going on?

cargo-dist/src/tasks.rs
Gankra: Is that a fundamental constraint of homebrew formulas, or just a lot nastier to implement?

(tbh this isn't a big deal, there's no way to make the targets diverge in binaries rn)

cargo-dist/src/tasks.rs
```diff
@@ -1103,6 +1108,134 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+        let release = self.release(to_release);
+        let release_id = &release.id;
+        let Some(download_url) = &self.inner.artifact_download_url else {
+            warn!("skipping Homebrew formula: couldn't compute a URL to download artifacts from");
+            return;
+        };
+        // TODO ensure this is Homebrew-legal
+        let artifact_name = format!("{release_id}.rb");
+        let artifact_path = self.inner.dist_dir.join(&artifact_name);
+        let hint = format!("brew install {artifact_name}");
```
Gankra, line 1124: This presumably will need some kind of reference to the tap in the final impl right?

Gankra (commented):
Seems like you figured out the general gist, nice!

cargo-dist/templates/installer/homebrew.rb.j2
mistydemeo: It's for the tarball itself, rather than any of the contents.

cargo-dist/src/tasks.rs
mistydemeo: Yeah, that's right. For this PR I was focused on getting the formula file rendered out - figured it would make sense to leave the tap stuff to a followup.

cargo-dist/src/tasks.rs
mistydemeo: I was trying to avoid having a potentially unnecessary `if Hardware::CPU` check, but I think I'll just add an `if/else` to the template that avoids it if both sets are the same.

cargo-dist/src/tasks.rs
```diff
@@ -1098,6 +1108,136 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
```
Gankra, line 1130: This should probably be deleted until the tap followup?

Gankra (approved):
Great work on the tests!

PR #321, opened 2023-08-11
feat(features): add the ability to specify features

### LINKED ISSUE #274: Add support binaries with required features
In https://github.com/probe-rs/probe-rs/, we have a couple of binaries which only get built when a feature is enabled. Right now, cargo-dist tries to release these binaries even though they can't be built without the required feature.

It would be great if cargo-dist could enable the required features automatically, or at least have a setting for it in the configuration.

Might be a duplicate of #22

Gankra:
This adds the following configs to metadata.dist:

```toml
features = ["blah", "blargh"]
default-features = false
all-features = true
```

If applied to [workspace.metadata.dist] it will be applied to all packages

If applied to [package.metadata.dist] it will be applied to just that package (overriding the workspace value if it exists).

If not all packages end up with the same feature flags then precise-builds=true will be forced on.

fixes #274

### DIFF
--- a/book/src/config.md
+++ b/book/src/config.md
@@ -217,7 +217,7 @@ Build only the required packages, and individually.
 
 [See "inferring precise-builds" for the default behaviour.](#inferring-precise-builds)
 
-By default when we need to build anything in your workspace, we try to build your entire workspace with --workspace. This setting tells cargo-dist to instead build each app individually.
+By default when we need to build anything in your workspace, we try to build your entire workspace with `--workspace`. This setting tells cargo-dist to instead build each app individually.
 
 On balance, the Rust experts we've consulted with find building with --workspace to be a safer/better default, as it provides some of the benefits of a more manual [workspace-hack][], without the user needing to be aware that this is a thing.
 
@@ -234,7 +234,7 @@ If that downside is big enough for you, this setting is a good idea.
 
 Although cargo-dist prefers `--workspace` builds ([precise-builds](#precise-builds) = `false`) for the reasons stated above, it *will* attempt to check if that's possible, and use `--package` builds if necessary (`precise-builds = true`).
 
-If you explicitly set `precise-builds = false` and we determine --package builds are required, cargo-dist will produce an error. `precise-builds = true` will never produce an error.
+If you explicitly set `precise-builds = false` and we determine `--package` builds are required, cargo-dist will produce an error. `precise-builds = true` will never produce an error.
 
 Precise-builds are considered required when you use any of [features](#features), [all-features](#all-features), or [default-features](#default-features) *and* not all of the packages in your workspace have the same values set.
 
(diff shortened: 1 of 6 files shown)


Gankra:
TODO:

* [ ] add tests
* [ ] docs
* [ ] add error check for --all-features being set when anything else is..?
* [ ] can we validate if the features exist on the packages and return earlier errors (probably cargo-metadata/guppy tells us that)?

book/src/config.md
```diff
@@ -213,9 +213,11 @@ Example: `precise-builds = true`
 
 **This can only be set globally**
 
-Build only the required packages, and individually (default: false)
+Build only the required packages, and individually.
 
-By default when we need to build anything in your workspace, we build your entire workspace with --workspace. This setting tells cargo-dist to instead build each app individually.
+[See "inferring precise-builds" for the default behaviour.](#inferring-precise-builds)
+
+By default when we need to build anything in your workspace, we try to build your entire workspace with --workspace. This setting tells cargo-dist to instead build each app individually.
```
Yatekii, line 220:
```suggestion
By default when we need to build anything in your workspace, we try to build your entire workspace with `--workspace`. This setting tells cargo-dist to instead build each app individually.
```

book/src/config.md
```diff
@@ -228,6 +230,31 @@ If that downside is big enough for you, this setting is a good idea.
 [workspace-hack]: https://docs.rs/cargo-hakari/latest/cargo_hakari/about/index.html
 
 
+#### inferring precise-builds
+
+Although cargo-dist prefers `--workspace` builds ([precise-builds](#precise-builds) = `false`) for the reasons stated above, it *will* attempt to check if that's possible, and use `--package` builds if necessary (`precise-builds = true`).
+
+If you explicitly set `precise-builds = false` and we determine --package builds are required, cargo-dist will produce an error. `precise-builds = true` will never produce an error.
```
Yatekii, line 237:
```suggestion
If you explicitly set `precise-builds = false` and we determine `--package` builds are required, cargo-dist will produce an error. `precise-builds = true` will never produce an error.
```

book/src/config.md
```diff
@@ -297,6 +324,47 @@ Future Improvements:
 (Please file an issue if you have other requirements!)
 
 
+### features
+
+> since 0.2.0
+
+Example: `features = ["serde-support", "fancy-output"]`
+
+Specifies feature-flags that should be passed to a package when building it. This lets you enable features that should on be on "in production" but for whatever reason shouldn't be on by default.
```
Yatekii, line 333:
```suggestion
Specifies feature-flags that should be passed to a package when building it. This lets you enable features that should be on "in production" but for whatever reason shouldn't be on by default.
```

book/src/config.md
```diff
@@ -297,6 +324,47 @@ Future Improvements:
...
+If you use this you *probably* want to set it on `[package.metadata.dist]` and not `[workspace.metadata.dist]`. See ["inferring precise-builds"](#inferring-precise-builds) for details.
+
+
+### all-features
+
+> since 0.2.0
+
+Example: `all-features = true`
+
+Specifies that all features for a package should be enabled when building it (when set to true this tells us to pass --all-features to Cargo).
```
Yatekii, line 360:
```suggestion
Specifies that all features for a package should be enabled when building it (when set to true this tells us to pass `--all-features` to Cargo).
```

Yatekii (commented):
I added some nitpicks to comments - pardon me. Everything else looks good to me =) Thank you for prioritizing this :)

mistydemeo (approved):

Gankra:
merging with less tests than i'd like because it's pretty simple and i want to mess with this in a prerelease a bit

PR #327, opened 2023-08-14
feat(ci-plus): make upload-artifacts have a dynamic matrix

### LINKED ISSUE #233: make 'plan' more like terraform
As discussed in https://github.com/axodotdev/cargo-dist/issues/178#issuecomment-1452020295, it would be more robust if our CI scripts used [dynamic matrices](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-contexts-to-create-matrices) that were defined by `cargo dist plan`.

Step 1 to this is to have `plan` always spit out the information we compute in ci.rs here:

https://github.com/axodotdev/cargo-dist/blob/4b3f86083fb008707665199611f697fc1aaa8d1d/cargo-dist/src/ci.rs#L80-L100

An agnostic version of this output could also be used to empower others to unilaterally bring up their own CI infra by parsing it and lowering it to their own system. This seems plausible because cargo-dist really doesn't care about the specifics of a machine -- it just cares about "windows, linux, mac" and requires that machine to have rustup and enough of a system to run the curl-sh/irm-iex installer to bootrstrap itself.

Gankra:
This also improves dist-manifest.json to include CI commands in its output, making it more like terraform

fixes #233

### DIFF
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -45,11 +45,12 @@ on:
       - '*-?v[0-9]+*'
 
 jobs:
-  # Create the Github Release™ so the packages have something to be uploaded to
   create-release:
+    # Create the Github Release™ so the packages have something to be uploaded to
     runs-on: ubuntu-latest
     outputs:
       has-releases: ${{ steps.create-release.outputs.has-releases }}
+      ci: ${{ steps.create-release.outputs.matrix }}
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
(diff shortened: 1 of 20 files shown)


Gankra:
Gonna be hammering https://github.com/Gankra/cargodisttest with this for a while to test/experiment.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -110,39 +99,56 @@ jobs:
...
+        with:
+          name: artifacts
+          path: ${{ steps.cargo-dist.outputs.paths }}
 
-  # Mark the Github Release™ as a non-draft now that everything has succeeded!
   publish-release:
-    # Only run after all the other tasks, but it's ok if upload-artifacts was skipped
-    needs: [create-release, upload-artifacts]
-    if: ${{ always() && needs.create-release.result == 'success' && (needs.upload-artifacts.result == 'skipped' || needs.upload-artifacts.result == 'success') }}
+    # Create a Github Release with all the results once everything is done,
```
mistydemeo, line 127: No trademark? 😆

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -110,39 +99,56 @@ jobs:
...
+    # regardless of success. The release is only marked as non-draft if all tasks succeeded.
+    needs:
+      - plan
+      - upload-dist-artifacts
+    if: ${{ always() }}
```
mistydemeo, line 132: Curious what "if always" is doing here.

mistydemeo (approved):
Looking good to me.

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: As soon as you want to permit some tasks to run if previous ones fail/cancel, you need to sprinkle this around a lot (can be useful for the publish-github-release step to make drafts still happen, although for now i've given up on pulling that functionality forward).

Gankra:
Closing in favour of #378

PR #331, opened 2023-08-15
docs(book): remove some repetition

tshepang:


Gankra (approved):
thanks as always tshepang :)

PR #333, opened 2023-08-15
feat(global-last): split out global task and have it fetch local results

Gankra:


mistydemeo (approved):
Looks good to me. Incorporated it into the Homebrew PR and it's working.

Issue #180, opened 2023-03-02
Support semi-unified versions as a release workflow


Gankra:
I think cargo-dist will largely already Do The Right Thing for this workflow but we should test/document it properly.

This is a workflow @epage described, and is used by [clap](https://github.com/clap-rs/clap). Essentially it's the same as "all versions are bumped in lockstep" *but you don't bother bumping a version for a library if it didn't change*. This is something that would work perfectly for cargo-dist and many of the workspaces I use fully-unified versions in.

Here's what it would look like in cargo-dist:

Since everything in the workspace goes into cargo-dist, every git tag is definitely bumping the version of that crate. However, cargo-dist-schema is updated much less often, so currently we do a lot of "nothing changed" releases for no real reason other than having the versions exactly the same.

However we could instead just say: yeah cargo-dist 0.2.0 uses cargo-dist-schema 0.1.0, because nothing changed in the schema in that version jump. But if 0.3.0 does need a schema change (say a new field is added), then we cut 0.3.0 for *both* crates, skipping over 0.2.0 for cargo-dist-schema.

This is also a lot friendlier to people depending on the libraries, as version resolution will be less finicky. Note that cargo-release already will complain to us when we needlessly cut a new cargo-dist-schema, so there's tooling to help you select the new release. I need to look into the exact cargo-release config that works best for this.

Note however that cargo-dist will probably still have a few "useless" releases due to prerelease/bootstrapping shenanigans.

ashleygwilliams:
i have been wondering about this workflow for axohtml - tbh i think it makes a lot of sense and preserves information about whether there was a change which feels like a positive

Gankra:
tentatively adding this to 0.1.0 as a nice-to-have

PR #334, opened 2023-08-17
0.2.0-prerelease.1

mistydemeo:
I haven't published this as a tag yet, but prepping so we can test the new feature while working on followups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,8 @@
 # Unreleased
 
-Nothing Yet!
-
-# Version 0.2.0-prerelease.1 (2023-08-17)
-
 This release includes preliminary support for creating Homebrew packages on
-macOS.
+macOS. It also allows you to specify `--features` your application should be
+built with for production releases.
 
 ## Features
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 
 Nothing Yet!
 
+# Version 0.2.0-prerelease.1 (2023-08-17)
```
Gankra, line 5: If you remove this heading (and the "Nothing Yet!"), cargo-dist will detect that the "Unreleased" heading applies to the prerelease, so we don't need to churn the heading all the time.

Gankra (approved):
gonna push the changes since I gotta add my own entry

cargo-dist/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cargo-dist"
 description = "Shippable application packaging for Rust"
-version = "0.1.0"
+version = "0.2.0-prerelease.1"
```
Gankra, line 4: cargo-dist-schema and cargo-dist are always versioned in lockstep (this is why i like using `cargo release` to automate that).

Gankra: (fixing this would be #180)

PR #339, opened 2023-08-17
feat(installers): Commit Homebrew formula to tap in CI

mistydemeo:
This is the first pass at automatically committing and publishing a Homebrew formula to a tap the user specifies.

The user can specify the tap in their `Cargo.toml`, alongside specifying the installers and their other metadata. We use this to add a new entry to the GitHub Actions YAML which handles publishing all releases contained within this run. We use the existing, already-provisioned `GITHUB_TOKEN` to both clone and push to that repository.

Sample output:

```yml
  upload-homebrew-formula:
    needs: upload-global-artifacts
    runs-on: "ubuntu-20.04"
    env:
      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      RELEASES: ${{ needs.upload-global-artifacts.outputs.releases }}
      GITHUB_USER: "axo bot"
      GITHUB_EMAIL: "admin+bot@axo.dev"
    steps:
      - uses: actions/checkout@v3
        with:
          repository: "mistydemeo/homebrew-cargodisttest"
          token: ${{ secrets.HOMEBREW_TAP_TOKEN }}
      # So we have access to the formula
      - name: Fetch local artifacts
        run: |
          gh release download ${{ github.ref_name }} --dir Formula --repo ${GITHUB_REPOSITORY} --clobber
      - name: Commit formula files
        run: |
          git config --global user.name "${GITHUB_USER}"
          git config --global user.email "${GITHUB_EMAIL}"

          for release in $(echo $RELEASES | jq --compact-output '.[]'); do
            name=$(echo $release | jq .app_name --raw-output)
            version=$(echo $release | jq .app_version --raw-output)

            git add Formula/${name}.rb
            git commit -m "${name} ${version}"
          done
          git push
```


cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +167,34 @@ jobs:
           cat uploads.txt
           gh release upload ${{ github.ref_name }} $(cat uploads.txt)
           echo "uploaded!"
+
+  upload-homebrew-formula:
+    needs: upload-local-artifacts
+    if: {{{ tap }}}
```
mistydemeo, line 173: Right now we inline this instead of trying to read it dynamically from config.

mistydemeo:
I've done some testing. Thus far, it seems to be working except that the token we have isn't scoped to allow pushing to other repos the user has access to: https://github.com/mistydemeo/cargodisttest/actions/runs/5897060553/job/15996111408

It looks like the token we get automatically won't have write permissions to repos other than the one the action was run in. If we want to be able to publish to external repos, we'll need to get our users to provide us with a token they've provisioned themselves.

Gankra:
> Thus far, it seems to be working except that the token we have isn't scoped to allow pushing to other repos the user has access to

Alas, I was hoping you were right and the token would be way overpowered.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +168,35 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          repository: {{{ tap }}}
+      # So we have access to the formula
+      - name: Fetch local artifacts
+        run: |
+          gh release download ${{ github.ref_name }} --dir Formula --repo ${GITHUB_REPOSITORY}
+      - name: Commit formula files
+        run: |
+          for release in $(echo $RELEASES | jq --compact-output '.[]'); do
```
Gankra, line 189: should $RELEASES be in quotes?

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +168,35 @@ jobs:
...
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      RELEASES: ${{ needs.upload-local-artifacts.outputs.releases }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: {{{ tap }}}
+      # So we have access to the formula
+      - name: Fetch local artifacts
+        run: |
+          gh release download ${{ github.ref_name }} --dir Formula --repo ${GITHUB_REPOSITORY}
```
Gankra, line 186: Is there a reason to explicitly specify repo?

OH. I see we're running the CI in the target repo, that makes sense.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +168,35 @@ jobs:
...
+      - name: Commit formula files
+        run: |
+          for release in $(echo $RELEASES | jq --compact-output '.[]'); do
+            name=$(echo $release | jq .app_name)
+            version=$(echo $release | jq.app_version)
```
Gankra, line 191:
```suggestion
            version=$(echo $release | jq .app_version)
```

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +168,35 @@ jobs:
           cat uploads.txt
           gh release upload ${{ github.ref_name }} $(cat uploads.txt)
           echo "uploaded!"
+
+{{%- if tap %}}
+  upload-homebrew-formula:
+    needs: upload-local-artifacts
```
Gankra, line 174:
```suggestion
    needs: upload-global-artifacts
```

right?

cargo-dist/src/init.rs
```diff
@@ -598,6 +600,13 @@ fn update_toml_metadata(
         installers.as_ref(),
     );
 
+    apply_optional_value(
+        table,
+        "tap",
+        "# A GitHub repo to push the Homebrew formula to\n",
```
Gankra, line 606:
```suggestion
        "# A GitHub repo to push Homebrew formulas to\n",
```

(pedantically more accurate I guess)

Gankra (commented):
Things we also need to do:

* Duplicate the `init` logic that "prompts you for an npm scope if you have the npm installer enabled" but for homebrew taps.

* Write docs explaining how to:
    * (optional) create the repo (I know it can be empty, but good to say that)
    * get a key that can edit the repo
    * add the key to your SECRETS (let's call it `secrets.HOMEBREW_TOKEN` or `secrets.TAP_TOKEN`)
    * presumably it's not desirable to mess with the automatic `GITHUB_TOKEN`?

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: Hm, good question.

mistydemeo:
> Duplicate the init logic that "prompts you for an npm scope if you have the npm installer enabled" but for homebrew taps.

Added this message in the installer:

```
✔ To automatically push updates to the Homebrew
    installer to a tap (repository), you must provision a GitHub token and
    expose it as a secret named HOMEBREW_TAP_TOKEN in GitHub Actions.
    For more information, see the documentation.
```

Though we can also prompt it to add the `tap` field too, now I think about it.

I'll get the docs updated.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -167,6 +170,42 @@ jobs:
...
+        run: |
+          git config --global user.name "${GITHUB_USER}"
+          git config --global user.email "${GITHUB_EMAIL}"
+
+          for release in $(echo $RELEASES | jq --compact-output '.[]'); do
+            name=$(echo $release | jq .app_name --raw-output)
+            version=$(echo $release | jq .app_version --raw-output)
+
+            git add Formula/${name}.rb
+            git commit -m "${name} ${version}"
```
Gankra, line 203: just confirming: tagging would be useless since every commit to this repo is effectively a tag, yeah?

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -26,6 +26,8 @@ pub struct HomebrewInstallerInfo {
     pub homepage: Option<String>,
     /// A brief description of the application
     pub desc: Option<String>,
+    /// A GitHub repository to write the formula to, in owner/name format
+    pub tap: Option<String>,
```
Gankra, line 30: Does homebrew only support github taps, or is there also an extended format we should (one day) support?

cargo-dist/src/init.rs
```diff
@@ -439,6 +440,21 @@ fn get_new_dist_metadata(
...
+        .installers
+        .as_deref()
+        .unwrap_or_default()
+        .contains(&InstallerStyle::Homebrew)
+    {
+        eprintln!(
+            r#"{check} To automatically push updates to the Homebrew
+    installer to a tap (repository), you must provision a GitHub token and
+    expose it as a secret named HOMEBREW_TAP_TOKEN in GitHub Actions.
+    For more information, see the documentation."#
```
Gankra, line 454: probably worth adding a line with the URL to the docs here

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -135,6 +135,8 @@ jobs:
   upload-global-artifacts:
     needs: upload-local-artifacts
     runs-on: {{{ global_task.runner }}}
+    outputs:
+      releases: ${{ steps.upload-global-artifacts.outputs.releases }}
```
Gankra, line 139: Arguably this should be being done on the create-release task, but this is fine for now.

Gankra (approved):
looks solid

cargo-dist/templates/ci/github_ci.yml.j2
ashleygwilliams: this is an interesting comment- while i agree every commit is effectively a tag- since our other tooling is using the github releases endpoint- is there a universe where we would want to consume info from the tap repo using that api? if so, it may be useful to tag?

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: The default `brew tap` and `brew install` syntaxes uses GitHub, but you can use non-GitHub git repos by specifying a full URL to the `brew tap` command.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: From a Homebrew perspective, there's no expectation of tagging within a tap. `brew update` just updates to the latest git tip, and there isn't a mechanism to use tags for anything.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: Easy change to make!

PR #297, opened 2023-08-01
migrate everything to jinja2

Gankra:
fixes #283

PR #342, opened 2023-08-17
fix(ci): fix a typo in rustup update

mistydemeo:
This looks like a possible typo; it was introduced in #297. Fixes a parse error I ran into when `cargo init`ing from a fresh repo using this prerelease version.


Gankra:
!#$%? i swear i fixed this in #336 but it vanished

PR #348, opened 2023-08-21
fix(homebrew): add explicit version tag

mistydemeo:
A real-world test indicated that Homebrew is reliably misparsing the version from the URLs. Instead of letting Homebrew guess, let's just be explicit.


ashleygwilliams (approved):
🚀

PR #352, opened 2023-08-22
chore: update CI to latest cargo-dist

mistydemeo:
This newer version will let us test pushing to our Homebrew tap.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -162,6 +162,7 @@ jobs:
           cat uploads.txt
           gh release upload ${{ github.ref_name }} $(cat uploads.txt)
           echo "uploaded!"
+
   upload-homebrew-formula:
     needs: [create-release, upload-global-artifacts]
     runs-on: "ubuntu-20.04"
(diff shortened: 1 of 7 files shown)


mistydemeo:
Decided to update this with the new stuff to publish.

@Gankra I regenerated our `release.yml`; is it safe to include these changes right now? I assume yes.

.github/workflows/release.yml
```diff
@@ -132,11 +130,74 @@ jobs:
...
+          echo "dist ran successfully"
+          cat dist-manifest.json
+
+          # Parse out what we just built and upload it to the Github Release™
+          jq --raw-output ".artifacts[]?.path | select( . != null )" dist-manifest.json > uploads.txt
+          echo "uploading..."
+          cat uploads.txt
+          gh release upload ${{ github.ref_name }} $(cat uploads.txt)
+          echo "uploaded!"
+  upload-homebrew-formula:
```
Gankra, line 165: ...we should really fix this lack of newline in the template (you need to put the newline *within* the if so that it doesn't get stripped. learned this the hard way.).

Gankra (approved):

.github/workflows/release.yml
mistydemeo: Agggh. Yeah that explains it. Let me fix that

PR #357, opened 2023-08-23
feat: restore release table headings, add platform

### LINKED ISSUE #356: reintroduce column headers, add human-readable platform column
in 0.1.0 the download artifacts table lost it's headers due them being identified, reasonably, as redundant. however github markdown doesn't allow tables w/o headers so the outcome is that the table looks "broken":
Screenshot 2023-08-23 at 4 53 32 PM

i propose:
1. adding a new "platform" column that gives a "human readable" version of the target triple
2. re-add the headers, given the additional utility they provide for said platform column

added benefit: it's symmetrical with how we handle this in oranda:
Screenshot 2023-08-23 at 4 53 42 PM

mistydemeo:
This brings back column headings that had been removed in 0.1.0, and adds an extra platform column. ~~Right now those are displayed as triples rather than as the human-formatted ones that oranda uses.~~ They're now printed as pretty human-readable names like oranda!

Fixes #356.


mistydemeo:
Switching to draft while it's using a prerelease axoproject.

cargo-dist/src/tasks.rs
```diff
@@ -1808,8 +1810,8 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             if !other_artifacts.is_empty() && download_url.is_some() {
                 let download_url = download_url.as_ref().unwrap();
                 writeln!(gh_body, "## Download {heading_suffix}\n",).unwrap();
-                gh_body.push_str("|        |        |\n");
-                gh_body.push_str("|--------|--------|\n");
+                gh_body.push_str("|  File  | Platform | Checksum |\n");
+                gh_body.push_str("|--------|----------|----------|\n");
```
Gankra, line 1814: huh interesting, I apparently didn't make this code tolerant of the fact that checksums can be disabled. Whelp, pre-existing issue.

Gankra (approved):
nice and simple

PR #359, opened 2023-08-24
feat: add publish_jobs config field

mistydemeo:
This is a quick spike of the new `publish_jobs` config field. Now, instead of just using `tap` to decide whether to enable a Homebrew publish job, we also have this new field. Right now, we don't explicitly ask the user to enable it and instead implicitly enable it if `tap` is enabled, but we can update the `init` UI for that if we want.

### DIFF
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -221,6 +221,13 @@ pub struct DistMetadata {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "all-features")]
     pub all_features: Option<bool>,
+
+    /// Publish jobs to run in CI
+    ///
+    /// (defaults to none)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "publish-jobs")]
+    pub publish_jobs: Option<Vec<PublishStyle>>,
 }
 
 impl DistMetadata {
@@ -248,6 +255,7 @@ impl DistMetadata {
             features: _,
             default_features: _,
             all_features: _,
+            publish_jobs: _,
         } = self;
         if let Some(include) = include {
             for include in include {
@@ -284,6 +292,7 @@ impl DistMetadata {
             features,
             default_features,
             all_features,
+            publish_jobs,
         } = self;
 
         // Check for global settings on local packages
@@ -345,6 +354,9 @@ impl DistMetadata {
         }
         if tap.is_none() {
             *tap = workspace_config.tap.clone();
+        }
+        if publish_jobs.is_none() {
+            *publish_jobs = workspace_config.publish_jobs.clone();
         }
 
         // This was historically implemented as extend, but I'm not convinced the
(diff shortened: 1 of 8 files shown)


cargo-dist/src/config.rs
```diff
@@ -221,6 +221,11 @@ pub struct DistMetadata {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "all-features")]
     pub all_features: Option<bool>,
+
+    /// Publish jobs to run in CI
+    ///
+    /// (defaults to none)
+    pub publish_jobs: Option<Vec<PublishStyle>>,
```
Gankra, line 228: `#[serde(rename)]` to kebab case for consistency

cargo-dist/src/config.rs
```diff
@@ -346,6 +353,9 @@ impl DistMetadata {
         if tap.is_none() {
             *tap = workspace_config.tap.clone();
         }
+        if publish_jobs.is_none() {
+            *publish_jobs = workspace_config.publish_jobs.clone();
+        }
```
Gankra, line 358: should be a workspace-only field, i think...?

hmm. actually I guess we could make the CI intelligently only publish certain apps to certain venues? if like one only makes sense on pypi or w/e. So I guess package-granularity makes sense.

cargo-dist/src/init.rs
```diff
@@ -475,6 +479,8 @@ fn get_new_dist_metadata(
                 meta.tap = None;
             } else {
                 meta.tap = Some(tap.to_owned());
+                meta.publish_jobs = Some(vec![PublishStyle::Homebrew]);
```
Gankra, line 482: We should write this more compositionally. Start with the existing inputs read from the config, then push this value if it's not already in the array (will make things a lot better when we start adding more).

cargo-dist/src/init.rs
```diff
@@ -759,6 +766,13 @@ fn update_toml_metadata(
         *all_features,
     );
 
+    apply_string_list(
+        table,
+        "publish_jobs",
```
Gankra, line 771: This will need to be renamed too

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -170,7 +170,7 @@ jobs:
           gh release upload ${{ github.ref_name }} $(cat uploads.txt)
           echo "uploaded!"
 
-{{%- if tap %}}
+{{%- if 'homebrew' in publish_jobs and tap %}}
```
Gankra, line 173: Somewhere in the rust code should probably error out if the first part is true but the second part isn't. The "npm has to have tar.gz" checks might be a good example to go off of..?

Gankra (changes requested):

cargo-dist/src/init.rs
Gankra: in retrospect this one is ignorable if we just want to get it out the door

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: also ignorable if we just want it out the door

cargo-dist/src/config.rs
mistydemeo: Yeah, it seems like there's room for some extra detail there.

Gankra (approved):

cargo-dist/src/init.rs
mistydemeo: Done.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: We don't have an error case for this in npm right now except in `init`, and the Homebrew `init` prevents the case from existing. I did add a warning for if it comes up during rendering though, same as with npm.

PR #370, opened 2023-08-28
feat(msi): add msi installer

Gankra:
The "msi" installer vendors the binaries into a windows msi.
msi generation is implemented by using cargo-wix as a library,
which in turn handles generating templates and invoking WiX (v3).

The resulting msi's are unsigned -- signing will be handled by a followup,
as changes to OV certs mean that "proper" signing is now very complicated
and messy.

This implementation has some notable details:

main.wxs generation
--------------------

WiX requires an xml-based template called main.wxs for each msi it generates.
This template includes things like the application name, version, publisher,
EULA, embedded files, and so on.

cargo-wix's default workflow is for you to run `cargo wix init` once to generate
that file and check it in. At this point the developer is free to hand-edit
the template to suit their needs. We like this idea, `cargo dist generate-ci`
is based on that same premise!

But our experience with generate-ci has also taught us that this approach
is really prone to frustrating desyncs. The main.wxs bakes in several pieces
of information that were sourced from your Cargo.toml. As a result, if you
ever change your Cargo.toml, you now need to remember to apply the same
change to your main.wxs, or rerun `cargo wix init` and potentially clobber
any hand-edits you made.

As such, for the first draft of this feature we've opted for a more aggressive
solution: whenever you run `cargo dist init` we will invoke `cargo wix init`
to force the template to have up to date contents, essentially forbidding
hand-edits.

When you run `cargo dist build` or `cargo dist plan`
we will also error out if we detect that `cargo wix init` would modify
main.wxs.

A new `allow-dirty = ["msi"]` can be added to your cargo-dist config to
prevent `init` from regenerating main.wxs or checking that it's up to date,
essentially restoring the original design of cargo-wix. We believe this
should be opt-in because most people ideally shouldn't want to do hand-edits,
and desyncing is a very nasty footgun.

GUID persistence
----------------

`cargo dist init` will also modify your Cargo.tomls to add
`[package.metadata.wix]` with two generated GUIDs (if those GUIDs aren't
already present).

These GUIDs are used by windows to identify that two different msis actually
intend to refer to the same application and install dir. As such, each release
of your app needs to use the same GUID.

cargo-wix bakes these GUIDs into your main.wxs, and critically it defaults to
generating new random ones every time you run init (because constantly
re-initing wasn't part of the design).

By persisting the GUIDs to your package's Cargo.toml, we unlock the ability
to freely regenerate main.wxs whenever we want.

### DIFF
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -46,7 +46,7 @@ pub enum DistError {
     /// Error from (cargo-)wix
     #[error("WiX returned an error while building {msi}")]
     Wix {
-        /// The MSI we were trying to build
+        /// The msi we were trying to build
         msi: String,
         /// The underyling wix error
         #[source]
@@ -202,22 +202,22 @@ pub enum DistError {
         /// The problematic mode
         generate_mode: crate::config::GenerateMode,
     },
-    /// MSI with too many packages
+    /// msi with too many packages
     #[error("{artifact_name} depends on multiple packages, which isn't yet supported")]
     #[diagnostic(help("depends on {spec1} and {spec2}"))]
-    MultiPackageMSI {
-        /// Name of the MSI
+    MultiPackageMsi {
+        /// Name of the msi
         artifact_name: String,
         /// One of the pacakges
         spec1: String,
         /// A different package
         spec2: String,
     },
-    /// MSI with too few packages
+    /// msi with too few packages
     #[error("{artifact_name} has no binaries")]
     #[diagnostic(help("This should be impossible, you did nothing wrong, please file an issue!"))]
-    NoPackageMSI {
-        /// Name of the MSI
+    NoPackageMsi {
+        /// Name of the msi
         artifact_name: String,
     },
     /// These GUIDs for msi's are required and enforced by `cargo dist generate --check`
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -43,6 +43,26 @@ pub enum DistError {
         details: minijinja::Error,
     },
 
+    /// Error from (cargo-)wix
+    #[error("WiX returned an error while building {msi}")]
+    Wix {
+        /// The MSI we were trying to build
+        msi: String,
+        /// The underyling wix error
+        #[source]
+        details: wix::Error,
+    },
+
+    /// Error from (cargo-)wix init
+    #[error("Couldn't generate main.wxs for {package}'s msi installer")]
+    WixInit {
+        /// The package
+        package: String,
+        /// The underlying wix error
+        #[source]
+        details: wix::Error,
+    },
+
     /// Error parsing metadata in Cargo.toml (json because it's from cargo-metadata)
     #[error("Malformed metadata.dist in {manifest_path}")]
     CargoTomlParse {
@@ -114,7 +134,7 @@ pub enum DistError {
 
     /// Use explicitly requested workspace builds, but had packages with custom feature settings
     #[error("precise-builds = false was set, but some packages have custom build features, making it impossible")]
-    #[help("these packages customized either features, no-default-features, or all-features: {packages:?}")]
+    #[diagnostic(help("these packages customized either features, no-default-features, or all-features: {packages:?}"))]
     PreciseImpossible {
         /// names of problem packages
         packages: Vec<String>,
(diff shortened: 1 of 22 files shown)


cargo-dist/src/errors.rs
```diff
@@ -114,11 +133,29 @@ pub enum DistError {
...
+        /// Name of the MSI
+        artifact_name: String,
+        /// One of the pacakges
+        spec1: String,
+        /// A different package
+        spec2: String,
+    },
+    /// MSI with too few packages
+    #[error("The MSI {artifact_name} has no binaries associated with this")]
+    #[diagnostic(help("This should be impossible, you did nothing wrong, please file an issue!"))]
```
mistydemeo, line 154: Love "you did nothing wrong" as a user message. I want more software to tell me that.

cargo-dist/src/tasks.rs
```diff
@@ -1002,18 +1008,36 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
     }
 
+    /// Register that `for_artifact` requires `binary_idx` to actually be built for
+    /// `for_variant`.
+    ///
+    /// `dest_path` is an optional file path to copy the binary to (used for Archives)
+    /// as soon as they're built.
+    ///
+    /// Note that it's dangerous to not use `dest_path`, as cargo does not guarantee that
+    /// multiple invocations will not overwrite eachother's outputs. Since we always
```
mistydemeo, line 1018:
```suggestion
    /// multiple invocations will not overwrite each other's outputs. Since we always
```

cargo-dist/src/tasks.rs
```diff
@@ -1511,12 +1543,86 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            let variant_id = &variant.id;
+            let artifact_name = format!("{variant_id}.msi");
+            let artifact_path = self.inner.dist_dir.join(&artifact_name);
+
+            // Compute which package we're actually building, based on the binaries
+            let mut package_spec: Option<String> = None;
+            for &binary_idx in &binaries {
+                let binary = self.binary(binary_idx);
+                if let Some(existing_spec) = &package_spec {
+                    // cargo-wix doesn't clearly support multi-package, so bail
```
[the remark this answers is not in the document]
Gankra: Up until this point I've vaguely managed to keep cargo-dist's code permissive of the future possibility of a Release containing multiple cargo packages, but it's increasingly difficult to keep up the charade.

parasyte:
Is there a mode of operation where the platform-specific metadata for the package (`main.wxs` for instance) is recreated on every package build? Something should be capable of creating these kinds of things from a separate source of truth which is, with best effort, shared across all platforms. (I am indifferent to the choice between using cargo metadata or a separate TOML file for the metadata that is only applicable to one specific platform. Though I suspect a separate TOML will be nicer to maintain over time.)

The reason I ask is because it seems like this would be more robust than attempting to merge hand edits to an XML file committed to the repo. And more robust to modifications of the source of truth. So long as the XML can be reproduced deterministically, I don't see any reason at all to commit it.

Perhaps `cargo-wix` is missing some capabilities required to make `main.wxs` transient, and I could see this as an argument against the idea. But I believe it is right path forward. Personally, I am not building a Windows app (I'm building a platform-independent app) and I don't want to maintain Windows-specific details, if I can avoid it.

Gankra:
> Is there a mode of operation where the platform-specific metadata for the package (main.wxs for instance) is recreated on every package build

Yes, that's the plan. Some minor upstream work is required to not *at all* check in the file, but I'm personally biased towards  checking it in to make things more... continuous... for lack of a better word. With #381 we'll basically be discouraging hand-editing the file, requiring you to opt into being allowed to. The implementation in this PR currently persists all the necessary state to your Cargo.toml to be able to regenerate it reproducibly every time.

At very least this MVP impl will probably require checking it in, and we might make it totally hidden in future releases when we've had more time to work the details out.

Gankra:
The code is ready for review but landing is blocked on https://github.com/volks73/cargo-wix/issues/219

Gankra:
depends on https://github.com/volks73/cargo-wix/pull/222

mistydemeo (approved):
Looks good, but a few comments:

* Can we add `cargo-dist/wix` to our own `.gitignore` so it doesn't get tracked in git if we're not committing it right now?
* The `generate` command didn't learn `msi` as a subcommand like it learned `ci`.

PR #378, opened 2023-08-29
ci++

Gankra:


mistydemeo:
> I broke the publish-homebrew-formula task
> it's expecting steps.plan.outputs.releases which doesn't exist anymore
> should be changed to some use of fromJson(needs.plan.outputs.val)

Fixed.

> I haven't tested that publish-release still works (moderately confident it works)

It took a few tweaks to get this in good shape, but it's now working, including both artifacts and Homebrew. https://github.com/mistydemeo/cargodisttest/releases/tag/v0.1.31

> The rebase was hastily done, it's very possible I accidentally dropped some new improvements to the current github_ci.yml so we should both do a detailed test/review of this to be sure

I believe this is fine now.

mistydemeo:
> What security measures if any exist to prevent a user from editing the workflow in their PR to remove this checks that this is a PR?

It's not perfect, but a maintainer would have to grant permission to the PR author to let it run.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -85,7 +87,7 @@ jobs:
   upload-local-artifacts:
     # Let the initial task tell us to not run (currently very blunt)
     needs: plan
-    if: ${{ fromJson(needs.plan.outputs.val).releases != null }}
+    if: ${{ fromJson(needs.plan.outputs.val).releases != null && fromJson(needs.plan.outputs.val).ci.github.pr_run_mode == "upload" }}
```
Gankra: this isn't quite right, this will make pr_run_mode = "skip" prevent releases altogether. needs to be combine with whether we're releasing for realz.

mistydemeo (approved):

PR #381, opened 2023-08-29
feat: generate++

### LINKED ISSUE #371: generate++
Improve the `generate` system as follows:

##  Replace `generate-ci` with a `generate` command that takes a list of `GenerateStyle`

* "ci" being the one and only GenerateStyle
* in the future "wix" will be added to this list by #370 
* if no GenerateStyles are passed, the command is in "all mode", meaning "all GenerateStyles that this workspace targets"
  * Similar concept already exists for places you can pass InstallerStyles, CiStyles, and Targets
  * As with those, we should arguably also clamp the GenerateStyle lists to the "all" set 
* (Optional) keep a generate-ci alias for back-compat

## Add some mode flags to `generate`

* `--check` changes the default behaviour from "write to disk" to "error if writing to disk would change the file"
* (Optional) `--print` flag that just prints what would be written to disk 
   * with `==== package/main.wxs ====` headings or something?

## Add `allow-dirty: Option<Vec<GenerateStyle>>` to config

* Tells `generate` "all mode" to skip that style
* This implies `generate --check` won't complain about it
* Unsure: can non-all-mode override allow-dirty?
* Usure: Should there be a flag to disable allow-dirty?
    
## Make `generate --check` block most other commands

* `manifest`, `plan`, and `check` should all run `generate --check` on their DistGraphs before proceeding
* (obviously not the full CLI command but the subroutine that picks up from the DistGraph)

mistydemeo:
Closes #371.

This version covers steps 1 and 2, but not 3 and 4 yet.


cargo-dist/src/errors.rs
```diff
@@ -159,6 +159,13 @@ pub enum DistError {
         #[source]
         details: semver::Error,
     },
+
+    /// Not an error; indicates that a file's contents differ via --check
```
Gankra, line 163: wdym not an error?

cargo-dist/src/cli.rs
```diff
@@ -233,8 +236,33 @@ pub struct InitArgs {
...
+    pub check: bool,
+}
+
+#[derive(Args, Clone, Debug)]
+pub struct GenerateCiArgs {
+    /// Check if the generated output differs from on-disk config without writing it
+    #[clap(long)]
+    #[clap(default_value_t = false)]
+    pub check: bool,
+}
```
Gankra, line 265: wait... OH this is your preserving the alias (no idea if there's a less painful way to do it, so respect)

cargo-dist/src/main.rs
```diff
@@ -220,7 +221,25 @@ fn cmd_init(cli: &Cli, args: &InitArgs) -> Result<(), miette::Report> {
     do_init(&config, &args)
 }
 
-fn cmd_generate_ci(cli: &Cli, _args: &GenerateCiArgs) -> Result<(), miette::Report> {
+fn cmd_generate(cli: &Cli, args: &GenerateArgs) -> Result<(), miette::Report> {
+    let modes = if args.mode.is_empty() {
+        vec![GenerateMode::Ci]
+    } else {
+        args.mode.clone()
+    };
```
Gankra, line 229: I'd prefer that this be resolved deeper in the code. `generate`'s library impl should be reading the full config (probably just building a DistGraph) and determining which generate backends are in the "universe" (e.g. "they have ci=github, so I want to generate github ci", "they have installer=msi, so i want to generate wix files", ...).

If the input is an empty array, that's basically saying "no filters, generate everything". If the input isn't empty, that says "do the intersection of everything and this list".

cargo-dist/src/main.rs
```diff
@@ -230,7 +249,7 @@ fn cmd_generate_ci(cli: &Cli, _args: &GenerateCiArgs) -> Result<(), miette::Repo
         installers: cli.installer.iter().map(|ins| ins.to_lib()).collect(),
         announcement_tag: cli.tag.clone(),
     };
-    let args = cargo_dist::GenerateCiArgs {};
+    let args = cargo_dist::GenerateCiArgs { check: args.check };
     do_generate_ci(&config, &args)
 }
```
Gankra, line 254: Similarly it would be nice if this was calling `do_generate(&[GenerateStyle::Ci], args.check)` so that the alias boils away immediately.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -34,19 +37,43 @@ struct CiTask {
...
+///
+/// This checks whether the new configuration differs from the config on disk
+/// writhout actually writing the result.
+pub fn check_github_ci(dist: &DistGraph) -> Result<bool, miette::Report> {
+    let ci_file = github_ci_path(dist);
+
+    let rendered = generate_github_ci(dist)?;
+    // FIXME: should we catch all errors, or only LocalAssetNotFound?
+    let existing = LocalAsset::load_string(ci_file).unwrap_or("".to_owned());
+    Ok(rendered == existing)
```
Gankra, line 66: Hmm, feels like this probably wants to be just returning `Result<(), ...>` and producing the DistError::CheckFileMismatch itself, so that paths/details can be given?

Gankra (changes requested):

cargo-dist/src/cli.rs
```diff
@@ -233,8 +236,33 @@ pub struct InitArgs {
     pub with_json_config: Option<Utf8PathBuf>,
 }
 
+/// Which style(s) of configuration to generate
+#[derive(ValueEnum, Copy, Clone, Debug)]
+pub enum GenerateMode {
+    /// Generate all configuration types (currently: ci)
+    All,
```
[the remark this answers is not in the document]
mistydemeo: Yeah, that's fair. I'll probably switch to just having it be implicit if nothing's specified.

cargo-dist/src/cli.rs
mistydemeo: Haha, yeah, that's right.

mistydemeo:
Updated to add steps 3 and 4 - `allow-dirty` field added to config, added automatic checking during `dist` and manifest commands, and exercised `allow-dirty` in tests.

Gankra (approved):

PR #388, opened 2023-08-31
remove stray char

tshepang:


mistydemeo (approved):
Good catch, thank you!

PR #391, opened 2023-09-01
feat(generate): prepare generate for msi

Gankra:
This reworks the generate design to work better for slotting in msi stuff.

* allow-dirty takes a list of GenerateMode instead of CiStyle
* all commands with integrity checks defer to generate --check
* a global cli --allow-dirty flag was added for cli-tests
* a DirtyMode type was added to represent the cli flag's changes to logic
* integration-tests which call generate were changed to do that first so build_and_plan succeeds without needing to allow-dirty
* computed allow_dirty conclusions are no longer stored within GithubCi, and instead are just used by the top-level orchestration
* error reporting was made a bit better to help debug what changed (done for my own confused debugging of my own mistakes)
* allow-dirty is now always respected, even if explicit args passed to generate contradict it (this is what I planned but I 100% get the argument for the previous design, should be discussed).
* some minor refactors/cleanups that will be useful once msi stuff is integrated


Gankra:
completes my vision of #371

Gankra:
oh, I also forgot that I removed allow_dirty from dist-manifest, as it didn't seem like something that needs to be passed between tasks.

Gankra:
PR run demonstrating a pull-request failing because I hand-edited the Cargo.toml and didn't rerun init: https://github.com/oranda-gallery/cargodisttest/pull/3

Gankra:
Note: this is now rebased on top of #391

Gankra:
Also blocked on https://github.com/volks73/cargo-wix/pull/212

Gankra:
oh lol, those last two comments were supposed to be on the MSI PR (deleted the comments).

Gankra:
oh noooo

Gankra:
rebasing...

cargo-dist/src/lib.rs
```diff
@@ -609,11 +594,19 @@ fn do_generate_preflight_checks(dist: &DistGraph) -> Result<()> {
...
     let dist = gather_work(cfg)?;
-    do_generate_preflight_checks(&dist)?;
+
+    run_generate(&dist, &args.modes, args.check)?;
+
+    Ok(())
+}
+
+/// The inner impl of do_generate
+pub fn run_generate(dist: &DistGraph, modes: &[GenerateMode], check: bool) -> Result<()> {
```
mistydemeo, line 604: Don't know how I feel about the last arg just being a boolean - the other calls to `run_generate()` aren't really clear. I tend to find positional booleans illegible.

mistydemeo (approved):
> allow-dirty is now always respected, even if explicit args passed to generate contradict it (this is what I planned but I 100% get the argument for the previous design, should be discussed).

Not sure how I feel about that - it feels to me like if we have a conflict in requested behaviour, we should at least notify the user about it.

Gankra:
> Not sure how I feel about that - it feels to me like if we have a conflict in requested behaviour, we should at least notify the user about it.

I agree, it should just be an error.

cargo-dist/src/lib.rs
Gankra: you're not wrong, i should add an enum, then?

cargo-dist/src/errors.rs
```diff
@@ -173,6 +173,15 @@ pub enum DistError {
         /// The line number
         line_number: usize,
     },
+
+    /// Not an error; indicates that a file's contents differ via --check
```
mistydemeo, line 177: Wrong docstring?

mistydemeo (approved):

PR #392, opened 2023-09-01
feat(ci): cleanup default/init logic for pr_run_mode

Gankra:
* make plan the default (as originally agreed?)
* make the init code actually remember what you had set before
* make the init code not write a value if it's the default
* make the init code produce compile errors if a new PrRunMode is added
* rework the prompt order/messages to be clearer


Gankra:
Before:
image


After:
image

mistydemeo (approved):

PR #401, opened 2023-09-05
feat(ci): allow disabling publishing prereleases

### LINKED ISSUE #398: Homebrew taps update with prerelease versions
Currently, we push every tagged release to a Homebrew tap. For example: https://github.com/axodotdev/homebrew-tap/commit/8f5ef81fc87944d7c1eb83b037b3d40e71646450

While this is consistent with how we always push to releases, we're able to mark an entire GitHub release as a prerelease in a way we don't do to the Homebrew formula. We should likely either stop pushing prereleases to the tap, or add a configuration flag as to whether to push prereleases to package managers.

mistydemeo:
We currently unconditionally publish prereleases to package managers. We should instead disable that by default while optionally enabling it.

I've left it out of `init` because I see this as more of a power user feature; I've documented it but don't think we should prompt most users about it.

Fixes #398.

### DIFF
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -243,6 +243,13 @@ pub struct DistMetadata {
     #[serde(rename = "publish-jobs")]
     pub publish_jobs: Option<Vec<PublishStyle>>,
 
+    /// Whether to publish prereleases to package managers
+    ///
+    /// (defaults to false)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "publish-prereleases")]
+    pub publish_prereleases: Option<bool>,
+
     /// Whether we should create the Github Release for you when you push a tag.
     ///
     /// If true (default), cargo-dist will create a new Github Release and generate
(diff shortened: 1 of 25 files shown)


cargo-dist/src/config.rs
```diff
@@ -398,6 +407,9 @@ impl DistMetadata {
         if publish_jobs.is_none() {
             *publish_jobs = workspace_config.publish_jobs.clone();
         }
+        if publish_prereleases.is_none() {
+            *publish_prereleases = workspace_config.publish_prereleases;
+        }
```
Gankra, line 412: We've really gotta stop merging things into packages if we're not gonna respect it (we only respect the workspace setting).

Gankra (approved):
Seems reasonable.

cargo-dist/src/config.rs
mistydemeo: Should I just remove this?

mistydemeo:
Just tested, confirmed it works.

With this new option omitted: https://github.com/mistydemeo/cargodisttest/actions/runs/6102078852

With it on: https://github.com/mistydemeo/cargodisttest/actions/runs/6102303747

cargo-dist/src/config.rs
Gankra: change it to the other pattern where we warn that we're ignoring it

PR #408, opened 2023-09-07
fix: prevent append-release from running on PRs

Gankra:


mistydemeo (approved):

Gankra:
Kinda just have to merge this and see if it fixes it (the triggering workflow is on main rn)

PR #410, opened 2023-09-07
fix(check): analysis uses --artifacts=all

mistydemeo:
Fixes an issue where the preflight config check would use whichever artifact options the user had passed. If --artifacts=local was passed, we would be inappropriately flagging GitHub CI as out of date because we'd try to omit the global artifacts.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -23,7 +23,9 @@ use backend::{
 };
 use camino::{Utf8Path, Utf8PathBuf};
 use cargo_dist_schema::{Asset, AssetKind, DistManifest, ExecutableAsset};
-use config::{ChecksumStyle, CompressionImpl, Config, DirtyMode, GenerateMode, ZipStyle};
+use config::{
+    ArtifactMode, ChecksumStyle, CompressionImpl, Config, DirtyMode, GenerateMode, ZipStyle,
+};
 use semver::Version;
 use tracing::{info, warn};
 
@@ -42,8 +44,9 @@ mod tests;
 
 /// cargo dist build -- actually build binaries and installers!
 pub fn do_build(cfg: &Config) -> Result<DistManifest> {
+    check_integrity(cfg)?;
+
     let dist = tasks::gather_work(cfg)?;
-    check_integrity(&dist)?;
 
     // FIXME: parallelize this by working this like a dependency graph, so we can start
     // bundling up an executable the moment it's built! Note however that you shouldn't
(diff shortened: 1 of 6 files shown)


Gankra (changes requested):
more lines to make sure aren't double-printed

Gankra:
ugh whoops

cargo-dist/tests/snapshots/error_manifest.snap
```diff
@@ -6,7 +6,10 @@ stdout:
 {"diagnostic": {"message": "This workspace doesn't have anything for cargo-dist to Release!","severity": "error","causes": [],"labels": [],"related": []}}
 
 stderr:
-analyzing workspace:
+  cargo-dist
+    [bin] cargo-dist
+  cargo-dist-schema (no binaries)
+
```
Gankra, line 12: these lines are still getting double-printed. it's also an interesting point that we're not forwarding --tag in this... probably right on balance

cargo-dist/src/lib.rs
```diff
@@ -661,9 +664,23 @@ pub fn run_generate(dist: &DistGraph, args: &GenerateArgs) -> Result<()> {
 /// Run any necessary integrity checks for "primary" commands like build/plan
 ///
 /// (This is currently equivalent to `cargo dist generate --check`)
-pub fn check_integrity(dist: &DistGraph) -> Result<()> {
+pub fn check_integrity(cfg: &Config) -> Result<()> {
+    // We need to avoid overwriting any parts of configuration from CLI here,
+    // so construct a clean copy of config to run the check generate
+    let check_config = Config {
+        needs_coherent_announcement_tag: true,
```
Gankra, line 671:
```suggestion
        needs_coherent_announcement_tag: false,
```

I was like "there was a weird thing about bypassing tags wasn't there..." and aha, here it is indeed. Setting this to false is basically "i work like init/generate/plan".

cargo-dist/tests/snapshots/error_manifest.snap
```diff
@@ -6,11 +6,6 @@ stdout:
 {"diagnostic": {"message": "This workspace doesn't have anything for cargo-dist to Release!","severity": "error","causes": [],"labels": [],"related": []}}
 
 stderr:
-analyzing workspace:
-  cargo-dist (didn't match tag v1.0.0-FAKEVERSION)
```
mistydemeo, line 10: Maybe we still want to print this one?

Gankra: If we want to preserve it in normal runs then we probably need to do the "store it and have the caller choose to print it" thing (you can `writeln!` to a String, I'm sure there's an example somewhere in this codebase...)

mistydemeo: We don't actually pass the info this one line is keyed off of during the `check` runs, so we could just print it to stderr.

Gankra: There's several different conditions though, some of them static (publish=false, dist=false...)

mistydemeo: Hmm, right. Well, in the interest of getting this fix out, let's just leave it like this for now.

Gankra (approved):

PR #415, opened 2023-09-07
fix(generate): provide a warning if something is skipped

mistydemeo:
In testing, I found it too easy to miss that `generate` was skipping generation; a printed warning makes that clearer.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -630,12 +630,16 @@ pub fn run_generate(dist: &DistGraph, args: &GenerateArgs) -> Result<()> {
     } else {
         // Check that we're not being told to do a contradiction
         for &mode in &args.modes {
-            if !dist.allow_dirty.should_run(mode)
-                && matches!(dist.allow_dirty, DirtyMode::AllowList(..))
-            {
-                return Err(DistError::ContradictoryGenerateModes {
-                    generate_mode: mode,
-                })?;
+            if !dist.allow_dirty.should_run(mode) {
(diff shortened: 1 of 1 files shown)


Gankra (changes requested):
i'm very confused by this implementation, which seems to make it so:

* if you run `cargo dist generate ci --allow-dirty` we always produce an error (reasonable, it's a no-op)
* if you run `cargo dist generate --allow-dirty` it's still a no-op but we don't complain
* if you run `cargo dist generate ci` with `allow-dirty = ["ci"]` we will warn
* if you run `cargo dist generate` with `allow-dirty = ["ci"]` we don't say anything

Presumably if you want to warn about skips you want to do it in the next loop (and only do it if there was actually something selected to be generated..?)

Gankra:
ohh i guess there's complication around --check vs not and commands implicitly invoking this..?

mistydemeo:
> ohh i guess there's complication around --check vs not and commands implicitly invoking this..?

Yeahhhh, that's right.

> if you run cargo dist generate ci with allow-dirty = ["ci"] we will warn
> if you run cargo dist generate with allow-dirty = ["ci"] we don't say anything

Now that I think about it this really should be backwards, even though I did put it in the explicitly-specified mode. Handling it on the inferred side probably makes more sense rather than just being silent.

ashleygwilliams:
there's some conflicts here and the longer we wait the more we'll run into- what are we thinking about this PR generally?

mistydemeo:
I don't really care as much about this now. If we get complaints from users we can revisit.

PR #417, opened 2023-09-08
feat(ci): add support for user-defined jobs

### LINKED ISSUE #379: release lifecycle plugins
This is the followup and ultimate payoff of #378 

Expose/upgrade `publish-jobs = [...]` and `upload-jobs = [...]` so that users can refer to their own .yml files that cargo-dist embeds into release.yml and weaves into the dependency graph. This could be used for:

* Adding custom publish steps ("[publish to this other package manager](https://github.com/axodotdev/cargo-dist/blob/e01b5e390e3f10cc6ff1812066b64c96e3a8e3ca/.github/workflows/publish-release.yml)", "post to social media", ...)
* Adding ("[generate a schema file](https://github.com/axodotdev/cargo-dist/blob/e01b5e390e3f10cc6ff1812066b64c96e3a8e3ca/.github/workflows/append-release.yml)", "build this other installer", ...)
* Post-processing artifacts ("sign my binaries", "test the binaries", ...)

mistydemeo:
This adds support for user-defined jobs in CI. Currently, we support publish jobs specifically, and these run parallel with other publish jobs. These are specified in TOML via the `user_publish_jobs` hash:

```toml
user_publish_jobs = { github = ["./.github/workflows/something.yml"] }
```

Fixes #379.

### DIFF
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -523,19 +523,48 @@ impl std::fmt::Display for InstallerStyle {
 }
 
 /// The publish jobs we should run
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Serialize, PartialEq, Eq)]
 pub enum PublishStyle {
     /// Publish a Homebrew formula to a tap repository
     #[serde(rename = "homebrew")]
     Homebrew,
+    /// User-supplied value
+    User(String),
+}
+
+impl std::str::FromStr for PublishStyle {
+    type Err = DistError;
+    fn from_str(s: &str) -> DistResult<Self> {
+        if let Some(slug) = s.strip_prefix("./") {
+            Ok(Self::User(slug.to_owned()))
+        } else if s == "homebrew" {
+            Ok(Self::Homebrew)
+        } else {
+            Err(DistError::UnrecognizedStyle {
+                style: s.to_owned(),
+            })
+        }
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for PublishStyle {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+
+        let path = String::deserialize(deserializer)?;
+        path.parse().map_err(|e| D::Error::custom(format!("{e}")))
+    }
 }
 
 impl std::fmt::Display for PublishStyle {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let string = match self {
-            PublishStyle::Homebrew => "homebrew",
-        };
-        string.fmt(f)
+        match self {
+            PublishStyle::Homebrew => write!(f, "homebrew"),
+            PublishStyle::User(s) => write!(f, "./{s}"),
+        }
     }
 }
 
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -229,6 +229,13 @@ pub enum DistError {
         /// The missing keys
         keys: &'static [&'static str],
     },
+    /// unrecognized style
+    #[error("{style} is not a recognized value")]
+    #[diagnostic(help("Jobs that do not come with cargo-dist should be prefixed with ./"))]
+    UnrecognizedStyle {
+        /// Name of the msi
+        style: String,
+    },
 }
 
 impl From<minijinja::Error> for DistError {
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -201,6 +201,8 @@ pub struct DistGraph {
     pub ci: CiInfo,
     /// List of publish jobs to run
     pub publish_jobs: Vec<PublishStyle>,
+    /// Extra user-specified publish jobs to run
+    pub user_publish_jobs: Vec<String>,
     /// Whether to publish prerelease builds to package managers
     pub publish_prereleases: bool,
     /// A GitHub repo to publish the Homebrew formula to
@@ -705,7 +707,27 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         };
 
         let templates = Templates::new()?;
-        let publish_jobs = workspace_metadata.publish_jobs.clone().unwrap_or(vec![]);
+        let publish_jobs: Vec<PublishStyle>;
+        let user_publish_jobs: Vec<PublishStyle>;
+        (publish_jobs, user_publish_jobs) = workspace_metadata
+            .publish_jobs
+            .clone()
+            .unwrap_or(vec![])
+            .into_iter()
+            .partition(|s| !matches!(s, PublishStyle::User(_)));
+        let user_publish_jobs = user_publish_jobs
+            .into_iter()
+            // Remove the ./ suffix for later; we only have user jobs at this
+            // point so we no longer need to distinguish
+            .map(|s| {
+                let string = s.to_string();
+                if let Some(stripped) = string.strip_prefix("./") {
+                    stripped.to_owned()
+                } else {
+                    string
+                }
+            })
+            .collect();
         let publish_prereleases = publish_prereleases.unwrap_or(false);
 
         let allow_dirty = if allow_all_dirty {
@@ -745,6 +767,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
                 pr_run_mode: workspace_metadata.pr_run_mode.unwrap_or_default(),
                 tap: workspace_metadata.tap.clone(),
                 publish_jobs,
+                user_publish_jobs,
                 publish_prereleases,
                 allow_dirty,
             },
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -290,6 +290,17 @@ jobs:
 
 {{%- endif %}}
 
+{{%- for job in user_publish_jobs %}}
+
+  custom-{{{ job|safe }}}:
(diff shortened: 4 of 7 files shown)


cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -229,6 +229,13 @@ jobs:
 
 {{%- endif %}}
 
+{{%- for job in user_publish_jobs %}}
+  user-publish-jobs:
```
mistydemeo, line 233: This naming is temporary - obviously this won't work in a loop with more than one thing in it. Need to decide how we want these jobs named.

Gankra: This is part of why it seemed reasonable to me for this to be slug-driven (gonna elaborate in a top-level comment).

Gankra:
For discussion, writing out the slug-driven approach I had pictured:

```
publish-jobs = ["itch"]
```

Would tell us that:

* When building github CI, we expect to find a file called `.github/workflows/publish-itch.yml`
* We generate a step called `publish-itch` in `.github/workflows/release.yml` containing that fragment

With a hypothetical buildkite workflow (i have no idea what buildkite looks like, let's pretend it's vaguely github-shaped):

* When building buildkite CI, we expect to find a file called `.buildkite/pipelines/publish-itch.yml`
* We generate a step called `publish-itch` in `.buildkite/pipelines/release.yml` containing the fragment

----

I also figured it would be more ergonomic to have builtins and user-defined jobs sharing the publish-jobs / upload-jobs array, but I respect wanting to make them separate arrays.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -229,6 +229,13 @@ jobs:
...
+    needs: [plan, should-publish]
+    if: ${{ !fromJson(needs.plan.outputs.val).announcement_is_prerelease || fromJson(needs.plan.outputs.val).publish_prereleases }}
+    uses: {{{ job }}}
```
Gankra, line 236: Oh wait, you want to invoke them? I'm not sure if that will be pleasant (passing stuff between workflow files is a huge pain). I was assuming cargo-dist would load the file and paste its contents into here. Maybe I'm in the wrong though.

Gankra: hmm, i suppose if we just hand them `needs.plan.outputs.val` that's all the input they really need..?

mistydemeo: Yeah, that gives an input we've already committed to keeping stable(ish) so I think it's a reasonable approach.

Gankra: worth a try

mistydemeo:
Hm, that slug approach seems reasonable. While it commits us to a specific directory structure in the target repo, that's not really a bad thing.

I do still want to keep it separated out by CI system. I can see a scenario where someone has more than one configured in the same repo and doesn't have 1:1 job parity between systems. Even if it requires more work from the user, I think it's better to specify extra jobs per CI system.

Gankra:
Can you envision an approach where the per-ci-system stuff can be sugarred away for people who don't need it?

mistydemeo:
Do you think that's necessary? Specifying a hash isn't enough extra work for the user that supporting a sugar syntax in addition to the other one feels like it's a bit more complex than we want to commit to.

Gankra:
tbh I'll be surprised if any user ever needs that level of generality, so I respect the desire to have it available but it seems like an ergonomics pain to force everyone to refer to the possibility but never use it.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: This seems happy; I did some tests with this and it's working fine. I added secret inheritance as well, ensuring that user-defined workflows can make use of any secrets in the repository.

mistydemeo:
I got talked into the simpler list syntax in place of the more complex hash syntax; pushed a version with that. Think this is ready now, save docs.

mistydemeo:
The most recent version places both types of publish jobs in a single array. We distinguish between types via prefixing user jobs with a `@`. For example:

```toml
publish-jobs = ["homebrew", "@greeter"]
```

`homebrew` is a cargo-dist builtin, and `@greeter` is mapped to `.github/workflows/publish-greeter.yml`.

cargo-dist/src/errors.rs
```diff
@@ -229,6 +229,13 @@ pub enum DistError {
         /// The missing keys
         keys: &'static [&'static str],
     },
+    /// unrecognized style
+    #[error("{style} is not a recognized value")]
+    #[diagnostic(help("Styles that do not come with cargo-dist should be prefixed with an @"))]
```
Gankra, line 234: error message now out of date, also "Styles" is too vague here. publish-jobs or jobs or... something more precise

cargo-dist/src/tasks.rs
```diff
@@ -705,7 +707,27 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+        let user_publish_jobs: Vec<PublishStyle>;
+        (publish_jobs, user_publish_jobs) = workspace_metadata
+            .publish_jobs
+            .clone()
+            .unwrap_or(vec![])
+            .into_iter()
+            .partition(|s| !matches!(s, PublishStyle::User(_)));
+        let user_publish_jobs = user_publish_jobs
+            .into_iter()
+            // Remove the @ suffix for later; we only have user jobs at this
```
Gankra, line 720: comment out of date

cargo-dist/src/config.rs
```diff
@@ -523,19 +523,48 @@ impl std::fmt::Display for InstallerStyle {
...
     Homebrew,
+    /// User-supplied value
+    User(String),
+}
+
+impl std::str::FromStr for PublishStyle {
+    type Err = DistError;
+    fn from_str(s: &str) -> DistResult<Self> {
+        if s.starts_with("./") {
+            Ok(Self::User(s.strip_prefix("./").unwrap().to_owned()))
```
Gankra, line 539:
```suggestion
        if let Some(slug) = s.strip_prefix("./") {
            Ok(Self::User(slug.to_owned()))
```

cargo-dist/src/tasks.rs
```diff
@@ -705,7 +707,27 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            // point so we no longer need to distinguish
+            .map(|s| {
+                let string = s.to_string();
+                if let Some(stripped) = string.strip_prefix("./") {
+                    stripped.to_owned()
+                } else {
+                    string
+                }
```
Gankra, line 728: the ./ is stripped by this point in the code, so no need to map

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -290,6 +290,17 @@ jobs:
 
 {{%- endif %}}
 
+{{%- for job in user_publish_jobs %}}
+
+  user-publish-job-{{{ job|safe }}}:
```
Gankra, line 295: The full `user-publish-job-` prefix seems a bit cumbersome, `publish-{slug}` too dangerous for your tastes? Or maybe just `custom-{slug}` since the user will likely have names like `publish-npm.yml`?

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -290,6 +290,17 @@ jobs:
...
+    needs: [plan, should-publish]
+    if: ${{ !fromJson(needs.plan.outputs.val).announcement_is_prerelease || fromJson(needs.plan.outputs.val).publish_prereleases }}
+    uses: ./.github/workflows/{{{ job|safe }}}.yml
+    with:
+      plan: ${{ needs.plan.outputs.val }}
+    secrets: inherit
```
Gankra, line 301: woah, TIL

cargo-dist/tests/integration-tests.rs
```diff
@@ -194,6 +194,37 @@ path-guid = "BFD25009-65A4-4D1E-97F1-0030465D90D6"
...
+fn axolotlsay_user_publish_job() -> Result<(), miette::Report> {
+    let test_name = _function_name!();
+    AXOLOTLSAY.run_test(|ctx| {
+        let dist_version = ctx.tools.cargo_dist.version().unwrap();
+        ctx.patch_cargo_toml(format!(r#"
+[workspace.metadata.dist]
+cargo-dist-version = "{dist_version}"
+installers = ["shell", "powershell", "homebrew", "npm"]
+tap = "axodotdev/homebrew-packages"
+publish-jobs = ["homebrew", "./custom-task"]
```
Gankra, line 207: how hard would it be to make this n=2 to make sure the loop expands pretty and there's no weird conflicts?

cargo-dist/src/tasks.rs
mistydemeo: Not true, the `publish_jobs` vec we're fetching from has the original prefixed names.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: `custom-` sounds good to me.

cargo-dist/tests/integration-tests.rs
mistydemeo: Good call, made it two custom tasks.

cargo-dist/src/tasks.rs
Gankra: image

both versions of this code were stripping it, how are we ending up with multiple copies of the strip being needed?

mistydemeo: Good question! I'm... not really sure? I checked and it's absolutely true that the `FromStr` is constructing the `PublishStyle` with no `./` prefix, while the type in the task *does* have it until I strip it here.

PR #418, opened 2023-09-08
fix: don't forward WiX output

Gankra:
it writes to stdout which messes up --output-format=json


Gankra:
example failure: https://github.com/oranda-gallery/cargodisttest/actions/runs/6116688741/job/16602267777?pr=4

mistydemeo (approved):

Issue #422, opened 2023-09-12
Feature: linkage checker


mistydemeo:
While many Rust binaries only use other Rust code, some projects need to link against external libraries - whether dynamically or statically. Worse, sometimes projects will *unknowingly* link against a non-default library in the system. This can lead to binaries that aren't actually usable by end users who are missing that library, or unreliable behaviour.

I'm planning to add a new "linkage checker" step to the build which reports which libraries the package built against us. This will be divided into system-provided libraries (those all users of the same OS version should have) and package manager-provided libraries (which a user would have to choose to install in order to be present). This will let us do a few different things:

* Provide greater insight into dependencies for developers who are less familiar with that;
* Warn about possible compatibility issues ahead of time;
* Warn about libraries we think developers should avoid dynamically linking against (#290);
* Record a set of dependencies that we can inject into package manager manifests, allowing them to be automatically installed for the user by Homebrew/apt/&c.

PR #426, opened 2023-09-14
feat: add a linkage report subcommand

mistydemeo:
This adds a new linkage checker subcommand, making it easier for us to determine what a packages' binaries have linked against. It can output in both human-readable and JSON forms, making it useful as a CI step and as input to other commands.

The human-readable output looks like this:

```
cargo-dist (aarch64-apple-darwin):

System: /usr/lib/libiconv.2.dylib /usr/lib/libSystem.B.dylib
Homebrew:
Public (unmanaged):
Frameworks: /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation /System/Library/Frameworks/Security.framework/Versions/A/Security
Other:
```

The JSON output looks like this:

```json
[
  {
    "binary": "akextract",
    "target": "aarch64-apple-darwin",
    "system": [
      "/usr/lib/libiconv.2.dylib",
      "/usr/lib/libSystem.B.dylib"
    ],
    "homebrew": [],
    "public_unmanaged": [],
    "other": [],
    "frameworks": []
  },
  {
    "binary": "akmetadata",
    "target": "aarch64-apple-darwin",
    "system": [
      "/usr/lib/libiconv.2.dylib",
      "/usr/lib/libSystem.B.dylib"
    ],
    "homebrew": [],
    "public_unmanaged": [],
    "other": [],
    "frameworks": []
  },
  {
    "binary": "akrepack",
    "target": "aarch64-apple-darwin",
    "system": [
      "/usr/lib/libiconv.2.dylib",
      "/usr/lib/libSystem.B.dylib"
    ],
    "homebrew": [],
    "public_unmanaged": [],
    "other": [],
    "frameworks": []
  }
]
```

The current version only implements macOS, but I'll be adding Linux support soon. Windows support will require more research.

refs #422.

### DIFF
--- a/book/src/ci/github.md
+++ b/book/src/ci/github.md
@@ -163,7 +163,7 @@ By default cargo-dist breaks build tasks onto more machines than strictly necess
 
 Although most Rust builds are statically linked and contain their own Rust dependencies, some crates will end up dynamically linking against system libraries. It's useful to know what your software picked up&mdash;sometimes this will help you catch things you may not have intended, like dynamically linking to OpenSSL, or allow you to check for package manager-provided libraries your users will need to have installed in order to be able to run your software.
 
-cargo-dist provides a linkage report during your CI build in order to allow you to check for this. For macOS and Linux, it's able to categorize the targets it linked against to help you gauge whether or not it's likely to cause problems for your users. To view this, check the detailed view of your CI build and consult the "Linkage report" step from the `upload-local artifacts` jobs.
+cargo-dist provides a linkage report during your CI build in order to allow you to check for this. For macOS and Linux, it's able to categorize the targets it linked against to help you gauge whether or not it's likely to cause problems for your users. To view this, check the detailed view of your CI build and consult the "Build" step from the `upload-local artifacts` jobs.
 
 This feature is defined for advanced users; most users won't need to use it. It's most useful for developers with specialized build setups who want to ensure that their binaries will be safe for all of their users. A few examples of users who may need to use it:
 
--- a/book/src/ci/github.md
+++ b/book/src/ci/github.md
@@ -157,6 +157,57 @@ By default cargo-dist breaks build tasks onto more machines than strictly necess
 
 
 
+### Checking what your build linked against
+
+> since 0.4.0
+
+Although most Rust builds are statically linked and contain their own Rust dependencies, some crates will end up dynamically linking against system libraries. It's useful to know what your software picked up&mdash;sometimes this will help you catch things you may not have intended, like dynamically linking to OpenSSL, or allow you to check for package manager-provided libraries your users will need to have installed in order to be able to run your software.
+
+cargo-dist provides a linkage report during your CI build in order to allow you to check for this. For macOS and Linux, it's able to categorize the targets it linked against to help you gauge whether or not it's likely to cause problems for your users. To view this, check the detailed view of your CI build and consult the "Linkage report" step from the `upload-local artifacts` jobs.
+
+This feature is defined for advanced users; most users won't need to use it. It's most useful for developers with specialized build setups who want to ensure that their binaries will be safe for all of their users. A few examples of users who may need to use it:
+
+* Users with custom runners with extra packages installed beyond what's included in the operating system;
+* Users who have installed extra packages using cargo-dist's system dependency feature;
+* Users whose cargo buildsystems include extra C dependencies.
+
+The report is divided into categories to help you make sense of where these libraries are from and what it might mean for your users. These categories are:
+
+* System: Libraries that come with your operating system. On Linux, these packages are all provided by the system's package manager, and the linkage report includes information about which package includes each library. Some of these packages will be included in the base OS, and will be safe to rely on, while you'll need to ensure your users have others. If you're using standard base images like GitHub Actions's and haven't installed additional packages using apt, the packages in this list should be preinstalled for your users. On macOS, these packages are shipped with the operating system and not managed by a package manager; you can always rely on these being there within the same version of macOS.
+* Homebrew (macOS only): Libraries that are provided by the Homebrew package manager for macOS. These packages are not installed by default, so your users will need to have them installed in order to be able to use your software.
+* Public (unmanaged): Libraries which are present in public locations, but which are not managed or provided by the system or a package manager. Because these are not standard parts of the operating system, your users will be unlikely to have them.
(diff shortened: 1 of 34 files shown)


book/src/ci/github.md
```diff
@@ -157,6 +157,51 @@ By default cargo-dist breaks build tasks onto more machines than strictly necess
 
 
 
+### Checking what your build linked against
+
+> since 0.4.0
+
+Although most Rust builds are statically linked and contain their own Rust dependencies, some crates will end up dynamically linking against system libraries. It's useful to know what your software picked up&mdash;sometimes this will help you catch things you may not have intended, like dynamically linking to OpenSSL, or allow you to check for package manager-provided libraries your users will need to have installed in order to be able to run your software.
```
Gankra, line 164: emdashes want whitespace around them, right?

book/src/ci/github.md
```diff
@@ -157,6 +157,51 @@ By default cargo-dist breaks build tasks onto more machines than strictly necess
...
+
+cargo-dist provides a linkage report during your CI build in order to allow you to check for this. For macOS and Linux, it's able to categorize the targets it linked against to help you gauge whether or not it's likely to cause problems for your users. To view this, check the detailed view of your CI build and consult the "Linkage report" step from the `upload-local artifacts` jobs.
+
+The report is divided into categories to help you make sense of where these libraries are from and what it might mean for your users. These categories are:
+
+* System: Libraries that come with your operating system. These are usually safe to link against, but you should make sure that they come from a package your users will have. On Linux, these packages are all provided by the system's package manager, and the linkage report includes information about which package includes one of these libraries. On macOS, these packages are shipped with the operating system and not managed by a package manager.
```
Gankra, line 170: > Libraries that come with your operating system. These are usually safe to link against, but you should make sure that they come from a package your users will have.

Hmm, this seems kinda-incorrect (we want those things in system-dependencies, right?) and unactionable (how would I know if my users will have it?)

book/src/ci/github.md
```diff
@@ -157,6 +157,51 @@ By default cargo-dist breaks build tasks onto more machines than strictly necess
...
+* Homebrew (macOS only): Libraries that are provided by the Homebrew package manager for macOS. These packages are not installed by default, so your users will need to have them installed in order to be able to use your software.
+* Public (unmanaged): Libraries which are present in public locations, but which are not managed or provided by the system or a package manager. Because these are not standard parts of the operating system, your users will be unlikely to have them.
+* Frameworks (macOS only): Frameworks, a special type of library provided by macOS. Frameworks installed in the `/System` directory come with the operating system and are available to all users.
```
Gankra, line 173: Why don't /System frameworks show up under "System"?

mistydemeo: Good question. In my head they're a "special" enough kind of library that it's worth separating out, but that may be a bit too specific as Mac thing.

book/src/ci/github.md
mistydemeo: Depends on our style guide, I guess. I usually don't, but happy to add spaces.

cargo-dist/src/cli.rs
```diff
@@ -128,6 +128,9 @@ pub enum Commands {
     #[clap(disable_version_flag = true)]
     #[clap(hide = true)]
     GenerateCi(GenerateCiArgs),
+    /// Report on the dynamic libraries used by the built artifacts.
+    #[clap(disable_version_flag = true)]
+    Linkage(LinkageArgs),
```
Gankra, line 133: Ignoring the detail of human-readable output, is there any reason not to run the linkage checker internally on every run of `cargo dist build`?

Gankra (approved):

cargo-dist/src/lib.rs
```diff
@@ -117,6 +117,13 @@ fn build_manifest(cfg: &Config, dist: &DistGraph) -> DistManifest {
         })
     }
 
+    let linkage = fetch_linkage(
+        cfg.targets.clone(),
+        dist.artifacts.clone(),
+        dist.dist_dir.clone(),
+    )?;
+    let linkage = linkage.iter().map(|l| l.to_schema()).collect();
```
Gankra, line 125: hmm interesting, this doesn't make `cargo dist plan` explode?

cargo-dist/src/lib.rs
```diff
@@ -761,6 +772,22 @@ pub fn do_linkage(cfg: &Config, args: &LinkageArgs) -> Result<()> {
...
+
+/// Determinage dynamic linkage of built artifacts (impl of `cargo dist linkage`)
+pub fn do_linkage(cfg: &Config, args: &LinkageArgs) -> Result<()> {
+    let dist = gather_work(cfg)?;
+
+    let reports: Vec<Linkage> = if let Some(target) = args.from_json.clone() {
+        let json = std::fs::read_to_string(target)
+            .into_diagnostic()
+            .wrap_err("Failed to parse {target}")?;
+        serde_json::from_str(&json).unwrap()
```
Gankra, line 786: this should be using axoasset's builtin json support

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -143,9 +143,8 @@ jobs:
           echo "EOF" >> "$GITHUB_OUTPUT"
       - name: "Linkage report"
         run: |
-          cargo dist linkage ${{ matrix.dist_args }} --print-output --print-json > linkage.json
-
-          echo "linkage=$(cat linkage.json)" >> "${GITHUB_OUTPUT}"
+          jq .linkage dist-manifest.json > linkage.json
+          cargo dist linkage ${{ matrix.dist_args }} --from-json=linkage.json --print-output
```
Gankra, line 147: oh interesting. at this point do we just want `cargo dist build` to have a flag to include linkage info in the human readable output?

Gankra (approved):
not 100% sold on from_json but general approach seems good

cargo-dist/src/lib.rs
mistydemeo: If there are no files to check, `fetch_linkage` is a no-op. No error.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: Yeah, that's definitely an option! So long as it doesn't get buried in the rest of the output, I guess.

Issue #423, opened 2023-09-12
Feature: specify system packages


mistydemeo:
Right now, users can specify *cargo* dependencies for their package via `Cargo.toml`, but we don't provide a way for users to specify *system* packages that need to be present during the build. We should provide a way for developers to specify system packages that will be present on the builders during their build.

My current plan is to add a new `system-dependencies` key to `Cargo.toml`. This allows the user to specify the dependencies *per package manager*. When we run a build on a system that uses that package manager, we install those dependencies prior to running the build. It looks something like this:

```toml
system-dependencies = { apt = ["libcue-dev"], homebrew = ["libcue"] }
```

*Some* of these dependencies will end up being required at runtime for users of the software, but not all of them. Some of these dependencies will be build tools such as `cmake`, which are totally unnecessary once the software is built, while others will be libraries that will be dynamically linked against and required by the resulting binary. We should leverage #422 to determine which packages are actually needed at runtime, and inject those packages as dependencies of our package manager manifests such as the Homebrew formula.

For Homebrew builds, I plan to use `Brewfile`s and [`brew bundle exec`](https://www.mistys-internet.website/blog/blog/2020/11/23/bundle-install-with-homebrew-magic-using-brew-bundle-exec/). This will allow us to ensure that the build is able to find Homebrew-provided dependencies even if they end up outside cargo's default search paths with the least fuss.

FreeMasen:
With #428 merged, is this now available? The `mdbook` states that `workspace.metadata.dist.dependencies.<pkg-mgr>.<dep>` should work, is that the case today?

I ask because I am looking to add a step like this in a CI job but adding these keys and running `cargo dist genrate` doesn't seem to update `.github/workflows/release.yml`

mistydemeo:
We haven't yet released 0.4.0, which is the release that will include this feature. While we may have a prerelease out before then, it'll be a little longer until it's included in a stable release.

FreeMasen:
Thank you for the reply. I didn't realize that the `since <version>` header could include not-yet released versions and didn't look closely enough. Sorry for the extra noise here.

PR #428, opened 2023-09-14
feat: add support for installing system dependencies

mistydemeo:
This is a first pass at support for installing system dependencies (#423). It supports Linux (apt) and macOS (Homebrew). Users specify their deps via a new key in `Cargo.toml`, like so:

```toml
system-dependencies = { homebrew = ["libcue"], apt = ["libcue-dev"] }
```

There's a new "install dependencies" build step that comes prior to "build" within the same job, which is a noop if there are no deps to install. For apt, we simply `apt-get install` the requested deps. For Homebrew, we generate a `Brewfile` with the requested deps and then install them.

I believe I may be injecting too much work into GitHub CI here; when we end up with support for additional CI systems, that would require duplicating far too much work. I think it'll be a better approach to back this by a new `cargo dist` subcommand which takes care of more of the work for us, giving us something smaller we can inline into CI config.

An issue this reveals which I hadn't considered: GitHub only provide Intel macOS runners. They don't provide ARM runners at all. While we can cross-compile native Rust just fine, we're not in a position where we can cross-compile with Apple Silicon Homebrew-provided libraries. That may be an acceptable limitation though - we certainly can't use a package manager that doesn't exist on the runner we have.


mistydemeo:
About this in the new build section:

```bash
CARGO="cargo"
if [[ -f "Brewfile" ]]; then
    CARGO="brew bundle exec -- cargo"
fi
```

If we're installing packages for Homebrew, we generate a `Brewfile` containing the dependencies. The `brew bundle exec` command injects Homebrew's build environment into `cargo dist build`, ensuring that the resulting build can see all of the dependencies even if they're not present in `/usr/local`.

mistydemeo:
I've updated the syntax. It's cargo-like, and now looks like this:

```toml
[workspace.metadata.dist.dependencies.homebrew]
cmake = '*'
libcue = { version = "2.2.1" }

[workspace.metadata.dist.dependencies.apt]
cmake = '*'
libcue-dev = { version = "2.2.1-3build3" }
```

Unwanted package managers can be omitted. The '*' version specifier means "anything". Versions are ignored by Homebrew and will stay ignored; they're temporarily ignored by chocolatey until I hook up the proper version specs to the installer. In the future, I'll have the Homebrew installer at least check if the version you actually received seems to match the version you specified, given that you can't tell Homebrew what to actually provide.

Sample Linux build using versions - it failed just because the version I asked for isn't present in this Linux distro. https://github.com/mistydemeo/cargodisttest/actions/runs/6268604459/job/17023895071

We have a few other fields we're interested in adding, though we should discuss whether they need to make the initial implementation or not. Shouldn't take too long, though. They are:

* `targets`: Limit the package install to a list of one or more target triples. If unspecified, it'll be installed on everything. Useful if, for example, you only want to install something on Intel or ARM macOS.
* `phase`: Specify whether the package is only needed at buildtime, or whether it's needed at runtime. If the latter is specified, the dependency specification will be injected into package definitions like the Homebrew formula. We plan to auto-detect runtime usage via #422, so most users won't need to use this. The default will be buildtime only.

cargo-dist/src/config.rs
```diff
@@ -814,6 +826,129 @@ impl std::fmt::Display for GenerateMode {
     }
 }
 
+/// Packages to install before build from the system package manager
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct SystemDependencies {
+    /// Packages to install in Homebrew
+    #[serde(default = "BTreeMap::new")]
```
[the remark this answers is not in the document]
Gankra: Also iirc we have a SortedMap alias somewhere in the codebase to make it easy to swap out BTreeMap if we need (and make intent clear).

cargo-dist/src/tasks.rs
```diff
@@ -736,6 +741,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
 
+        let system_dependencies =
+            workspace_metadata
+                .system_dependencies
+                .clone()
+                .unwrap_or(SystemDependencies {
+                    homebrew: BTreeMap::new(),
+                    apt: BTreeMap::new(),
+                    chocolatey: BTreeMap::new(),
+                });
```
Gankra, line 752: should ideally be unwrap_or_default

cargo-dist/src/tasks.rs
```diff
@@ -736,6 +741,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             DirtyMode::AllowList(allow_dirty.clone().unwrap_or(vec![]))
         };
 
+        let system_dependencies =
+            workspace_metadata
+                .system_dependencies
```
Gankra, line 746: hmm wait are the package-local dependencies being ignored? it makes sense that we'd want to gather a global final list to emit in the CI, but I would think it would want to be "union together all dependencies for each package we've decided we're building right now")

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -115,15 +115,24 @@ jobs:
...
         # to "real" actions without writing to env-vars, and writing to env-vars has
         # inconsistent syntax between shell and powershell. cargo-dist and jq work fine
         # in powershell.
         shell: bash
         run: |
+          CARGO="cargo"
+          if [[ -f "Brewfile" ]]; then
+            CARGO="brew bundle exec -- cargo"
+          fi
+
```
Gankra, line 133: Oh yeah, didn't we agree it should be possible/ideal for these expressions to be something that `cargo dist build` does for you..?

cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -30,6 +30,11 @@ class {{ formula_class }} < Formula
   {%- if license %}
   license "{{ license }}"
   {%- endif %}
+  {%- if dependencies|length > 0 %}
+  {% for dep in dependencies %}
+  depends_on "{{ dep }}"
+  {%- endfor %}
+  {%- endif %}
```
Gankra, line 37: hmm what does the if-guard eliminate here? An extra newline?

cargo-dist/tests/integration-tests.rs
```diff
@@ -225,6 +225,84 @@ scope = "@axodotdev"
     })
 }
 
+#[test]
+fn axolotlsay_homebrew_runtime_dep() -> Result<(), miette::Report> {
```
Gankra, line 229: skimming over these

cargo-dist/src/config.rs
```diff
@@ -814,6 +826,129 @@ impl std::fmt::Display for GenerateMode {
...
+    }
+}
+
+impl DependencyKind {
+    /// Returns the default phases used for dependencies with nothing specified
+    pub fn default_phases() -> Vec<Self> {
+        vec![Self::Build]
+    }
+}
+
```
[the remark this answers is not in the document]
mistydemeo: Yeah, good point. Maybe we want to make the default an empty array, which has special-cased behaviour that specifying `Build` doesn't - like with the targets.

cargo-dist/templates/installer/homebrew.rb.j2
mistydemeo: Yeah, that's right. I'm placing a newline between the "package definition" stuff that ends with the version and the start of the dependencies. That's Homebrew's own standard.

cargo-dist/src/config.rs
```diff
@@ -814,6 +826,129 @@ impl std::fmt::Display for GenerateMode {
...
+        /// Version specification, as expected by the underlying package manager
+        version: Option<String>,
+        /// Phases at which the dependency is required
+        #[serde(default = "DependencyKind::default_phases")]
+        phase: Vec<DependencyKind>,
+        /// One or more targets this package should be installed on; defaults to all targets if not specified
+        #[serde(default = "Vec::new")]
+        targets: Vec<String>,
+    },
+}
```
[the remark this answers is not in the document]
mistydemeo: I don't remember why I wrote it this way, but with some tweaking it seems to work.

cargo-dist/src/tasks.rs
mistydemeo: Hmmm, right. Good point - I hadn't really handled the interaction between the two, but I should.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: I think so, yeah. I don't mind keeping this like this for now though - a new subcommand that handles this would be a nice stretch goal, especially since we *may* have an external crate to use for that. It's IMO not critical until we have a second CI backend.

cargo-dist/src/tasks.rs
mistydemeo: Updated to do this, though it looks like we don't have integration tests that make use of package-local stuff so I didn't add an integration test.

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: ok i'm fine with landing it like this, if we don't like it we can clean it up in followups

cargo-dist/src/tasks.rs
```diff
@@ -736,6 +740,18 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            .system_dependencies
+            .clone()
+            .unwrap_or_default();
+
+        // Include all package-local dependencies from everything we're building
+        for package in &package_metadata {
+            if let Some(package_deps) = &package.system_dependencies {
+                system_dependencies.append(&mut package_deps.clone());
+            }
+        }
```
Gankra, line 753: We're too early in the code to make this determination -- we only have that info around here, when we know `announcing.rust_releases`, because we """need""" to be factoring in the whether they're releasing `my-app-v1.0.0` (just one app in the workspace) or `v1.0.0` (~everything).

https://github.com/axodotdev/cargo-dist/blob/815ca1096e382c3c31f074d921f4d67844c09906/cargo-dist/src/tasks.rs#L2235

cargo-dist/src/tasks.rs
```diff
@@ -1373,6 +1390,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+        let dependencies: Vec<String> = self
+            .inner
+            .system_dependencies
+            .homebrew
+            .clone()
+            .into_iter()
+            .filter(|(_, package)| package.0.stage_wanted(&DependencyKind::Run))
+            .map(|(name, _)| name)
+            .collect();
+
```
Gankra, line 1402: We could be making multiple homebrew packages (multiple packages in the workspace, each getting its own rb), and ideally each one precisely only depends on what it needs, yeah?

Gankra (changes requested):

Gankra (approved):

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -115,11 +115,21 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          ${{ matrix.packages_install }}
       - name: Build artifacts
+        if: ${{ hashFiles('Brewfile') == '' }}
         run: |
           # Actually do builds and make zips and whatnot
           cargo dist build ${{ needs.plan.outputs.tag-flag }} --output-format=json ${{ matrix.dist_args }} > dist-manifest.json
           echo "cargo dist ran successfully"
+      - name: Build artifacts (using Brewfile)
```
mistydemeo, line 127: The duplication here is because Windows Powershell doesn't support the syntax that would have let me do the `$CARGO dist blah` exactly the same way as it works on bash. It worked out to be easier to just duplicate this with slightly different logic for now.

PR #437, opened 2023-09-18
feat(init): various fixes/improvements

Gankra:
* new target selection UI
* if you have Github CI enabled, we won't ask anymore
* if you have pr-run-mode set, we won't ask anymore (we also no longer omit it from output when the default is selected, so that we can distinguish new init from made-a-decision)
* extra unimportant words removed
* if you have no CI backend enabled we will still allow you to turn on msi
* we will always run generate at the end (unless --no-generate-ci is set) because generate now detects which parts need to run on its own (and the current logic was preventing CI-less MSI from working).
* `[profile]` de-emphasized, sent to the end, not mentioned if unchanged

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           submodules: recursive
       - name: Install cargo-dist
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.3.0-prerelease.4/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.3.0-prerelease.5/cargo-dist-installer.sh | sh"
       - id: plan
         run: |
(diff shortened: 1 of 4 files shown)


Gankra:
Here are some screenshots of the 3 main init workflows. Each screenshot is a spot where your are stopped to make a decision. The displayed value is the the one `--yes` would have selected. I always press enter, getting the same result as --yes.


# First Init Workflow

This is a True Init, where cargo-dist wasn't setup at all.

image
image
image
image
----
final output:
image



# Upgrade Workflow

This is someone running `cargo dist init` to update to a new version of cargo-dist. In this case they are prompted to setup the new pr-run-mode feature:

image
image
image
image
----
final output:
image


# Up To Date Workflow

This is someone running `cargo dist init` with an up to date cargo-dist. In this case they're either looking to change targets/installers, or just want to run `generate`.

image
image
----
final output:
image

Gankra:
Got tired of seeing the `[profile]` stuff so prominently, so I removed it completely from the top, and now have it only show up if it's being made for the first time:

# First Init

image

# Update / Up To Date

image

mistydemeo (approved):
Tested, confirmed working

PR #447, opened 2023-09-25
Better help

Gankra:


cargo-dist/src/errors.rs
```diff
@@ -149,7 +149,11 @@ pub enum DistError {
 
     /// parse_tag concluded there was nothing to release
     #[error("This workspace doesn't have anything for cargo-dist to Release!")]
-    NothingToRelease,
+    NothingToRelease {
+        /// full help printout (very dynamic)
```
ashleygwilliams, line 153: may be useful to hint at how it is dynamic

cargo-dist/src/lib.rs
```diff
@@ -511,7 +511,11 @@ fn build_cargo_target(dist_graph: &DistGraph, target: &CargoBuildStep) -> Result
         for (exe_name, to_copy) in &exes {
             for (src_path, dest_path) in to_copy {
                 if src_path.as_str().is_empty() {
-                    return Err(miette!("failed to find bin {} ({})", exe_name, package_id));
+                    return Err(miette!(
+                        "failed to find bin {} ({}) -- did the cargo build above have errors?",
```
ashleygwilliams, line 515: oh this is good

cargo-dist/src/tasks.rs
```diff
@@ -2280,16 +2280,22 @@ pub fn gather_work(cfg: &Config) -> Result<DistGraph> {
...
-/// See if we should dist this package
+/// See if we should dist this package.
 ///
-/// Some(disabled_reason) is returned if it shouldn't be
+/// Some(disabled_reason) is returned if it shouldn't be.
+///
+/// This code is written to assume a package and its binaries should be distable,
+/// and then runs through a battery of disqualifying reasons.
+///
+/// A notable consequence of this is that if --tag wasn't passed, then we will default to
```
ashleygwilliams, line 2290: (caveat: i did not yet cross check our publish facing docs) it might be useful to also doc this behavior for end users?

cargo-dist/src/tasks.rs
```diff
@@ -2280,16 +2280,22 @@ pub fn gather_work(cfg: &Config) -> Result<DistGraph> {
...
+/// publish=false. Later steps will then check if a coherent announcement tag exists that
+/// covers everything this function spat out.
 fn check_dist_package(
     graph: &DistGraphBuilder,
     pkg_id: PackageIdx,
     pkg: &axoproject::PackageInfo,
-    announcement_tag: Option<&str>,
-    announcing_package: Option<PackageIdx>,
-    announcing_version: Option<&Version>,
+    announcing: &PartialAnnouncementTag,
```
ashleygwilliams, line 2298: it's not clear to me what "partial" means here

cargo-dist/src/tasks.rs
```diff
@@ -2313,21 +2319,21 @@ fn check_dist_package(
     }
 
     // If we're announcing a package, reject every other package
-    if let Some(id) = announcing_package {
+    if let Some(id) = announcing.package {
```
ashleygwilliams, line 2322: oh this is nice

cargo-dist/src/tasks.rs
```diff
@@ -2428,17 +2434,112 @@ pub(crate) struct AnnouncementTag {
     pub rust_releases: Vec<(PackageIdx, Vec<String>)>,
 }
 
+/// details on what we're announcing (partially computed)
```
ashleygwilliams, line 2437: ah this might answer my question from above. to check my understanding- partially computer means some info is potentially provided by the user where others are coming from compute we're doing?

cargo-dist/src/tasks.rs
```diff
@@ -2428,17 +2434,112 @@ pub(crate) struct AnnouncementTag {
...
+///
+/// `tag` being None here is equivalent to `--tag` not being passed, and tells us to infer
+/// the tag based on things like "every package has the same version, assume we're
+/// announcing that version".
+///
+/// `needs_coherent_announcement_tag = false` tells us to produce a result even if inference
+/// fails to find a tag that will unambiguously work. This is used by commands like `init`
+/// and `generate` which want to consider "everything" even if the user never actually
+/// could announce everything at once. In this case dummy values will appear in every field
+/// except for `AnnouncementTag::rust_releases` which will contain every distable binary
```
ashleygwilliams, line 2459: should we rename `rust_releases` `distable_binaries`?

cargo-dist/src/tasks.rs
```diff
@@ -2428,17 +2434,112 @@ pub(crate) struct AnnouncementTag {
...
+/// in the workspace.
+pub(crate) fn select_tag(
     graph: &DistGraphBuilder,
```
ashleygwilliams, line 2462: this is starting to feel like a struct and methods

cargo-dist/src/tasks.rs
```diff
@@ -2428,17 +2434,112 @@ pub(crate) struct AnnouncementTag {
...
     tag: Option<&str>,
     needs_coherent_announcement_tag: bool,
 ) -> DistResult<AnnouncementTag> {
+    // Parse the tag
+    let mut announcing = parse_tag(graph, tag)?;
+    // Select which packages/binaries are available from that tag
+    let rust_releases = select_packages(graph, &announcing);
+
+    // Don't proceed if the conclusions don't make sense
+    if rust_releases.is_empty() {
```
ashleygwilliams, line 2472: i might break this and the `if tag is none` block into functions, esp if you turn this into struct/methods (easier to unit test also)

cargo-dist/src/tasks.rs
```diff
@@ -2558,79 +2669,80 @@ pub(crate) fn parse_tag(
...
-        } else {
-            // We don't need a coherent announcement tag so use a fake one to continue on
-            announcement_tag = Some("v1.0.0-FAKEVER".to_owned());
-            announcing_prerelease = true;
-            announcing_version = Some("1.0.0-FAKEVER".parse().unwrap());
+    let Some(some_pkg) = versions
+        .first_key_value()
+        .and_then(|(_, packages)| packages.first())
+    else {
+        return r#"It appears that you have no packages in your workspace with distable binaries. You can rerun with "--verbose=info" to see what cargo-dist thinks is in your workspace. Here are some typical issues:
```
ashleygwilliams, line 2707: nice

cargo-dist/src/tasks.rs
```diff
@@ -2558,79 +2669,80 @@ pub(crate) fn parse_tag(
...
+        "--tag={}-v{}",
+        info.name,
+        info.version.as_ref().unwrap().cargo()
+    );
+
+    writeln!(
+        help,
+        "you can also request any single package with {some_tag}"
+    )
+    .unwrap();
```
ashleygwilliams, line 2743: is there a reason to use writeln + unwrap over println? (same above)

cargo-dist/src/tasks.rs
Gankra: yeah so the inference system exists to enable a few usecases:

* greasing the wheels on someone just messing around locally and trying cargo-dist out
* pr checks don't have a tag to use so need to use a fake one and "whatever the current version is" works great for that

definitely worth calling out in the pr-check docs if it isn't already

cargo-dist/src/tasks.rs
Gankra: Yeah basically. Notably there's more Option/missing data that is guaranteed to exist by the end.

cargo-dist/src/tasks.rs
Gankra: Yeah I think this code is so old it was gesturing at the possibility of there being like, npm packages in the workspace that we understand and can announce releases of. I'm gonna avoid doing this rn for churn but will keep an eye on it for when we factor this stuff out to a module.

cargo-dist/src/tasks.rs
Gankra: Yep, planned.

cargo-dist/src/tasks.rs
Gankra: We need to shove the result into an Error type for miette.

mistydemeo (approved):

PR #471, opened 2023-10-03
feat: push `brew bundle exec` deeper

mistydemeo:
This wraps the `cargo` command with `brew bundle` at the immediate place we call it, instead of wrapping the entire `cargo dist` invocation. This should simplify calling it and ensures we retain more environment variables.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -12,6 +12,7 @@
 
 use std::{
     collections::{BTreeMap, HashMap},
+    env,
     process::Command,
 };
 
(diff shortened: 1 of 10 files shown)


cargo-dist/src/lib.rs
```diff
@@ -372,7 +373,17 @@ fn build_cargo_target(dist_graph: &DistGraph, target: &CargoBuildStep) -> Result
         target.target_triple, target.profile
     );
 
-    let mut command = Command::new(&dist_graph.tools.cargo.cmd);
+    let cargo = &dist_graph.tools.cargo.cmd;
+    let skip_brewfile = env::var("DO_NOT_USE_BREWFILE").is_ok();
+
+    let mut command;
+    if Utf8Path::new("Brewfile").exists() && !skip_brewfile {
+        command = Command::new("brew");
```
Gankra, line 381: Can we register `brew` as an optional `dist_graph.tools` item (like `rustup`), and also make this path conditional on its presence? Could imagine some deeply cursed situations where someone checks in a Brewfile and it breaks windows builds.

Gankra (approved):

PR #474, opened 2023-10-10
feat(packages): get name of Homebrew tap if exists

mistydemeo:
This ensures that we preserve the full name of a package, tap included, if it was installed from a non-default tap. We do this by reading Homebrew's install receipt for packages we detected as coming from Homebrew. If the receipt exists, we read it out and check what tap the package is from; if it's not `homebrew/core`, we then prepend it to the detected package name.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -966,7 +966,29 @@ impl Library {
         if let Some(prefix) = brew_prefix {
             let cloned = library.clone();
             let stripped = cloned.strip_prefix(prefix).unwrap();
-            let package = stripped.split('/').nth(0).unwrap();
+            let mut package = stripped.split('/').nth(0).unwrap().to_owned();
+
+            // The path alone isn't enough to determine the tap the formula
+            // came from. If the install receipt exists, we can use it to
+            // get the name of the source tap.
+            let receipt = Utf8PathBuf::from(&prefix)
+                .join(&package)
+                .join("INSTALL_RECEIPT.json");
+
+            // If the receipt doesn't exist or can't be loaded, that's not an
+            // error; we can fall back to the package basename we parsed out
+            // of the path.
+            if receipt.exists() {
+                let _ = SourceFile::load_local(&receipt)
+                    .and_then(|file| file.deserialize_json())
+                    .map(|parsed: serde_json::Value| {
+                        if let Some(tap) = parsed["source"]["tap"].as_str() {
+                            if tap != "homebrew/core" {
+                                package = format!("{tap}/{package}");
+                            }
+                        }
+                    });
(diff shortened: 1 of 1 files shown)


cargo-dist/src/lib.rs
```diff
@@ -966,7 +966,26 @@ impl Library {
...
+
+            // The path alone isn't enough to determine the tap the formula
+            // came from. If the install receipt exists, we can use it to
+            // get the name of the source tap.
+            let receipt = Utf8PathBuf::from(&prefix)
+                .join(&package)
+                .join("INSTALL_RECEIPT.json");
+
+            if receipt.exists() {
+                let _ = SourceFile::load_local(&receipt).and_then(|file| {
```
Gankra, line 979: just checking we are in fact intentionally ignoring all errors and silently moving forward?

cargo-dist/src/lib.rs
```diff
@@ -966,7 +966,26 @@ impl Library {
...
+                    file.deserialize_json().map(|parsed: serde_json::Value| {
```
Gankra, line 980: idiomatically you should end the and_then after deserialize_json() and have the `map` chain after it (removes a level of indent)

cargo-dist/src/lib.rs
```diff
@@ -966,7 +966,26 @@ impl Library {
...
+                        if let Some(tap) = parsed["source"]["tap"].as_str() {
+                            if tap != "homebrew/core" {
```
Gankra, line 982: just checking, do we need to care about case here?

cargo-dist/src/lib.rs
```diff
@@ -966,7 +966,26 @@ impl Library {
...
+            // came from. If the install receipt exists, we can use it to
+            // get the name of the source tap.
+            let receipt = Utf8PathBuf::from(&prefix)
+                .join(&package)
+                .join("INSTALL_RECEIPT.json");
+
+            if receipt.exists() {
+                let _ = SourceFile::load_local(&receipt).and_then(|file| {
+                    file.deserialize_json().map(|parsed: serde_json::Value| {
+                        if let Some(tap) = parsed["source"]["tap"].as_str() {
```
Gankra, line 981: 👍 confirmed serde_json lets you do this and won't panic

Gankra (approved):
looks good, just some "does this work the way i think it does" questions

cargo-dist/src/lib.rs
mistydemeo: Oh cool, done.

cargo-dist/src/lib.rs
mistydemeo: Added a comment to that effect.

PR #475, opened 2023-10-11
feat(homebrew): use linkage to infer deps

mistydemeo:
As planned in #426 and #428, this leverages the linkage reporter to detect which packages to add as dependencies to the Homebrew formula. Even if the user doesn't specify them as runtime dependencies, if we saw that a Homebrew package was dynamically linked against, we'll add it to the list of dependencies.

This requires us to store linkage data between runs so we can fetch it later. We do that by storing the dist metadata after each build in the temporary artifact bucket, and then download it at the time we build the Homebrew formula in order to gain access to linkage data for every build.

For a sample of how this was used, I tested with a package which specified two Mac dependencies:

```toml
[workspace.metadata.dist.dependencies.homebrew]
cmake = { targets = ["x86_64-apple-darwin"] }
libcue = { version = "2.2.1", targets = ["x86_64-apple-darwin"] }
```

Neither is marked as a runtime dependency. The linkage report was able to correctly determine that libcue would be needed at runtime, and add it as a dependency, while cmake would not. https://github.com/mistydemeo/homebrew-cargodisttest/commit/c5127d36c12e985d53a2d7c5501cd1a5e15338ac

### DIFF
--- a/cargo-dist/src/backend/installer/homebrew.rs
+++ b/cargo-dist/src/backend/installer/homebrew.rs
@@ -1,7 +1,8 @@
 //! Code for generating installer.sh
 
-use axoasset::LocalAsset;
+use axoasset::{LocalAsset, SourceFile};
 use camino::Utf8PathBuf;
+use cargo_dist_schema::DistManifest;
 use serde::Serialize;
 
 use super::InstallerInfo;
@@ -48,6 +49,41 @@ pub(crate) fn write_homebrew_formula(
     source_info: &HomebrewInstallerInfo,
 ) -> DistResult<()> {
     let mut info = source_info.clone();
+
+    // Collect all dist-manifests and fetch the appropriate Mac ones
+    let mut manifests = vec![];
+    for file in graph.dist_dir.read_dir()? {
+        let path = file?.path();
+        if let Some(filename) = path.file_name() {
+            if !filename.to_string_lossy().ends_with("-dist-manifest.json") {
+                continue;
+            }
+        }
+
+        let json_path = Utf8PathBuf::try_from(path)?;
+        let data = SourceFile::load_local(json_path)?;
+        let manifest: DistManifest = data.deserialize_json()?;
+
+        if manifest.linkage.iter().any(|l| {
+            info.arm64.is_some() && l.target == "aarch64-apple-darwin"
+                || info.x86_64.is_some() && l.target == "x86_64-apple-darwin"
+        }) {
+            manifests.push(manifest);
+        }
+    }
+
+    // Fetch any detected dependencies from the linkage data
+    let dependencies = manifests.into_iter().flat_map(|m| {
+        m.linkage.into_iter().map(|l| {
+            l.homebrew
+                .into_iter()
+                .filter_map(|lib| lib.source)
+                .collect()
+        })
+    });
+
+    // Merge with the manually-specified deps
+    info.dependencies.extend(dependencies);
 
     // Generate sha256 as late as possible; the artifacts might not exist
     // earlier to do that.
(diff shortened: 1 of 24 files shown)


cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -49,6 +51,40 @@ pub(crate) fn write_homebrew_formula(
 ) -> DistResult<()> {
     let mut info = source_info.clone();
 
+    // Collect all dist-manifests and fetch the appropriate Mac ones
+    let mut manifests = vec![];
+    for file in glob(&format!("{}/*-dist-manifest.json", &graph.dist_dir))? {
```
mistydemeo, line 56: I used a glob for convenience, but given the simplicity of the naming format, we can probably forego the extra dependency and manually iterate every file in that directory to check if it fits our naming format.

mistydemeo: Right now the code to parse and instantiate the dist manifests lives here, but it doesn't have to stay that way. If we anticipate using it elsewhere, it may be better to have it in a helper function somewhere.

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -49,6 +51,40 @@ pub(crate) fn write_homebrew_formula(
...
+                        .into_iter()
+                        .filter_map(|lib| lib.source)
+                        .collect()
+                })
+                .collect::<Vec<String>>()
+        })
+        .collect();
+
+    // Merge with the manually-specified deps
+    info.dependencies.append(&mut dependencies);
```
Gankra, line 86: i believe all 3 of these collects can be removed if you just replace `append` with `extend`.

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -141,11 +142,20 @@ jobs:
...
           name: artifacts
-          path: ${{ steps.cargo-dist.outputs.paths }}
+          path: |
+            ${{ steps.cargo-dist.outputs.paths }}
+      - name: "Upload manifest"
+        uses: actions/upload-artifact@v3
+        with:
+          name: artifacts
+          path: |
+            ${{ env.BUILD_MANIFEST_NAME }}
```
Gankra, line 158: if it works, copying the manifest to target/distrib/ and just doing one upload seems simpler/cleaner

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -181,6 +191,9 @@ jobs:
           echo "paths<<EOF" >> "$GITHUB_OUTPUT"
           jq --raw-output ".artifacts[]?.path | select( . != null )" dist-manifest.json >> "$GITHUB_OUTPUT"
           echo "EOF" >> "$GITHUB_OUTPUT"
+
+          # Remove the granular manifests
+          rm target/distrib/*-dist-manifest.json
```
Gankra, line 196: seems we need to do this in publish-release

Gankra (approved):

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Looks like we still need `l.homebrew`'s `collect` for it to figure out that the type has changed to `String` instead of `Library`, but otherwise, yeah - simplifies it. I'll push that up.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: I don't want to deal with canonicalizing the path so both branches of the `path` are absolute. The environment variable's assigned outside bash

Gankra: i don't think you need to canonicalize the path?

mistydemeo: I should have just trusted it'd work, looks like it is happy with that!

PR #479, opened 2023-10-12
feat(installers): pass data from local builds to global

mistydemeo:
This is a followup to #475. In that PR, I enabled reusing previous dist manifests from previous builds in a CI environment. It doesn't, however, cover the case where the previous builds happened within the *same run*. This fixes that. I've split local/global builds into separate categories from each other, instead of just by ordering them within a single array. I then added a temporary computation of dist manifests after local builds are run, but before global builds are run; that temporary manifest is then passed to any global job that cares to ask for it. Currently, it's only consumed by the Homebrew installer, but it might be wanted by something else later.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -72,8 +72,13 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
     }
     eprintln!();
 
-    // Run all the build steps
-    for step in &dist.build_steps {
+    // Run all the local build steps first
+    for step in &dist.local_build_steps {
+        run_build_step(&dist, step)?;
+    }
+
+    // Next the global steps
+    for step in &dist.global_build_steps {
         run_build_step(&dist, step)?;
     }
 
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -74,12 +74,16 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
 
     // Run all the local build steps first
     for step in &dist.local_build_steps {
-        run_build_step(&dist, step)?;
+        run_build_step(&dist, step, &[])?;
     }
+
+    // Calculate a temporary build manifest from the output of the local builds
+    // (includes their linkage data)
+    let manifests = vec![build_manifest(cfg, &dist)?];
 
     // Next the global steps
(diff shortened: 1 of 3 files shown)


cargo-dist/src/lib.rs
```diff
@@ -72,9 +72,18 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
...
-    // Run all the build steps
-    for step in &dist.build_steps {
-        run_build_step(&dist, step)?;
+    // Run all the local build steps first
+    for step in &dist.local_build_steps {
+        run_build_step(&dist, step, &[])?;
+    }
+
+    // Calculate a temporary build manifest from the output of the local builds
+    // (includes their linakge data)
```
Gankra, line 81:
```suggestion
    // (includes their linkage data)
```

cargo-dist/src/tasks.rs
```diff
@@ -188,7 +188,9 @@ pub struct DistGraph {
     pub artifact_download_url: Option<String>,
 
     /// Targets we need to build
-    pub build_steps: Vec<BuildStep>,
+    pub local_build_steps: Vec<BuildStep>,
+    /// Targets we need to build
+    pub global_build_steps: Vec<BuildStep>,
```
Gankra, line 193: Probably worth distinguishing the doc-comments

cargo-dist/src/lib.rs
```diff
@@ -72,9 +72,18 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
...
+    let manifests = vec![build_manifest(cfg, &dist)?];
```
Gankra, line 82: Wow if this just works that's incredibly fucking elegant, kudos.

(It is *slightly* a lie because it claims the global things are already built, I think, but that's fine if global things don't depend on other global things. And if they did we'd need to do even more complex work beyond the scope of this PR.)

Gankra (approved):
Damn that was way simpler than i thought it would need to be, great work!

cargo-dist/src/lib.rs
mistydemeo: AFAICT it's actually working! It's the same dist manifest we get via the "copy JSON back and forth between files" method, functionally, and linkage data is intact. Right now, global builds only depend on local builds, so I think this is safe.

PR #481, opened 2023-10-12
chore: improve the quality of diffs

Gankra:


Gankra:
image

Cargo.lock
```diff
@@ -319,6 +319,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "similar",
```
mistydemeo, line 322: TIL this was already in our dependency tree. Only via `insta` though so I guess it wasn't in release builds.

mistydemeo (approved):
Nice! Really like this improved output.

Cargo.lock
Gankra: yeah i was just like "ok what the fuck does insta use" and the answer was "a library they made for insta and gloriously spun off as a really nice standalone project"

PR #483, opened 2023-10-17
feat(build): enable Linux musl builds

### LINKED ISSUE #75: musl support (prefer musl for linux?)
Prebuilt binaries for linux are a mess for various reasons (e.g. #73), and preferring musl over glibc fixes some of those issues (in exchange for other subtle problems that I'm assured are minor in comparison).

This requires some basic cross-compilation support to get the toolchain from rustup and to use musl-tools (#74).

mistydemeo:
This adds support for musl builds on Linux. This currently supports x86_64 but not aarch64, as with glibc. Sample successful build: https://github.com/mistydemeo/cargodisttest/actions/runs/6540051759/job/17759406467 A few notes:

* We're building on an Ubuntu host, so it's not being built on a musl-native distro.
* We use the official, default Rust tooling for this from `rustup`, which means musl is statically linked instead of dynamically linked like some musl-based distros prefer.
* Because of this, *dynamically linking to external dependencies will fail* in musl builds. I briefly toyed with disallowing system dependency installation for that reason, but there's no reason to block people from installing cmake/whatever build tools just because they won't be able to link against some system libraries. It does, however, mean there are gotchas with musl support that we have to document, because they don't exist for glibc builds.

Fixes #75.

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -306,8 +306,13 @@ fn package_install_for_targets(
 
                 return Some(brew_bundle_command(&packages));
             }
-            "i686-unknown-linux-gnu" | "x86_64-unknown-linux-gnu" | "aarch64-unknown-linux-gnu" => {
-                let packages: Vec<String> = packages
+            "i686-unknown-linux-gnu"
+            | "x86_64-unknown-linux-gnu"
+            | "aarch64-unknown-linux-gnu"
+            | "i686-unknown-linux-musl"
+            | "x86_64-unknown-linux-musl"
+            | "aarch64-unknown-linux-musl" => {
+                let mut packages: Vec<String> = packages
                     .apt
                     .clone()
                     .into_iter()
@@ -321,6 +326,12 @@ fn package_install_for_targets(
                         }
                     })
                     .collect();
+
+                // musl builds may require musl-tools to build;
+                // necessary for more complex software
+                if target.ends_with("linux-musl") {
+                    packages.push("musl-tools".to_owned());
+                }
 
                 if packages.is_empty() {
                     return None;
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -1143,7 +1143,12 @@ fn determine_linkage(path: &Utf8PathBuf, target: &str) -> DistResult<Linkage> {
     let libraries = match target {
         // Can be run on any OS
         "i686-apple-darwin" | "x86_64-apple-darwin" | "aarch64-apple-darwin" => do_otool(path)?,
-        "i686-unknown-linux-gnu" | "x86_64-unknown-linux-gnu" | "aarch64-unknown-linux-gnu" => {
+        "i686-unknown-linux-gnu"
+        | "x86_64-unknown-linux-gnu"
+        | "aarch64-unknown-linux-gnu"
+        | "i686-unknown-linux-musl"
+        | "x86_64-unknown-linux-musl"
+        | "aarch64-unknown-linux-musl" => {
             // Currently can only be run on Linux
             if std::env::consts::OS != "linux" {
                 return Err(DistError::LinkageCheckInvalidOS {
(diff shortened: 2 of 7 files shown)


cargo-dist/src/backend/ci/github.rs
```diff
@@ -253,11 +258,18 @@ fn install_dist_for_github_runner<'a>(
...
-    if runner == GITHUB_LINUX_RUNNER || runner == GITHUB_MACOS_RUNNER {
-        install_sh
+    targets: &[&String],
+) -> String {
+    if runner == GITHUB_LINUX_RUNNER && targets.contains(&&"x86_64-unknown-linux-musl".to_owned()) {
+        // Necessary to ensure we have the proper toolchain
+        format!(
+            r#"rustup target add x86_64-unknown-linux-musl
+{install_sh}"#
+        )
```
Gankra, line 268: cargo-dist already has logic for calling rustup internally, you should be generalizing that instead https://github.com/axodotdev/cargo-dist/blob/7544b23b5fa938f0bb6e2b46b1dece9a79ed157b/cargo-dist/src/tasks.rs#L1875-L1889

mistydemeo:
> do we not need to setup musl-tools? i thought github's images didn't include those...

I'm as surprised as you are! But you can see the build succeeded without installing it. I was able to confirm locally; so long as build-essential is installed, I can `cargo build --release --target=aarch64-unknown-linux-musl` without `musl-tools`.

cargo-dist/src/lib.rs
```diff
@@ -1250,6 +1255,7 @@ pub fn default_desktop_targets() -> Vec<String> {
     vec![
         // Everyone can build x64!
         axoproject::platforms::TARGET_X64_LINUX_GNU.to_owned(),
+        axoproject::platforms::TARGET_X64_LINUX_MUSL.to_owned(),
```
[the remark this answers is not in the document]
mistydemeo: Yeah, right now they're the same list. I'll adjust it so there's two lists.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -253,11 +258,18 @@ fn install_dist_for_github_runner<'a>(
     runner: GithubRunner,
     install_sh: &'a str,
     install_ps1: &'a str,
-) -> &'a str {
-    if runner == GITHUB_LINUX_RUNNER || runner == GITHUB_MACOS_RUNNER {
-        install_sh
+    targets: &[&String],
+) -> String {
+    if runner == GITHUB_LINUX_RUNNER && targets.contains(&&"x86_64-unknown-linux-musl".to_owned()) {
```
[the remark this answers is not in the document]
mistydemeo: It does, but everywhere else in the file we're just inlining the triples. I figure we should be consistent.

Gankra:
It might be worth checking a bigger project that might be pulling in transitive C deps -- oranda for instance.

mistydemeo:
Boo, turns out oranda does need it. Adjusted the build to make sure we install it.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -253,11 +258,18 @@ fn install_dist_for_github_runner<'a>(
...
+        // Necessary to allow certain more complex builds
+        format!(
+            r#"sudo apt-get install musl-tools
+{install_sh}"#
```
Gankra, line 267: wasn't the whole idea that we could smuggle this into native sysdeps?

cargo-dist/src/tasks.rs
```diff
@@ -1888,6 +1888,19 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
                 }
             }
 
+            if target.ends_with("linux-musl")
+                && self.inner.tools.cargo.host_target.ends_with("linux-gnu")
+            {
```
Gankra, line 1893: hmm, any particular motivation to dropping the "target != host_target" check? (Not sure if it would ever trip, but would it ever be wrong..?)

I was thinking we might want to simplify this + macos down to just "always run rustup if we're not building for host", but that might be too far..?

Gankra (approved):

cargo-dist/src/tasks.rs
mistydemeo: Just figure that since this check implicitly contains that check, there's no reason to add the extra conditional.

bjorn3:
Could you explicitly enable `-Ctarget-feature=+crt-static`? The intent is to switch the musl target to default to dynamic linking in the future. Having everyone who depends on musl statically linking explicitly specify this will make it easier to flip the switch in the future.

ashleygwilliams:
@bjorn3 gonna pull this comment out into an issue so we don't miss it!

PR #495, opened 2023-10-19
fix(linkage): improve apt package resolution

mistydemeo:
The path linked against might be a symlink that `dpkg -S` can't resolve to the providing package. If we do a `realpath` call on it, we get a path that we can more reliably resolve later.


ashleygwilliams:
i'm not qualified to review this- @Gankra can you take a look today?

Gankra (approved):

PR #497, opened 2023-10-19
Installers: add musl detection

mistydemeo:
Now that we have musl binaries, we should teach the installers to use them. In particular, we want to ensure musl gets used in the following circumstances:

* The user is running a non-glibc-based distro, in which case the glibc binaries are guaranteed not to run
* The user is running a glibc-based distro, but their glibc is too old to run the glibc binaries we built

This makes the following changes:

* `install.sh` uses musl
* npm uses musl
* npm inherits the macOS Rosetta code that it was missing


Gankra (approved):
We paired on this, already reviewed 😎

ashleygwilliams:
do we know what's causing the failure on this PR?

mistydemeo:
It's shellcheck; I'll apply the errors it flagged. I should see if it's possible for us to simulate this locally so we don't need to rely on just CI for this, too.

Issue #498, opened 2023-10-19
sysdeps on aarch64-apple-darwin


ashleygwilliams:
currently, cargo-dist targets aarch64-apple-darwin as a cross target, which means that projects that require native sysdeps are particularly challenging. this is a tracking issue for folks who have this issue, and thoughts about how to approach it.

mistydemeo:
We're slightly constrained here because we don't have access to native M1 runners. GitHub [recently launched](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/) M1 runners, but only as an "xlarge" runner. This is one of their categories of runners that's exclusively available as a paid option; thus far, we've tried to ensure that all of our CI options can run on the free runners.

I do realize, though, that we're a bit more constrained than we need to be because I have us unconditionally `brew bundle exec`ing on arm64. That's failing not just because libraries are compiled for Intel, but because it's injecting build environment variables that are inappropriate for cross-compiling to arm64. I'm going to improve this case before ship. I have two ideas:

* aarch64-apple-darwin builds still `brew bundle install` deps, but no longer do `brew bundle exec`. As a result, it doesn't get any of the build environment modification that Intel does, but commandline tools that are installed onto the `PATH` will work.
* Replace `brew bundle exec` with tooling that picks and chooses which parts of `brew bundle exec`'s environment we want, and then use only those variables/values within our builds. In this case we can actually inherit the same behaviour on Intel. This still won't let aarch64-apple-darwin builds get access to C libraries, but otherwise keeps a bunch of extra niceties we get from `brew bundle exec`.

mistydemeo:
After shipping #499/#504, I've improved the case in GitHub CI. We now no longer always tank builds, and builds using non-library dependencies will work fine. Builds that actually require C libraries will still fail, however, and that won't be fixable until we have a CI option with ARM64 runners.

weiznich:
I believe that this is still broken with the `cargo dist` 0.4 release from today. If you specify dependencies via `[package.metadata.dist.dependencies.homebrew]` key, cargo dist currently seem to install always the x86_64 version of the libraries. That leads to linker errors, complaining about incompatible library versions. [See here for an example CI run where this error occurs) 

[...]

PR #499, opened 2023-10-20
feat(deps): pick and choose parts of brew bundle exec

mistydemeo:
This replaces the invocation to `brew bundle exec` with a lightweight Rusty replacement. Rather than running our entire build within a `brew bundle exec` subprocess, we effectively replace it using data that `brew bundle` gives us. We accept the following environment variables unconditionally:

* `PKG_CONFIG_PATH`
* `PKG_CONFIG_LIBDIR`
* `CMAKE_INCLUDE_PATH`
* `CMAKE_LIBRARY_PATH`

In addition, based on the dependencies contained within `HOMEBREW_DEPENDENCIES` that we receive from `brew bundle`, we calculate the following:

* Paths to add to `PATH`
* Library search paths to add to `RUSTFLAGS`

What we *don't* do, which `brew bundle exec` does, is inherit Homebrew's compiler shims. This means that we don't inherit a handful of things that Homebrew sets for us, but it also means that it doesn't capture references to the system compiler to try to inject extra flags we may decide we don't want.

I tested this with my `cue2ccd`; `cargo dist build --target=aarch64-apple-darwin --artifacts=local` works properly. Cross-compiling to aarch64 from x86_64 and vice versa should work if the Homebrew dependencies being requested are binaries to place in the `PATH`, which didn't work before.

This is a bit messy; I wrote it to get a working version out, but there's stuff to clean up and testing to do before it gets merged

Should address part of #498 - it's solution 2.
Closes #471.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -12,7 +12,7 @@
 
 use std::{
     collections::{BTreeMap, HashMap},
-    fs::File,
+    fs::{self, File},
     io::{Cursor, Read},
     process::Command,
 };
@@ -1122,7 +1122,12 @@ fn do_ldd(path: &Utf8PathBuf) -> DistResult<Vec<String>> {
 
         // Format: libname.so.1 => /path/to/libname.so.1 (address)
         if let Some(path) = line.split(" => ").nth(1) {
-            libraries.push((path.split(' ').next().unwrap()).to_owned());
+            // This may be a symlink rather than the actual underlying library;
+            // we resolve the symlink here so that we return the real paths,
+            // making it easier to map them to their packages later.
+            let lib = (path.split(' ').next().unwrap()).to_owned();
+            let realpath = fs::canonicalize(&lib)?;
+            libraries.push(realpath.to_string_lossy().to_string());
         } else {
             continue;
         }


cargo-dist/src/errors.rs
```diff
@@ -259,6 +259,15 @@ pub enum DistError {
     /// random camino conversion error
     #[error(transparent)]
     FromPathBufError(#[from] camino::FromPathBufError),
+
+    /// Error parsing a string containing an environment variable
+    /// in VAR=value syntax
+    #[error("Unable to parse line as a key/value pair: {line}")]
+    #[diagnostic(help("key=value syntax expected"))]
+    EnvParseError {
```
Gankra, line 267: The error message is a bit vague for an end-user

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+        match line.split_once('=') {
+            Some(parsed) => (key, value) = parsed,
+            None => {
+                return Err(DistError::EnvParseError {
+                    line: line.to_owned(),
+                })
+            }
+        };
+
+        parsed.insert(key, value);
```
Gankra, line 416: a little simpler written as:

```suggestion
        let Some((key, value)) = line.split_once('=') else {
            return Err(DistError::EnvParseError {
                line: line.to_owned(),
            })
        };
        parsed.insert(key, value);
```

Gankra, line 416: Is there established semantics for a key being defined multiple times? (This impl picks the last one)

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
     );
 
+    let mut rustflags = target.rustflags.clone();
+    let mut desired_extra_env = vec![];
+    let skip_brewfile = env::var("DO_NOT_USE_BREWFILE").is_ok();
+    if let Some(brew) = &dist_graph.tools.brew {
+        if Utf8Path::new("Brewfile").exists() && !skip_brewfile {
+            // Uses `brew bundle exec` to just print its own environment,
+            // allowing us to capture what it generated and decide what
+            // to do with it.
```
Gankra, line 508: Does this mean that the code to invoke `brew bundle exec` in `ci.yml.j2` can be removed? (If so, please do)

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+            let result = Command::new(&brew.cmd)
+                .arg("bundle")
+                .arg("exec")
+                .arg("--")
+                .arg("/usr/bin/env")
```
Gankra, line 513: /usr/bin/env here is I assume a binary that just dumps all env-vars in scope when its ran, making this a simple way of telling bundle exec to tattle on itself?

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+                .output()
+                .into_diagnostic()
+                .wrap_err_with(|| "failed to exec brew bundle exec".to_string())?;
+
+            let env_output = String::from_utf8_lossy(&result.stdout).to_string();
```
Gankra, line 518: Don't have a strong opinion, but curious if there is any particular reason why utf8_lossy is better/worse than "error out on non-utf8" in this situation?

cargo-dist/src/lib.rs
```diff
@@ -438,6 +559,11 @@ fn build_cargo_target(dist_graph: &DistGraph, target: &CargoBuildStep) -> Result
             eprintln!(" --package={})", package);
         }
     }
+    // If we generated any extra environment variables to
+    // inject into the environment, apply them now.
+    for (key, value) in desired_extra_env {
+        command.env(key, value);
+    }
```
Gankra, line 566: Checking you're aware of https://doc.rust-lang.org/std/process/struct.Command.html#method.envs ?

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+    if let Some(value) = environment.get("CMAKE_LIBRARY_PATH") {
+        desired_env.push(("CMAKE_LIBRARY_PATH".to_owned(), value.to_string()))
+    }
+    let mut paths = vec![];
+    // For each listed dependency, add it to the PATH
+    if let Some(optprefix) = environment.get("HOMEBREW_OPT") {
+        // Comma-separated list of all dependencies in the recursive tree
+        if let Some(dependencies) = environment.get("HOMEBREW_DEPENDENCIES") {
+            for dep in dependencies.split(',') {
+                let short_name = dep.split('/').next().unwrap();
```
Gankra, line 448: Would be nice to not unwrap here, if possible

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+                paths.push(format!("{pkg_opt}/sbin"));
+            }
+        }
+    }
+    if !paths.is_empty() {
+        let our_path = env!("PATH");
+        let desired_path = format!("{our_path}:{}", paths.join(":"));
+
+        desired_env.insert(0, ("PATH".to_owned(), desired_path));
+    }
```
Gankra, line 463: this part could use with a comment

(A++++ commenting in this PR, btw)

Gankra: ohhh i see, maybe it doesn't need one

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+
+    desired_env
+}
+
+// Similar to the above, we read Homebrew's recursive dependency tree and
+// then append link flags to cargo-dist's rustflags.
+// These ensure that Rust can find C libraries that may exist within
+// each package's prefix.
```
Gankra, line 471:
```suggestion
/// Similar to the above, we read Homebrew's recursive dependency tree and
/// then append link flags to cargo-dist's rustflags.
/// These ensure that Rust can find C libraries that may exist within
/// each package's prefix.
```

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+
+        parsed.insert(key, value);
+    }
+
+    Ok(parsed)
+}
+
+// Takes a BTreeMap of key/value environment variables produced by
+// `brew bundle exec` and decides which ones we want to keep for our own builds.
+// Returns a Vec containing (KEY, value) tuples.
```
Gankra, line 424:
```suggestion
/// Takes a BTreeMap of key/value environment variables produced by
/// `brew bundle exec` and decides which ones we want to keep for our own builds.
/// Returns a Vec containing (KEY, value) tuples.
```

cargo-dist/src/lib.rs
```diff
@@ -395,13 +396,133 @@ fn write_checksum(checksum: &str, src_path: &Utf8Path, dest_path: &Utf8Path) ->
...
+// These ensure that Rust can find C libraries that may exist within
+// each package's prefix.
+fn determine_brew_rustflags(base_rustflags: &str, environment: &BTreeMap<&str, &str>) -> String {
+    let mut rustflags = base_rustflags.to_owned();
+    // For each listed dependency, add it to CFLAGS/LDFLAGS
+    if let Some(optprefix) = environment.get("HOMEBREW_OPT") {
+        if let Some(dependencies) = environment.get("HOMEBREW_DEPENDENCIES") {
+            for dep in dependencies.split(',') {
+                let short_name = dep.split('/').next().unwrap();
+                let pkg_opt = format!("{optprefix}/{short_name}");
```
Gankra, line 479: it's not a big deal but it does seem slightly odd that we parse this format twice (at very least could be factored to a reusable func)

Gankra (approved):
neato

cargo-dist/src/lib.rs
mistydemeo: Unfortunately, nope.

cargo-dist/src/lib.rs
mistydemeo: Yep! Thanks for the reminder, I actually missed adding that when I wrote this as quickly as I did.

cargo-dist/src/lib.rs
Gankra: Hmm, miscommunication? (wasn't asking if there was a better way, was just asking if i had understood how this was being used)

mistydemeo: Oh sorry, yeah, I misread. `env` is a standard POSIX binary which, when called with no arguments, prints out the process's environment in `sh` syntax.

cargo-dist/src/lib.rs
mistydemeo: Unless I'm missing something, it's impossible for `split` on a string to not return at last a first element. Is there a more idiomatic way to say "get me the first element with no unwrap"? I feel like we don't need to manage an impossible `None` scenario.

Gankra: great point! maybe just a comment reminding the reader that split is so nicely behaved, then

cargo-dist/src/lib.rs
mistydemeo: The string we're parsing from `env` will only ever return a given key once, so I'm trusting the input string to be well-formatted here. That said, I'll add a comment to that effect.

cargo-dist/src/lib.rs
mistydemeo: Yeah, I think I can DRY this up a bit.

PR #503, opened 2023-10-23
feat(eula): update cargo-wix and use new eula APIs for extra files

Gankra:
* Query the possible renders
* Diff them all
* Write them all
* Default to emitting license=false and eula=false to disable all the auto-eula stuff


Gankra:
requires https://github.com/volks73/cargo-wix/pull/247 to land and for a release to be cut

mistydemeo (approved):

PR #528, opened 2023-10-30
fix: make include file-or-dir check late-bound

Gankra:
This makes it possible for the included file to be generated by build.rs.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -313,14 +313,18 @@ fn run_build_step(
     match target {
         BuildStep::Cargo(target) => build_cargo_target(dist_graph, target),
         BuildStep::Rustup(cmd) => rustup_toolchain(dist_graph, cmd),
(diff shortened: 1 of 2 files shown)


cargo-dist/src/lib.rs
```diff
@@ -755,6 +759,14 @@ pub(crate) fn copy_dir(src_path: &Utf8Path, dest_path: &Utf8Path) -> Result<()>
     Ok(())
 }
 
+pub(crate) fn copy_file_or_dir(src_path: &Utf8Path, dest_path: &Utf8Path) -> Result<()> {
```
mistydemeo, line 762: Wonder if we want this in `AxoAsset` at some point.

mistydemeo (approved):

cargo-dist/src/lib.rs
Gankra: Could be -- these shims are all weird artifacts of pre-axoasset

PR #546, opened 2023-11-01
chore: refactor codebase for artifact hosting

Gankra:
* new Hosting type added to dist-manifest
* dist-manifest merging now generalized
* tasks.rs and lib.rs broken up into more files
* serde annotations cleaned up


mistydemeo (approved):
Looks good. Tested, and the Homebrew formula works, complete with linkage: https://github.com/mistydemeo/homebrew-cargodisttest/commit/f87bc85a9e86fb040d3bef6a125f3264dda8fdb7#diff-e409d1e3fabe65a4757ff4bf243f2a3d51402b7dafa3d5b8dbb799865eccb562R7

I do notice that the `dist-manifest.json` we upload isn't the merged one, and is missing the linkage data. Was that intentional, or is that an oversight?

PR #548, opened 2023-11-02
feat(axo-releases): add Axo Releases as a hosting backend

Gankra:
* [x] gazenot client
* [x] compute arguments
* [x] create-artifact-set endpoint
   * [x] call it
   * [x] only call create-artifact-set endpoint once (add mock_networking)
   * [x] don't call create-artifact-set in PR checks (but maybe call a whoami endpoint)
   * [x] absolutely don't call create-artifact-set in PR checks from forks (not even whoami)
   * [x] don't call create-artifact-set when user locally runs `cargo dist plan`? (not even whoami?)
* [x] CI
   * [x] CI: place dist-manifest.json in the dist-dir of every machine so it inherits staging info
   * [x] CI: use the upload endpoint (new cargo-dist command?)
   * [x] CI: use the publish endpoint (new cargo-dist command?)
   * [x] CI: use the announce endpoint? (new cargo-dist command?)


gazenot/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+        //
+        // FIXME: this should be streamed to the request as it's loaded to disk
+        let data = LocalAsset::load(path)?;
+
+        // Send the bytes
+        let response = self
+            .client
+            .post(url.clone())
+            .header("AUTHORIZATION", &self.api_key)
+            .header("AUTHORIZATION", &self.api_key)
```
mistydemeo, line 435: Double-auth?

gazenot/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+            queries.push(tokio::spawn(async move {
+                handle
+                    .create_announcement(url, releases, announcement)
+                    .await
+            }));
+        }
+
+        // Then join on them all
+        for query in queries {
+            query.await??;
```
mistydemeo, line 556: Me too, code. Me too.

gazenot/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+            let server = &self.api_server;
+            let ArtifactSet {
+                owner,
+                package,
+                public_id,
+                ..
+            } = set;
+            format!("https://{server}/api/{owner}/{package}/artifacts/{public_id}/upload")
+        });
+        let url = Url::from_str(&format!("{base}/{filename}"))?;
```
mistydemeo, line 620: Can we use `join` on a parsed `Url` for this? https://docs.rs/reqwest/0.11.22/reqwest/struct.Url.html#method.join In general I wonder if we can get away with less string formatting/interpolation here

mistydemeo (commented):
I haven't had a chance to try yet, but from first glance the `gazenot` crate is making sense to me!

gazenot/src/lib.rs
```diff
@@ -0,0 +1,653 @@
...
+//!     let release_version = "1.0.0".to_owned();
+//!     let release_tag = "v1.0.0".to_owned();
+//!     let is_prerelease = false;
+//!     let announcement_body = "# v1.0.0 (2023-01-01)\n\nWow Cool Changelog".to_owned();
+//!     let files = [
+//!         ("app1", vec!["app1-v1.0.0.zip", "dist-manifest.json"]),
+//!         ("app2", vec!["app2-v1.0.0.zip", "dist-manifest.json"]),
+//!     ]
+//!     .into_iter()
+//!     .collect::<std::collections::HashMap<_, _>>();
```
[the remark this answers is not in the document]
Gankra: oh neat, nice

gazenot/src/lib.rs
Gankra: good catch, fixed

gazenot/src/lib.rs
Gankra: So all the URL path APIs are regrettably DEEPLY CURSED and basically unusable as a result. Unlike the std::path APIs which basically say "oh ok, if you join/push a component, we'll assume the lhs was a dir" the URL crate will only do that if the input ends in a trailing slash. Otherwise it will clobber the last component, on the assumption that it was a file.

This is why oranda has been plagued with bugs that are like "oh shit we forgot a trailing slash and everything broke".

So at that point you're finding yourself needing to be super pedantic about the format of your input anyway, and it's easier/clearer to just use string formatting 🙃

Gankra:
Latest changes since misty review:

* Added proper contextful errors to gazenot by splitting turning the current error type into GazenotErrorInner and and making a new GazenotError type that wraps it an includes a context-message and help.
   * All the non-public "do it once" APIs return GazenotErrorInner, which the public ones wrap in context of "what was i trying to accomplish, and what url was I querying to do it"
* Most of the implementation has been moved to gazenot::client, which is hidden behind a default-on feature_flag. This makes `--no-default-features` gazenot into a really bad gazenot_schema crate (didn't want to work through the full design of that).
   * To streamline this all the APIs have had their `join` logic unified to a join_all function
* cargo_dist_schema "hosting" field reworked
   * now on releases instead of the announcement
   * now an enum with Github | Abyss
   * depends on --no-default-features gazenot so that it can embed gazenot::ArtifactSet, making it trivial to "rehydrate" the client on a different machine
* cargo_dist_schema now has `ensure_release(&mut self, name, version) -> &mut Release`
   * this makes it easier for different parts of the code to do the equivalent of `manifest.entry(release).or_default().actual_modification_i_want_to_make()`, as we now needing to incrementally fill in the fields of a release across different steps of `gather_work`
* cargo_dist impl redone to use the new APIs 
   * ...wait i probably forgot to update the manifest merging code... that's probably really easy now that i have ensure_release though, so... TODO!

Gankra:
pushed up:

* temporarily added "create-release" and "create-announcement" calls to `cargo dist host` for end-to-end testing, can work out exactly how they're invoked later.

gazenot/dist-manifest.json
```diff
@@ -0,0 +1 @@
+{ }
```
mistydemeo, line 1: Was this a mistake?

gazenot/src/error.rs
```diff
@@ -0,0 +1,70 @@
...
+    Join(#[from] tokio::task::JoinError),
+    #[error(transparent)]
+    Url(#[from] url::ParseError),
+    #[error(transparent)]
+    Asset(#[from] axoasset::AxoassetError),
+    #[error("server error {status}")]
+    ResponseError {
+        status: reqwest::StatusCode,
+        #[related]
+        errors: Vec<SimpleError>,
```
mistydemeo, line 59: This is just wrapping strings from the server?

gazenot/src/lib.rs
```diff
@@ -0,0 +1,142 @@
+//! Gaze Not Into The Abyss, Lest You Become A Release Engineer.
+//!
+//! Gazenot is a client library for accessing the Abyss service,
+//! which hosts Releases of various Packages (apps).
+//!
+//! Shipping applications with the Abyss is a 5 step process:
+//!
+//! ```no_run
```
mistydemeo, line 8: Looks like the docstring got snipped?

gazenot/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+mod client;
+#[cfg(feature = "client_lib")]
+pub mod error;
+#[cfg(feature = "client_lib")]
+pub use client::Gazenot;
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// A map that has some stable ordering (for determinism)
```
mistydemeo, line 32: Which map is the comment referring to?

gazenot/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+axoasset = { version = "0.6.0", features = ["json-serde"], optional = true }
+url = { version = "2.4.1", features = ["serde"], optional = true }
+tracing = { version = "0.1.36", features = ["log"], optional = true }
+tokio = { version = "1.12.0", features = ["full"], optional = true }
+camino = { version = "1.1.6", optional = true }
+reqwest = { version = "0.11.22", default-features = false, optional = true, features = [
+    "gzip",
+    "rustls-tls",
+    "json",
+]}
```
mistydemeo, line 30: Nitpick, but newline at the end of the file please

gazenot/src/client.rs
```diff
@@ -0,0 +1,672 @@
...
+struct AnnounceReleaseRequest {
+    releases: Vec<AnnounceReleaseKey>,
+    body: String,
+}
+
+impl Gazenot {
+    /// Gaze Not Into The Abyss, Lest You Become A Release Engineer
+    ///
+    /// This is the vastly superior alias for [`Gazenot::new`].
+    pub fn into_the_abyss(
```
mistydemeo, line 102: 💯

gazenot/src/client.rs
```diff
@@ -0,0 +1,672 @@
...
+    pub async fn upload_files(
+        &self,
+        files: impl IntoIterator<Item = (&ArtifactSet, Vec<Utf8PathBuf>)>,
+    ) -> Result<()> {
+        // Spawn all the queries in parallel...
+        let mut queries = vec![];
+        for (set, sub_files) in files {
+            for file in sub_files {
+                let handle = self.clone();
+                let filename = file.file_name().unwrap();
```
mistydemeo, line 245: Is this safe at this point?

gazenot/src/client.rs
```diff
@@ -0,0 +1,672 @@
...
+
+async fn join_all<T>(
+    queries: impl IntoIterator<Item = (String, Url, tokio::task::JoinHandle<ResultInner<T>>)>,
+) -> Result<Vec<T>> {
+    let mut results = Vec::new();
+    for (desc, url, query) in queries {
+        let result = query
+            .await
+            .map_err(|e| GazenotError::with_url(&desc, url.to_string(), e))?
+            .map_err(|e| GazenotError::with_url(&desc, url.to_string(), e))?;
```
mistydemeo, line 521: Dupe?

gazenot/src/client.rs
```diff
@@ -0,0 +1,672 @@
...
+        results.push(result);
+    }
+    Ok(results)
+}
+
+fn auth_headers(source: &SourceHost, owner: &Owner) -> ResultInner<HeaderMap> {
+    // extra-awkard code so you're on your toes and properly treat this like radioactive waste
+    // DO NOT UNDER ANY CIRCUMSTANCES PRINT THIS VALUE.
+    // DO NOT IMPLEMENT DEBUG ON Abyss OR AbyssInner!!
```
mistydemeo, line 530: Can you add versions of these comments on the `Abyss` and `AbyssInner` definitions? It'll be important to have it flagged up there as well to help avoid it happening by accident.

mistydemeo (approved):
Gazenot implementation looks good, just some minor comments.

gazenot/dist-manifest.json
Gankra: kind-of, it's there for main.rs, which is also to-be-deleted.

gazenot/src/error.rs
Gankra: Yeah, the schema allows the server to return a list of errors. miette/thiserror isn't willing to say "a string is a type of error" so you need to newtype them.

gazenot/src/lib.rs
Gankra: main.rs will go here, yes (really annoying to keep doc example in sync)

gazenot/src/lib.rs
Gankra: Whoops, good catch! (deleted)

gazenot/src/client.rs
Gankra: I should probably be a bit more paranoid about it for good errors, yeah

gazenot/src/client.rs
Gankra: alas no, this is the only way i could see to add context to the old `await??` line

cargo-dist/src/host.rs
```diff
@@ -0,0 +1,240 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+fn check_hosting(_dist: &DistGraph, _manifest: &DistManifest, _abyss: &Gazenot) -> DistResult<()> {
+    // FIXME: implement a ping/whoami API to check the Abyss client is working
```
mistydemeo, line 152: Is this getting done pre- or post-prerelease?

cargo-dist/src/host.rs
```diff
@@ -0,0 +1,240 @@
...
+    tokio::runtime::Handle::current().block_on(abyss.upload_files(files))?;
+    eprintln!("all artifacts hosted!");
+    Ok(())
+}
+
+fn release_hosting(_dist: &DistGraph, manifest: &DistManifest, abyss: &Gazenot) -> DistResult<()> {
+    // Perform all the releases
+    let releases = manifest.releases.iter().filter_map(|release| {
+        if let Some(cargo_dist_schema::Hosting::Abyss(set)) = &release.hosting {
+            // Upload all files associated with this Release, plus the dist-manifest.json
```
mistydemeo, line 183: Copy/pasted comment from above?

mistydemeo (approved):

cargo-dist/src/host.rs
Gankra: post, liv is out.

mistydemeo: 👍

cargo-dist/src/host.rs
Gankra: thanks!

PR #549, opened 2023-11-03
fix(config): handle missing PackageId

mistydemeo:
This is an `Option`, we shouldn't `unwrap()` in case it's actually missing.

I added an `expect()` to the version since that field's actually mandatory, but it seems like we can cope with the package ID being missing.

### DIFF
--- a/cargo-dist/src/cargo_build.rs
+++ b/cargo-dist/src/cargo_build.rs
@@ -227,7 +227,11 @@ pub fn build_cargo_target(dist_graph: &DistGraph, target: &CargoBuildStep) -> Re
         FastMap::<String, FastMap<String, Vec<(Utf8PathBuf, Utf8PathBuf)>>>::new();
     for &binary_idx in &target.expected_binaries {
         let binary = &dist_graph.binary(binary_idx);
-        let package_id = binary.pkg_id.to_string();
+        let package_id = binary
+            .pkg_id
+            .clone()
+            .expect("pkg_id is mandatory for cargo builds")
+            .to_string();
         let exe_name = binary.name.clone();
         for exe_dest in &binary.copy_exe_to {
             expected_exes
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -236,7 +236,7 @@ pub struct Binary {
     ///
     /// This is an "opaque" string that will show up in things like cargo machine-readable output,
     /// but **this is not the format that cargo -p flags expect**. Use pkg_spec for that.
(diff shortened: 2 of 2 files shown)


cargo-dist/src/tasks.rs
```diff
@@ -889,8 +889,12 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         for (pkg_idx, binary_name) in bins.clone() {
             let package = self.workspace.package(pkg_idx);
             let package_metadata = self.package_metadata(pkg_idx);
-            let version = package.version.as_ref().unwrap().cargo();
-            let pkg_id = package.cargo_package_id.clone().unwrap();
+            let version = package
+                .version
+                .as_ref()
+                .expect("Package version is mandatory!")
+                .cargo();
```
Gankra, line 896: Just checking you're aware that `.cargo()` will also panic if there's a Not Rust Project?

https://docs.rs/axoproject/latest/src/axoproject/lib.rs.html#337-344

cargo-dist/src/cargo_build.rs
```diff
@@ -227,7 +227,11 @@ pub fn build_cargo_target(dist_graph: &DistGraph, target: &CargoBuildStep) -> Re
         FastMap::<String, FastMap<String, Vec<(Utf8PathBuf, Utf8PathBuf)>>>::new();
     for &binary_idx in &target.expected_binaries {
         let binary = &dist_graph.binary(binary_idx);
-        let package_id = binary.pkg_id.to_string();
+        let package_id = if let Some(id) = &binary.pkg_id {
+            id.to_string()
+        } else {
+            binary.name.to_owned()
```
Gankra, line 233: wait will this actually work? "package_id" is very specifically a special id we need to use to communicate with rustc/cargo correctly, i believe. it can't be substituted

mistydemeo: We additionally use it as the key to the hash map of all binaries that exist. While I'll be factoring a bunch of cargo-specific stuff out, that piece of logic is independent of what the underlying builder is.

Gankra (approved):

cargo-dist/src/tasks.rs
mistydemeo: Yeah, I'll be dealing with that in other branches.

cargo-dist/src/cargo_build.rs
mistydemeo: Updated this to a simpler `expect()` in the place it's used, which should no longer overlap with non-cargo usage.

PR #553, opened 2023-11-07
feat: enable generic builds

mistydemeo:
This PR adds a new "generic" build type. Most of the process is identical to Cargo builds, but I've extended our build logic to add a new `GenericBuildStep` that encapsulates a large part of the different behaviour.

This is an MVP, and it's missing some features we likely want to see in the future:

* Only workspaces are supported; we don't have a concept of packages yet. Since we don't inherit anything like that from Cargo, we'll have to think about what this looks like.
* The user has to manually specify all of the binaries that the build output will contain, since we don't have Cargo to get that information from. In the future, we probably want to make these globs instead of literal paths so that we can accommodate more complex builds.

We defer most of the project loading and parsing logic to axoproject, like we do with Cargo projects.

Depends on https://github.com/axodotdev/axoproject/pull/45 being in a release before this is mergeable.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -48,12 +48,13 @@
 //! Also note that the BuildSteps for installers are basically monolithic "build that installer"
 //! steps to give them the freedom to do whatever they need to do.
 
+use std::process::Command;
+
 use axoproject::{PackageId, PackageIdx, WorkspaceInfo};
 use camino::Utf8PathBuf;
 use cargo_dist_schema::{DistManifest, Hosting};
 use miette::{miette, Context, IntoDiagnostic};
 use semver::Version;
-use std::process::Command;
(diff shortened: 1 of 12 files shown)


cargo-dist/src/tasks.rs
```diff
@@ -727,7 +743,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
 
         Ok(Self {
             inner: DistGraph {
-                is_init: dist_profile.is_some(),
+                is_init: desired_cargo_dist_version.is_some(),
```
mistydemeo, line 746: I changed this since `dist_profile` is no longer a decent "is this initted" proxy.

mistydemeo:
Modulo one bugfix, this is ready for review and good to go.

Gankra (approved):
continues to look totally reasonable

PR #554, opened 2023-11-07
fix(homebrew): reimplement Homebrew class code

### LINKED ISSUE #541: Class name for homebrew formula truncates `s` incorrectly
Inputting app name as `tetanes` results in `Tetane`

Seems like it's attempting to singularize the class name - but there are a lot of words that end with `s`.

Not sure if this has any impact on the actual homebrew formula installation or not.

### LINKED ISSUE #551: to_class_case implementation different than from what Homebrew expects
I am trying to use `cargo-dist` for my program called [mni2mz3](https://github.com/FNNDSC/mni2mz3) however the homebrew installer is not working.

https://github.com/axodotdev/cargo-dist/blob/c10eb4625374f0c2b673d6ad3961f3a851e6b77d/cargo-dist/src/tasks.rs#L1419

The return of `to_class_case` is passed to the template here:

https://github.com/axodotdev/cargo-dist/blob/c10eb4625374f0c2b673d6ad3961f3a851e6b77d/cargo-dist/templates/installer/homebrew.rb.j2#L1

For the project name `mni2mz3`, the generated `formula_class` is `Mni2Mz3` however this does not work with homebrew (version 4.1.19)

```shell
$ brew install FNNDSC/homebrew-mni2mz3/mni2mz3
# -- snip --
Error: No available formula with the name "fnndsc/mni2mz3/mni2mz3".
In formula file: /usr/local/Homebrew/Library/Taps/fnndsc/homebrew-mni2mz3/Formula/mni2mz3.rb
Expected to find class Mni2mz3, but only found: Mni2Mz3.
```

mistydemeo:
We previously depended on an external crate, but its implementation wasn't 100% identical to Homebrew's. This replaces that with a custom function that implements Homebrew's logic precisely, along with a test case that covers all of Homebrew's test cases plus the two recently reported bugs.

Fixes #541.
Fixes #551.

### DIFF
--- a/cargo-dist/src/backend/installer/homebrew.rs
+++ b/cargo-dist/src/backend/installer/homebrew.rs
@@ -81,3 +81,158 @@ pub(crate) fn write_homebrew_formula(
     LocalAsset::write_new(&script, &info.inner.dest_path)?;
     Ok(())
 }
+
+/// Converts the provided app name into a Ruby class-compatible
+/// string suitable for use as the class in a Homebrew formula.
+// Homebrew implementation is Formulary.class_s:
+// https://github.com/Homebrew/brew/blob/8c7cd3c0fd46f7808e782e40359c19271f950a75/Library/Homebrew/formulary.rb#L447-L453
+pub fn to_class_case(app_name: &str) -> String {
+    if app_name.is_empty() {
+        return app_name.to_owned();
+    }
+
+    let mut out = app_name.to_owned();
+    // First, we uppercase the first character in the string
+    out[..1].make_ascii_uppercase();
+
+    let mut chars = vec![];
+    let mut iter = out.chars().peekable();
+    let mut el = iter.next();
+    let mut at_replaced = false;
+    while el.is_some() {
+        let char = el.unwrap();
+        // -, _ and . are invalid characters in Ruby classes.
+        // Homebrew handles these by stripping them, then uppercasing
+        // the following character
+        match char {
+            '-' | '_' | '.' => {
+                // Only perform a replacement if the following character is
+                // in the range [a-zA-Z0-9]
+                if let Some(next) = iter.peek() {
+                    if next.is_ascii_digit() || next.is_ascii_alphabetic() {
+                        chars.push(next.to_ascii_uppercase());
+                        iter.next();
+                    } else {
+                        chars.push(char);
+                    }
+                } else {
+                    chars.push(char);
+                }
+            }
+            // Perform an @ replacement, but only if followed by a digit
+            // We also perform this replacement only once
+            '@' => {
+                if let Some(next) = iter.peek() {
+                    if next.is_ascii_digit() && !at_replaced {
+                        chars.push('A');
+                        chars.push('T');
+                        chars.push(*next);
+                        iter.next();
+                        at_replaced = true;
+                    } else {
+                        chars.push(char);
+                    }
+                } else {
+                    chars.push(char);
+                }
+            }
+            // So that things like c++ become cxx
+            '+' => {
+                chars.push('x');
+            }
+            _ => chars.push(char),
+        }
+
+        el = iter.next();
+    }
+    chars.iter().collect()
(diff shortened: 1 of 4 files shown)


cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -81,3 +81,98 @@ pub(crate) fn write_homebrew_formula(
...
+        let char = el.unwrap();
+        // -, _ and . are invalid characters in Ruby classes.
+        // Homebrew handles these by stripping them, then uppercasing
+        // the following character
+        match char {
+            '-' | '_' | '.' => {
+                if let Some(next) = iter.next() {
+                    chars.push(next.to_ascii_uppercase());
+                }
+            }
```
Gankra, line 112: This would make `ab__c` (two underscores) be `Ab_c`, wouldn't it?

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -81,3 +81,98 @@ pub(crate) fn write_homebrew_formula(
...
+            _ => chars.push(char),
+        }
+
+        el = iter.next();
+    }
+    out = chars.iter().collect();
+    // Perform two more small replacements
+    out = out.replace('+', "x");
+    out.replace('@', "AT")
```
Gankra, line 121: The ruby impl seems a lot more pedantic about this replacement..? (although i don't think @ can show up in a crate name so moot...)

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -81,3 +81,98 @@ pub(crate) fn write_homebrew_formula(
...
+    }
+
+    #[test]
+    fn handles_pluralization() {
+        run_comparison("tetanes", "Tetanes");
+    }
+
+    #[test]
+    fn replaces_plus_with_x() {
+        run_comparison("c++", "Cxx");
```
Gankra, line 171: oh lol, that's why `+ -> x`

cargo-dist/src/backend/installer/homebrew.rs
Gankra: both of these replacements could have been done in the loop, no? Arguably this is more readable, tho.

Gankra (approved):
I think there might be an error but approving in case not

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Exciting, the Ruby impl has the same bug:

```
brew(main):002:0> Formulary.class_s("ab__c")
=> "Ab_C"
```

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -81,3 +81,98 @@ pub(crate) fn write_homebrew_formula(
...
+// https://github.com/Homebrew/brew/blob/8c7cd3c0fd46f7808e782e40359c19271f950a75/Library/Homebrew/formulary.rb#L447-L453
+pub fn to_class_case(app_name: &str) -> String {
+    if app_name.is_empty() {
+        return app_name.to_owned();
+    }
+
+    let mut out;
+    // First, we uppercase the first character in the string
+    let (first_char, rest) = app_name.split_at(1);
+    out = format!("{}{rest}", first_char.to_uppercase());
```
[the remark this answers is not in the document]
mistydemeo: Nice, I was wondering if it'd be possible to mutate the first char in-place.

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Haha, yup

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Went for a quirks-compatible approach, with a new test that checks how it stacks against Homebrew.

PR #555, opened 2023-11-08
feat(installer): add additional shell config

### LINKED ISSUE #547: Installer.sh sourcing ~/.cargo/bin in ~/.profile is not always sufficient
Hi folks! Love the project! 

I'm excited to have released my latest Rust project, a passphrase generator called Phraze, with cargo-dist v0.4.2. [Here's the first release that uses cargo-dist](https://github.com/sts10/phraze/releases/tag/v0.3.3). I checked the box for a shell script install otpino, and sure enough I get the pretty "Install prebuilt binaries via shell script" one-liner in my release notes for users to paste into their terminals: `curl --proto '=https' --tlsv1.2 -LsSf https://github.com/sts10/phraze/releases/download/v0.3.3/phraze-installer.sh | sh`

Excited to a make it easier for non-Rust devs to use my Rust tool, I booted up another machine of mine that runs Debian but doesn't have Rust or Cargo installed. (In a sense this is my targeted user for why I choose to use cargo-dist -- if the user has Cargo installed, they can just use cargo to install my program!) 

## Expected behavior 

I had my fingers crossed that, on that machine, I could simply run the `curl` one liner and then, poof, use my passphrase generator. 

Since I didn't have a `~/cargo` directory (or Cargo installed) on this particular machine, I just assumed that the script would install my executable elsewhere, likely somewhere already in this machine's PATH. 

When I ran the script, it told me the program had been installed successfully to `~/.cargo/bin`. I was suspicious since, not having Cargo installed on this machine, I knew that directory wasn't in PATH. I did also get some kind of message to `source` something... (I'm really sorry that I can't recreate this right now), but I figured it was just telling me to restart my terminal (a fair request). 

I then expected to be able to run `phraze` and have it work.

## Observed behavior
I tried running `phraze` but I got a command not found error. Even starting the terminal again, same thing: command not found. 

Just to check, when I ran `echo $PATH`, `~/.cargo/bin` was not in the PATH. 

## Suggestions

I was hoping that when the installer.sh script would have detected that `~/.cargo/bin` was NOT in the current PATH, the script would (A) install the executable to /usr/local/bin or somewhere already in the PATH; (B) automatically add `~/.cargo/bin` to the user's PATH (maybe writing the end of ~/.bashrc); or (c) Print to screen a very clear message like "To finish installation of <program name>, run: `echo "export PATH = "$HOME/.cargo/bin:$PATH" >> ~/.bashrc && source ~/.bashrc` or something. (I think the Rust install script does something like option C?) 

To generalize, I think that if the installer script installs the executable to a directory NOT currently in the PATH, it should, at minimum, print clear, next-step instructions on how to add the directory to the PATH. Otherwise a user may get/feel lost when confronting the "command not found" message when they first try to run the program.

I hope this makes sense -- in a bit of a hurry right now, but I can investigate further tomorrow.

mistydemeo:
~/.profile isn't sourced in many circumstances; we need to write this to a large number of locations to make sure it gets considered by bash and zsh under the largest number of circumstances.

Fixes #547.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -313,6 +313,28 @@ install() {
     fi
 }
 
+print_home_for_script() {
+    local script="$1"
+
+    local _home
+    case "$script" in
+        # zsh has a special ZDOTDIR directory, which if set
+        # should be considered instead of $HOME
+        .zsh*)
+            if [ -n "$ZDOTDIR" ]; then
+                _home="$ZDOTDIR"
+            else
+                _home="$HOME"
+            fi
+            ;;
+        *)
+            _home="$HOME"
+            ;;
+    esac
+
+    echo "$_home"
+}
+
 add_install_dir_to_path() {
     # Edit rcfiles ($HOME/.profile) to add install_dir to $PATH
     #
@@ -328,11 +350,13 @@ add_install_dir_to_path() {
 
     if [ -n "${HOME:-}" ]; then
         local _target
+        local _home
 
         # Find the first file in the array that exists and choose
         # that as our target to write to
         for _rcfile_relative in $_rcfiles; do
-            local _rcfile="$HOME/$_rcfile_relative"
+            _home="$(print_home_for_script "$_rcfile_relative")"
+            local _rcfile="$_home/$_rcfile_relative"
 
             if [ -f "$_rcfile" ]; then
                 _target="$_rcfile"
@@ -345,7 +369,8 @@ add_install_dir_to_path() {
         if [ -z "${_target:-}" ]; then
             local _rcfile_relative
             _rcfile_relative="$(echo "$_rcfiles" | awk '{ print $1 }')"
-            _target="$HOME/$_rcfile_relative"
+            _home="$(print_home_for_script "$_rcfile_relative")"
+            _target="$_home/$_rcfile_relative"
         fi
 
         # `source x` is an alias for `. x`, and the latter is more portable/actually-posix.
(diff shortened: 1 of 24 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -307,7 +307,10 @@ install() {
     say "everything's installed!"
 
     if [ "0" = "$NO_MODIFY_PATH" ]; then
-        add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr"
+        local _rcfiles=".profile .bash_profile .bash_login .bashrc .zshrc .zshenv"
```
mistydemeo, line 310: I originally wrote this as an array, but shellcheck didn't let me since dash doesn't support arrays.

Gankra: So although it's still kind of a shotgun blast, iirc rustup is *slightly* more surgical than this. I believe for each kind of shell they identify families of files and write to the first one that already exists, or failing that write to a "default" location.

(Also some extra logic that looks in $ZDOTDIR in addition to $HOME...)

So like you'd write to:

* up to one file to provide base posix support
* up to one file to provide zsh support
* up to one file to provide bash support

For a maximum of 3 files written for total platform coverage.

https://github.com/rust-lang/rustup/blob/b5d74ac69307de29cee586c779bdb6b3aee7f7af/src/cli/self_update/shell.rs#L193-L206

Gankra (approved):
Shippable, but see comment for how this differs from rustup

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Updated. Now we write to three sets of scripts; we pick the first script we find that already exists from each set, or create the first listed script if none of them already exist.

Gankra (approved):

PR #559, opened 2023-11-09
docs(generic): begin generic build docs

mistydemeo:
This is the start of our generic build documentation. While this covers all of the essential configuration, we may want the surrounding text to be more substantive. We also want to decide what this feature is actually called, because "generic builds" is probably not the best name.

### DIFF
--- a/book/src/generic-builds.md
+++ b/book/src/generic-builds.md
@@ -1,0 +1,80 @@
+# Generic Builds
+
+> since 0.5.0
+
+Although cargo-dist was originally designed specifically for Cargo-based builds, we've introduced experimental features to allow you to use it to build and distribute any software written in any language. This feature is currently a *prerelease*: details may change before it becomes stable.
+
+## Configuration
+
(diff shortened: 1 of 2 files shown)


book/src/generic-builds.md
```diff
@@ -0,0 +1,80 @@
+# Generic Builds
+
+> since 0.5.0
+
+Although cargo-dist was originally designed specifically for Cargo-based builds, we've introduced experimental features to allow you to use it to build and distribute any software written in any language. This feature is currently a *prerelease*: details may change before it becomes stable.
```
Gankra, line 5: > This feature is currently a *prerelease*: details may change before it becomes stable.

Is this a "change this line before cutting the release" or "this is unstable, even past the release"

Gankra (approved):

book/src/generic-builds.md
mistydemeo: We're considering the feature experimental at this point, so it's still the case that we might change some of these details at some point.

PR #567, opened 2023-11-14
feat: break tag parsing into crate

mistydemeo:
This will eventually be a separate repo so that it's usable by other packages.

### DIFF
--- a/cargo-dist/src/announce.rs
+++ b/cargo-dist/src/announce.rs
@@ -4,6 +4,7 @@
 
 use axoproject::platforms::triple_to_display_name;
 use axoproject::PackageIdx;
+use axotag::{parse_tag, Package, PartialAnnouncementTag, ReleaseType};
 use itertools::Itertools;
 use semver::Version;
 use tracing::{info, warn};
@@ -27,18 +28,6 @@ pub(crate) struct AnnouncementTag {
     pub prerelease: bool,
     /// Which packages+bins we're announcing
     pub rust_releases: Vec<(PackageIdx, Vec<String>)>,
-}
-
-/// details on what we're announcing (partially computed)
-struct PartialAnnouncementTag {
-    /// The full tag
-    pub tag: Option<String>,
-    /// The version we're announcing (if doing a unified version announcement)
-    pub version: Option<Version>,
-    /// The package we're announcing (if doing a single-package announcement)
-    pub package: Option<PackageIdx>,
-    /// whether we're prereleasing
-    pub prerelease: bool,
 }
 
 impl<'a> DistGraphBuilder<'a> {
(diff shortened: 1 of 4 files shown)


Gankra:
ahh interesting... i don't think we want abyss to gain a full dep on axoproject... but it would be annoying to strip+restore the abstraction

i forgot this would include my most hated child, (Partial)AnnouncementTag which absolutely should contain an enum instead of this "there's two options, only one is ever init" crap

if i had infinite time and was doing it:

don't let tag be an Option
make the version+package fields be properly expressed by an enum
find a way to avoid the axoproject dep (PackageIdx is just a usize at the end of the day...)
drop PackageInfo for a minimal name+version pair (pretty sure those are the only operative values for parse_tag?)

```
packages: &BTreeMap<PackageIdx, &PackageInfo>,
```
is morally a `Vec` (because PackageIdx is just a newtyped Vec index)

axotag/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub release: ReleaseType,
+    /// whether we're prereleasing
+    pub prerelease: bool,
+}
+
+impl Default for PartialAnnouncementTag {
+    /// Constructs an empty PartialAnnouncementTag
+    fn default() -> PartialAnnouncementTag {
+        PartialAnnouncementTag {
+            tag: String::new(),
```
Gankra, line 35: yeah i had been thinking about that but "using String::new as secret Option::None" always seems like it causes me a problem down the road so I didn't want to lead you down the path of my crimes (fuck it, ship it, PartialAnnouncementTag is a temp type anyway).

axotag/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            release: ReleaseType::None,
+            prerelease: false,
+        }
+    }
+}
+
+/// which type of release we're announcing
+pub enum ReleaseType {
+    /// none
+    None,
```
Gankra, line 45: oh interesting... any particular reason this works better than `Option<ReleaseType>`? The latter seems like it would be easier to work with in a lot of places -- especially since I think Announcement wants this to be non-None (i think...? maybe I'm wrong. possibly specifically in the case of needs_coherent = false)

cargo-dist/src/announce.rs
```diff
@@ -357,117 +354,29 @@ pub(crate) fn select_tag(
...
+
     Ok(AnnouncementTag {
-        tag: announcing
-            .tag
-            .expect("integrity error: failed to select announcement tag"),
-        version: announcing.version,
-        package: announcing.package,
+        tag: announcing.tag,
+        version,
+        package,
```
Gankra, line 374: oh i see you unpack them back to the Evil format to minimize churn.

Gankra (approved):
looks great

cargo-dist/src/announce.rs
mistydemeo: Yeah, eventually we can do a refactor here where the upstream type becomes less evil, but this gets us unblocked today.

PR #574, opened 2023-11-17
Expose setting to set arbitrary environment variables

### LINKED ISSUE #513: `RUSTFLAGS` via cargo config
I tried to integrate `cargo-dist` into my release workflow and hit a roadblock in form of my required rustflags.

In my cargo config I set a rustflag to compile the `uuid` crate with unstable features to be able to use UUIDv7.

`cargo-dist`'s usage of the environment variable completely overwrites the cargo configuration though.

Right now I'm resorting to allowing a dirty CI state and adding the `RUSTFLAGS` env-var with the appropriate value, since `cargo-dist` preserves those.

---

I personally see two possible solutions to this:

1. `cargo-dist` parses the cargo config and appends any rustflags to the environment variable it sets
2. `cargo-dist` adds the ability to add arbitrary environment variables 

Both of these would be nice features tbh

aumetra:
This PR implements the `environment-variables` field under `[workspace.metadata.dist]`, allowing arbitrary environment variables to be set during the CI runs.

These are set via the top-level `env` field in GitHub Actions.

(Kinda) closes #513 

---

If you look at the new snapshot, you might see that the key of the YAML is also quoted, I was thinking about how to get rid of that before remembering that YAML is *horrible* and actually just allows just about any horribleness you could think of.  
And since YAML is quite literally a JSON superset, quoted keys should be 100% a-okay.

So yeah, that should be fine.

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -5,6 +5,7 @@
 use axoasset::LocalAsset;
 use cargo_dist_schema::{GithubMatrix, GithubMatrixEntry};
 use serde::Serialize;
+use std::collections::HashMap;
 use tracing::warn;
 
 use crate::{
@@ -46,6 +47,8 @@ pub struct GithubCiInfo {
     pub ssldotcom_windows_sign: Option<ProductionMode>,
     /// what hosting provider we're using
     pub hosting_provider: HostingStyle,
+    /// Environment variables set during the build
+    pub environment_variables: Option<HashMap<String, String>>,
 }
 
 impl GithubCiInfo {
@@ -63,6 +66,8 @@ impl GithubCiInfo {
         let fail_fast = dist.fail_fast;
         let create_release = dist.create_release;
         let ssldotcom_windows_sign = dist.ssldotcom_windows_sign.clone();
+        let environment_variables = dist.environment_variables.clone();
+
         let mut dependencies = SystemDependencies::default();
 
         // Figure out what builds we need to do
@@ -142,6 +147,7 @@ impl GithubCiInfo {
             create_release,
             ssldotcom_windows_sign,
             hosting_provider,
+            environment_variables,
         }
     }
 
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -1,6 +1,6 @@
 //! Config types (for workspace.metadata.dist)
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use axoasset::{toml_edit, SourceFile};
 use axoproject::{WorkspaceKind, WorkspaceSearch};
@@ -272,6 +272,10 @@ pub struct DistMetadata {
     /// Hosting provider
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hosting: Option<HostingStyle>,
+
+    /// User-defined environment variables
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub environment_variables: Option<HashMap<String, String>>,
 }
 
 impl DistMetadata {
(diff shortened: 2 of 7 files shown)


Gankra:
for the record, the magic incantation to avoid quotes is `|safe`:

https://github.com/axodotdev/cargo-dist/blob/9be2d30c4ac6ca58c0aa1cdb8510358121da569b/cargo-dist/templates/ci/github_ci.yml.j2#L345

but yeah we avoid using that unless absolutely necessary because ✨reliability

aumetra:
> for the record, the magic incantation to avoid quotes is `|safe`

Oh, huh.. I never used `minijinja` before (or `jinja` in general), that's definitely good to know

aumetra:
> Just checking, is there any reason why setting these env-vars locally, and not only-in-ci would be an issue for your usecase?

There shouldn't be, no. I was just very laser-focussed on getting it to work the way I patched it into my workflow right now, forgetting about Potential other solutions, heh

---

Just a small question about the parts you linked: the first linked piece of code builds a command that then invokes `cargo-dist` again, running the second part of the code, right?

I just wanna make sure I understood correctly (sorry if the question is a little silly, GitHub mobile's Code browsing UX isn't the best..  and it's 4AM for me 🧍‍♀️)

Gankra:
> Just a small question about the parts you linked: the first linked piece of code builds a command that then invokes cargo-dist again, running the second part of the code, right?

Sorry about the delay, similar also hit by "it's late and weekend". So no, the two linked pieces of code are parallel paths in cargo-dist. The first one is the new "generic build" backend, which exists for invoking things like makefiles. The second one is the original "cargo" back end which is building up an invocation of "cargo build" (and any other tools).

I'll admit it's a bit more complex to merge random env-vars here than set them in the CI itself... hmm I wonder if we can get away with mutating our own process' environment earlier on...

mistydemeo (commented):
Sorry for the delay reviewing!

I agree with Gankra that it'd be nicer to help ensure we set these environment variables just for the subprocesses, and in a way that ensures non-CI builds get the same benefits. The good news is, we already do this in a pretty similar way in both Cargo and generic builds, so it won't require much of a change. What you'll want to do is, right under these two lines:

https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/cargo_build.rs#L194
https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/generic_build.rs#L109

Insert something like this:

```rust
    if let Some(environment) = &dist_graph.environment_variables {
        command.envs(environment);
    }
```

That'll add everything in that environment config to the environment that we use to call `cargo build` or the generic build.

Does this cover everything you're looking for? Or are you also looking to set `RUSTFLAGS`? (And want to set that via the environment rather than via `.cargo/config.toml`?)

aumetra:
> Or are you also looking to set RUSTFLAGS? (And want to set that via the environment rather than via .cargo/config.toml?)

Actually, that was the main motivator behind this PR. `cargo-dist` sets some RUSTFLAGS (last time I checked) via the environment, which then overwrites the flags set in the cargo config.

But `cargo-dist` preserves the environment variable. Hence this PR that allows me to set the RUSTFLAGS via the environment instead.

mistydemeo:
Got it! That makes sense. In that case, what about this - on this line: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/cargo_build.rs#L33

You can add a new check to query these environment variables you've added. In that case, I'd say the priorities should look like this:

* First, prefer the environment (like now)
* Then environment configured via `Cargo.toml`
* Then the default

Longterm, we should probably make sure that we're not overriding cargo config. That does actually sound like a bug.

Gankra:
Some variant of this would be good for https://github.com/axodotdev/cargo-dist/issues/788

Gankra:
oh haha, this was much more "i dropped the ball and should help you" than vice versa, no worries :)

PR #578, opened 2023-11-18
Remove rust-toolchain-version from the workspaces setup guide

rukai:
We should not mention `rust-toolchain-version` in the setup guide since cargo-dist has not generated this field since 0.1.0
However it remains documented as a deprecated config in the reference at https://opensource.axo.dev/cargo-dist/book/reference/config.html#rust-toolchain-version


Gankra (approved):
oh WOW i forgot the docs had this level of detail, oh boy yeah some New Stuff is worth adding, thanks!

PR #596, opened 2023-11-21
feat: cleanup github releases / ci contents

Gankra:
Kind of a grab-bag:

* rename tasks to match the conceptual model now that it's real
  * should-publish => host
  * announce-release/publish-release => announce
* clean up comments, add another test
* create an empty Release for a library-release, allowing everything to know it should be hosted/published/announced
  * in theory this means library-releases to axo hosting should Now Work, but, not yet tested
* github Releases now get the pretty Release urls from axo hosting
  * when performing `host --steps=release` we will update the artifact download urls with the result
    * only the axodotdev url will be updated, if github hosting also is enabled it will still refer to the github release
    * but we prefer the axodotdev url in multi-host-drifting so that's mostly academic
    * we also will find-and-replace the artifact download url in all install_hints
    * we will also regenerate the github releases announcement body to include the new hints
  * in CI we now save the dist-manifest at the end of the host job (which includes `--steps=release`)
    * it is written back to scratch storage (but *not* persistent storage on Abyss, which is logically Frozen at that point)
      * this propagates the updated URLs to announce/publish steps
      * this is actually technically Useless in the current impl but I'm sure we'll want it later so let's just do it now 
    * it is saved to an output of the host step in the same way the dist-manifest of "plan" is
  * in CI we now have the github announcement read from `host.val` instead of `plan.val` to get the new release notes
* PR includes prereleases 7, 8, and 9 for testing

### DIFF
--- a/cargo-dist/src/host.rs
+++ b/cargo-dist/src/host.rs
@@ -1,7 +1,7 @@
 //! Details for hosting artifacts
 
 use crate::{
-    announce::AnnouncementTag,
+    announce::{announcement_axodotdev, announcement_github, AnnouncementTag},
     check_integrity,
     config::{CiStyle, Config, HostArgs, HostStyle, HostingStyle},
     errors::{DistResult, Result},
@@ -23,7 +23,7 @@ pub fn do_host(cfg: &Config, host_args: HostArgs) -> Result<DistManifest> {
         create_hosting: host_args.steps.contains(&HostStyle::Create),
         ..cfg.clone()
     };
-    let (dist, manifest) = gather_work(&cfg)?;
+    let (dist, mut manifest) = gather_work(&cfg)?;
 
     // The rest of the steps are more self-contained
 
@@ -40,7 +40,7 @@ pub fn do_host(cfg: &Config, host_args: HostArgs) -> Result<DistManifest> {
                         upload_to_hosting(&dist, &manifest, &abyss)?;
                     }
                     if host_args.steps.contains(&HostStyle::Release) {
-                        release_hosting(&dist, &manifest, &abyss)?;
+                        release_hosting(&dist, &mut manifest, &abyss)?;
                     }
                     if host_args.steps.contains(&HostStyle::Announce) {
                         announce_hosting(&dist, &manifest, &abyss)?;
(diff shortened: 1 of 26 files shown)


cargo-dist/src/host.rs
```diff
@@ -208,7 +212,21 @@ fn release_hosting(_dist: &DistGraph, manifest: &DistManifest, abyss: &Gazenot)
...
+    // Update artifact download URLs with release results
+    for new_release in new_releases {
+        if let Some(new_url) = new_release.release_download_url {
+            manifest.update_release_axodotdev_artifact_download_url(&new_release.package, new_url);
+        }
+    }
+
+    // Update Github Announcement body with new URLs
+    announcement_github(manifest);
+
```
Gankra, line 229: Note to self: check if `should-publish` CI (the one that runs upload+release) actually reuploads the dist-manifest to scratch storage so that `announce` would see this update...

cargo-dist/src/host.rs
```diff
@@ -208,7 +212,21 @@ fn release_hosting(_dist: &DistGraph, manifest: &DistManifest, abyss: &Gazenot)
...
+    // Update artifact download URLs with release results
+    for new_release in new_releases {
+        if let Some(new_url) = new_release.release_download_url {
+            manifest.update_release_axodotdev_artifact_download_url(&new_release.package, new_url);
+        }
+    }
+
+    // Update Github Announcement body with new URLs
+    announcement_github(manifest);
```
mistydemeo, line 228: This is failing the test because the new logic is producing an empty GitHub release announcement.

The test in question is producing an announcement for `cargo-dist-schema`. At this point, we've filtered the list of releases down to zero - we're not releasing `cargo-dist`, and `cargo-dist-schema` has no binaries. As a result, `announcement_github()` is being called on a manifest that has zero releases in it. Whereas before you checked the global `self.inner.ci_style`, which would still be valid, here you're checking the manifest's releases and concluding that `announcing_github` should be `false`.

That... might be right, actually? Up to you whether this is actually more correct than the old behaviour.

Gankra:
cut pre7 on this branch, and used it to make this release:

https://github.com/Gankra/evil-workspace/releases/tag/v0.8.3

Gankra:
updated first comment to explain wtf i've done in here

mistydemeo (approved):

cargo-dist/tests/snapshots/lib_manifest.snap
```diff
@@ -24,5 +35,4 @@ stdout:
 }
 
 stderr:
- WARN You're trying to explicitly Release a library, only minimal functionality will work
```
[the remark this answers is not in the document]
Gankra: Yeah i think it's fine to drop it, since this PR makes it work... More.

PR #604, opened 2023-11-23
feat: add source tarballs

mistydemeo:
This adds a new global artifact kind: source tarballs. This ensures that a stable tarball gets added to releases, wherever those are hosted. They're generated using `git archive`, so they're semantically equivalent to GitHub's auto-generated tarballs (but not byte-for-byte identical).

Do we want to be able to turn this off, or always have it on?

~~The tests are currently failing because we don't actually have access to the appropriate git tags to run the archives. I should probably try to get us to skip the source for that.~~ Fixed by updating the tests to fetch tags.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -298,6 +298,8 @@ pub enum BuildStep {
     Zip(ZipDirStep),
     /// Generate some kind of installer
     GenerateInstaller(InstallerImpl),
+    /// Generates a source tarball
+    GenerateSourceTarball(SourceTarballStep),
     /// Checksum a file
     Checksum(ChecksumImpl),
     // FIXME: For macos universal builds we'll want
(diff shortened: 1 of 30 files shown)


mistydemeo:
Actions run with source: https://github.com/mistydemeo/cargodisttest/actions/runs/6974435346

Release with source: https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.57

cargo-dist/src/tasks.rs
```diff
@@ -1025,6 +1053,61 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+
+        let dist_dir = &self.inner.dist_dir.to_owned();
+
+        let filename = "source.tar.gz".to_owned();
+        let target_path = dist_dir.join(&filename);
+        let prefix = format!("{}-{}/", release.app_name, release.version);
+
+        let artifact = Artifact {
+            id: filename.to_owned(),
+            target_triples: vec![],
```
Gankra, line 1073: ...or is it all targets...? (probably irrelevant either way)

Gankra (approved):

PR #613, opened 2023-11-28
feat: extra build artifacts

### LINKED ISSUE #349: Additional non-Rust artifacts in the same Github release
Is there a way to insert additional workflow steps between Github release creation and marking it as non-draft, besides editing this into `release.yml` each time I regenerate it? This would be useful to upload additional (e.g. non-Rust) artifacts to the same release.

I believe that for my purposes the ability to copy additional jobs from a template into `release.yml` and making the final `publish-release` depend on their success would be sufficient. Does something like this already exist, or does it sound like a worthwhile addition?

mistydemeo:
This adds support for specifying arbitrary extra artifacts alongside the primary builds. These artifacts are currently expected to be architecture-independent and are uploaded as their own build artifacts rather than bundled into release tarballs. For example, we can use this to replace our current `append-release` code in to generate and upload the `dist-manifest-schema.json`.

I realized that most of our generic build code is very well-suited to this, so I split the existing generic builds into half. The code that runs the builds is now shared between this and the generic build path, while another section is separate for implementation detail reasons. There are two main distinctions between these and generic builds:

* Generic builds are oriented around *binaries*, which will eventually be collated into an artifact. Extra artifacts produce artifacts directly, and have no associated binaries.
* Generic builds (usually) build native code and have target triples; as a result, we get N copies of the generic build, one per desired triple. Extra artifacts do not have target triples, and we produce a single copy per release.

The syntax looks like this:

```toml
[[workspace.metadata.dist.extra-artifacts]]
artifacts = ["dist-manifest-schema.json"]
build = ["cargo", "dist", "manifest-schema", "--output=dist-manifest-schema.json"]
```

There can be arbitrary numbers of extra artifacts per build, so this is repeatable within a `Cargo.toml`/`dist.toml` as many times as needed.

Closes #349.

### DIFF
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -374,7 +374,11 @@ pub enum OutputFormat {
 }
 
 #[derive(Args, Clone, Debug)]
-pub struct ManifestSchemaArgs {}
+pub struct ManifestSchemaArgs {
+    /// Write the manifest schema to the named file instead of stdout
+    #[clap(long)]
+    pub output: Option<String>,
+}
 
 #[derive(Args, Clone, Debug)]
 pub struct HostArgs {
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -272,6 +272,10 @@ pub struct DistMetadata {
     /// Hosting provider
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hosting: Option<Vec<HostingStyle>>,
+
+    /// Any extra artifacts and their buildscripts
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extra_artifacts: Option<Vec<ExtraArtifact>>,
 }
 
 impl DistMetadata {
@@ -308,6 +312,7 @@ impl DistMetadata {
             ssldotcom_windows_sign: _,
             msvc_crt_static: _,
             hosting: _,
+            extra_artifacts: _,
         } = self;
         if let Some(include) = include {
             for include in include {
@@ -353,6 +358,7 @@ impl DistMetadata {
             ssldotcom_windows_sign,
             msvc_crt_static,
             hosting,
+            extra_artifacts,
         } = self;
 
         // Check for global settings on local packages
@@ -443,6 +449,9 @@ impl DistMetadata {
         }
         if publish_jobs.is_none() {
             *publish_jobs = workspace_config.publish_jobs.clone();
+        }
+        if extra_artifacts.is_none() {
+            *extra_artifacts = workspace_config.extra_artifacts.clone();
         }
 
         // This was historically implemented as extend, but I'm not convinced the
@@ -1032,6 +1041,16 @@ pub enum ProductionMode {
     Test,
     /// production mode
     Prod,
+}
+
+/// An extra artifact to upload alongside the release tarballs,
+/// and the build command which produces it.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct ExtraArtifact {
+    /// The build command to invoke
+    pub build: Vec<String>,
+    /// The artifact(s) produced via this build script
+    pub artifacts: Vec<String>,
 }
 
 impl std::fmt::Display for ProductionMode {
--- a/cargo-dist/src/generic_build.rs
+++ b/cargo-dist/src/generic_build.rs
@@ -3,7 +3,7 @@
 use std::{
     env,
     io::{stderr, Write},
-    process::Command,
+    process::{Command, Output},
 };
 
 use camino::Utf8Path;
@@ -14,7 +14,8 @@ use crate::{
     copy_file,
     env::{calculate_cflags, calculate_ldflags, fetch_brew_env, parse_env, select_brew_env},
     errors::Result,
-    BinaryIdx, BuildStep, DistGraph, DistGraphBuilder, GenericBuildStep, SortedMap, TargetTriple,
+    BinaryIdx, BuildStep, DistGraph, DistGraphBuilder, ExtraBuildStep, GenericBuildStep, SortedMap,
+    TargetTriple,
 };
 
 impl<'a> DistGraphBuilder<'a> {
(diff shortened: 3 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         with:
           submodules: recursive
       - name: Install cargo-dist
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.5.0/cargo-dist-installer.sh | sh"
+        run: "cargo install --git https://github.com/axodotdev/cargo-dist/ --branch=extra-artifacts cargo-dist"
```
Gankra, line 62: oroboros....

cargo-dist/src/cli.rs
```diff
@@ -374,7 +374,10 @@ pub enum OutputFormat {
 }
 
 #[derive(Args, Clone, Debug)]
-pub struct ManifestSchemaArgs {}
+pub struct ManifestSchemaArgs {
+    #[clap(long)]
+    pub output: Option<String>,
```
Gankra, line 379: needs docs

cargo-dist/src/config.rs
```diff
@@ -397,6 +403,9 @@ impl DistMetadata {
         if hosting.is_some() {
             warn!("package.metadata.dist.hosting is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
         }
+        if extra_artifacts.is_some() {
+            warn!("package.metadata.dist.extra_artifacts is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
+        }
```
Gankra, line 408: oh interesting, i would have assumed this was package-specific (*vaguely recalls some discussion we had about this but forgets the conclusion*)

cargo-dist/src/generic_build.rs
```diff
@@ -132,3 +148,37 @@ pub fn build_generic_target(dist_graph: &DistGraph, target: &GenericBuildStep) -
...
+
+    let result = run_build(dist_graph, &target.build_command, None)?;
+    let dest = dist_graph.dist_dir.to_owned();
+
+    if !result.status.success() {
+        println!("Build exited non-zero: {}", result.status);
+    }
+    eprintln!();
+    eprintln!("stdout:");
+    stderr().write_all(&result.stdout).into_diagnostic()?;
```
Gankra, line 168: this is a weird juxtaposition... but ok i think i see why

cargo-dist/src/tasks.rs
```diff
@@ -1053,11 +1079,40 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+                    id: filename.to_owned(),
+                    target_triples: vec![],
+                    file_path: target_path.to_owned(),
+                    required_binaries: FastMap::new(),
+                    archive: None,
+                    kind: ArtifactKind::ExtraArtifact(ExtraArtifactImpl {
+                        build: extra.build.to_owned(),
+                        artifact: target_path.to_owned(),
+                    }),
+                    checksum: None,
```
Gankra, line 1103: hmm i imagine we'll add a flag for that one day

cargo-dist/src/tasks.rs
```diff
@@ -2218,6 +2290,9 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             // Always add the source tarball
             self.add_source_tarball(&announcing.tag, release);
 
+            // Add any extra artifacts defined in the config
+            self.add_extra_artifacts(release);
+
```
Gankra, line 2295: wait so it's obligatorily-workspace-global but added to every Release anyway? At that point feels like it should be per-packageable...

Gankra (approved):
Seems right, just a question of the scope of the config

cargo-dist/src/tasks.rs
mistydemeo: I can update the config to try allowing it per-package.

cargo-dist/src/main.rs
```diff
@@ -492,10 +492,15 @@ fn print_help_markdown(out: &mut dyn Write) -> std::io::Result<()> {
...
     _config: &Cli,
-    _args: &cli::ManifestSchemaArgs,
+    args: &cli::ManifestSchemaArgs,
 ) -> Result<(), miette::ErrReport> {
     let schema = cargo_dist_schema::DistManifest::json_schema();
     let json_schema = serde_json::to_string_pretty(&schema).expect("failed to stringify schema!?");
-    println!("{json_schema}");
+
+    if let Some(destination) = args.output.to_owned() {
+        std::fs::write(destination, json_schema + "\n").into_diagnostic()?;
```
Gankra, line 501: should be using axoasset::LocalAsset to write the file, forget the exact API.

cargo-dist/src/tasks.rs
```diff
@@ -1053,11 +1079,40 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         }
     }
 
-    fn add_source_tarball(&mut self, _tag: &str, to_release: ReleaseIdx) {
+    fn add_extra_artifacts(&mut self, to_release: ReleaseIdx) {
         if !self.global_artifacts_enabled() {
             return;
         }
+        let dist_dir = &self.inner.dist_dir.to_owned();
+        let artifacts = &self.inner.extra_artifacts.clone();
```
Gankra, line 1087: this needs to be pulling the value off of the release now (there's that one function where we grab all the package configs and slap them onto a Release)

Gankra: also do we know how this would behave in e.g. evil-workspace? like does it run N times, or once-per-package, or..?

Gankra (approved):

cargo-dist/src/tasks.rs
mistydemeo: Tested with evil-workspace. We get one per package if you define it globally, but if you define it per-package you just get the one scoped to that package.

cargo-dist/src/generic_build.rs
mistydemeo: This is for consistency with the cargo builds, where we have things which consume the underlying build's stdout. I'd run into cases where allowing the output to go to stdout caused other parts of our build process to fail. (This code is actually from the original generic builds code, just duplicated here.)

Issue #324, opened 2023-08-11
Support for Linux Arm


gregorni:
It would be cool to have support for Linux on Arm64/Aarch64 devices with the `aarch64-unknown-linux-gnu` target, along with the current x86_64 support.

This would allow building for the Raspberry Pi, Asahi Linux systems, Aarch64 Flatpaks, Pine64 devices, and so on.

zahash:
Yes, please. I also have a raspberry pi and right now cargo-dist doesn't build for aarch64-unknown-linux-gnu

mistydemeo:
This will be supported with the upcoming 0.6.0 release. We're planning to begin distributing aarch64-unknown-linux-gnu and aarch64-unknown-linux-musl binaries.

DCNick3:
Do I understand it right, that I would have to provide a linux aarch64 runner myself to be able to use it, since github still doesn't provide hosted arm linux runners and cargo-dist doesn't do cross-compilation yet?

mistydemeo:
Yes, that's right. For our own binaries, we're using BuildJet's runners. GitHub have a closed beta of paid ARM Linux runners as well: https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/

Now that we have custom runner support, I'm going to close this issue. Feel free to open issues if you have further questions though!

PR #614, opened 2023-11-28
Support custom github runners (and arm64)

milesj:
Indirectly implements: https://github.com/axodotdev/cargo-dist/issues/324

I really need linux arm64 support soon, so I thought I would give this a shot.

I originally had a setting like `arm64-linux-runner` that I would use for linux arm64 triples, but it felt very restrictive. Instead, I went with a `custom-runners` map, that _allows any triple_ to define the runner it wants to use. This is nice.

Working in plan/manifest:
Screenshot 2023-11-27 at 9 46 36 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,5 @@
 [workspace]
-members = [
-    "cargo-dist",
-    "cargo-dist-schema",
-]
+members = ["cargo-dist", "cargo-dist-schema"]
 resolver = "2"
 
 [workspace.metadata.release]
@@ -24,13 +21,30 @@ tap = "axodotdev/homebrew-tap"
 # Publish jobs to run in CI
 publish-jobs = ["homebrew"]
 # Target platforms to build apps for (Rust target-triple syntax)
-targets = ["x86_64-unknown-linux-gnu", "aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-musl", "x86_64-pc-windows-msvc"]
+targets = [
+    "x86_64-unknown-linux-gnu",
+    "aarch64-apple-darwin",
+    "x86_64-apple-darwin",
+    "x86_64-unknown-linux-musl",
+    "x86_64-pc-windows-msvc",
+    "aarch64-unknown-linux-gnu",
+]
 # Publish jobs to run in CI
 pr-run-mode = "plan"
 
 [[workspace.metadata.dist.extra-artifacts]]
 artifacts = ["dist-manifest-schema.json"]
-build = ["cargo", "run", "--", "dist", "manifest-schema", "--output=dist-manifest-schema.json"]
+build = [
+    "cargo",
+    "run",
+    "--",
+    "dist",
+    "manifest-schema",
+    "--output=dist-manifest-schema.json",
+]
+
+[workspace.metadata.dist.custom-runners]
+aarch64-unknown-linux-gnu = "buildjet-16vcpu-ubuntu-2204-arm"
 
 # The profile that 'cargo dist' will build with
 [profile.dist]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace]
 members = [
-    "cargo-dist", 
+    "cargo-dist",
     "cargo-dist-schema",
 ]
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -24,10 +21,20 @@ tap = "axodotdev/homebrew-tap"
...
+    "x86_64-apple-darwin",
+    "x86_64-unknown-linux-musl",
+    "x86_64-pc-windows-msvc",
+    "aarch64-unknown-linux-gnu",
+]
 # Publish jobs to run in CI
 pr-run-mode = "plan"
 
+[workspace.metadata.dist.custom-runners]
+aarch64-unknown-linux-gnu = "buildjet-16vcpu-ubuntu-2204-arm"
```
milesj, line 36: just testing things

cargo-dist/src/lib.rs
```diff
@@ -307,7 +307,7 @@ fn do_generate_preflight_checks(dist: &DistGraph) -> Result<()> {
             && !desired_version.pre.starts_with("github-")
             && !matches!(dist.allow_dirty, DirtyMode::AllowAll)
         {
-            return Err(miette!("you're running cargo-dist {}, but 'cargo-dist-version = {}' is set in your Cargo.toml\n\nYou should update cargo-dist-version if you want to update to this version", current_version, desired_version));
+            // return Err(miette!("you're running cargo-dist {}, but 'cargo-dist-version = {}' is set in your Cargo.toml\n\nYou should update cargo-dist-version if you want to update to this version", current_version, desired_version));
```
milesj, line 310: ignore, just testing

cargo-dist/src/backend/ci/github.rs
```diff
@@ -238,30 +242,44 @@ const GITHUB_MACOS_RUNNER: &str = "macos-11";
...
 fn install_dist_for_github_runner<'a>(
-    runner: GithubRunner,
+    runner: &'a GithubRunner,
     install_sh: &'a str,
     install_ps1: &'a str,
 ) -> &'a str {
-    if runner == GITHUB_LINUX_RUNNER || runner == GITHUB_MACOS_RUNNER {
+    dbg!(runner);
+
+    if runner.contains("linux")
```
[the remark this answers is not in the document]
mistydemeo: Yeah, I think switching from runner to triple makes sense at this point.

cargo-dist/src/config.rs
```diff
@@ -272,6 +272,10 @@ pub struct DistMetadata {
     /// Hosting provider
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hosting: Option<Vec<HostingStyle>>,
+
+    /// Custom GitHub runners, mapped by triple
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub custom_runners: Option<HashMap<String, String>>,
```
mistydemeo, line 278: Do you mind renaming this field to `github_custom_runners`? We're planning to add support for other CI backends in the future, so we'll want to be able to distinguish between GitHub runners and other systems' runners.

Alternately, we could make `custom-runners` a struct where `github` is a top-level key, inside of which is a hash like the one you've defined here.

mistydemeo (changes requested):
Thank you so much! This is looking great, just a couple comments.

cargo-dist/src/config.rs
milesj: Renamed to `github_custom_runners`.

mistydemeo:
Thank you! I'm going to do some testing, but this looks good. A couple minor requests:

* Are you up for rebasing? Sorry for introducing merge conflicts before this shipped!
* Can you undo the changes to `Cargo.toml`? I assume your editor probably autoformatted it.

milesj:
@mistydemeo Updated. Also added a test, but not sure if done correctly. I also tried updating snapshots, but it changed like almost all of them...

milesj:
@mistydemeo Got the test working.

mistydemeo (approved):
Thank you!

milesj:
Awesome 👏

mistydemeo:
This has landed in the 0.6.0 prereleases, which are ready to try now, and will be in the final 0.6.0 release.

PR #618, opened 2023-11-29
fix(ci): fail fast if installer is missing

### LINKED ISSUE #593: CI should immediately error if installing cargo-dist fails
Right now it charges into the next step, which is weird. Not a Big deal but we should do better.

Screenshot is me absent-mindedly testing a prerelease before it actually exists:

image

mistydemeo:
We already use curl's `--fail` to try to propagate failure status, but the non-zero exit gets wiped away by the pipe. The shell only sees the last process in the chain's exit status, and since `sh` receives an empty string - which is a valid `sh` script - it treats it as a success.

I was going to explicitly `set -eo pipefail`, but turns out that explicitly marking the shell as `bash` gets you that; the shell configuration is different between the inferred shell and an explicit `bash` shell. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

Fixes #593.

### DIFF
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -79,6 +79,9 @@ jobs:
         run: rustup update {{{ rust_version }}} --no-self-update && rustup default {{{ rust_version }}}
       {{%- endif %}}
       - name: Install cargo-dist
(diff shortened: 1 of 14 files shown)


Gankra (approved):
eugh gross

PR #619, opened 2023-11-29
fix(ci): add shell to other install location

mistydemeo:
I managed to miss the other place this gets called.


Gankra:
Oh i thought you were intentionally only adding it to `plan` -- we actually install it in almost every task now.

I had already been thinking we might want to start factoring out this generic cargo-dist setup logic to some of minijinja's support for template inheritance/importing (although that may require some extra work in the code we have to statically embed templates in our binary): https://docs.rs/minijinja/latest/minijinja/syntax/index.html#-include-

Gankra (approved):

PR #622, opened 2023-12-04
chore: remove stubbed init code

mistydemeo:
After experimenting with some changes here, I think I've concluded we don't want this in `init`. Otherwise, our `toml_edit` code will reformat it into the less pretty syntax:

```toml
github-custom-runners = { aarch64-unknown-linux-gnu = "buildjet-16vcpu-ubuntu-2204-arm" }
```

instead of

```toml
[workspace.metadata.dist.github-custom-runners]
aarch64-unknown-linux-gnu = "buildjet-16vcpu-ubuntu-2204-arm"
```

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -967,14 +967,6 @@ fn apply_dist_to_metadata(metadata: &mut toml_edit::Item, meta: &DistMetadata) {
         hosting.as_ref(),
     );
 
-    // NOTE: HashMap not supported by axoasset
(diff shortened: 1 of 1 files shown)


Gankra:
There aughta be a way to tell cargo-fmt which style we want (essentially picking one to be "canonical"), but yeah turning this off for now is fine (we have a similar issue with `targets` having pretty user formatting stripped, so more elbow grease in Good Formatting would be good to do next quarter).

Gankra (approved):

PR #627, opened 2023-12-05
Fix ldd broken pipe error

### LINKED ISSUE #624: Installer sh script logging `/usr/bin/ldd: line 41: printf: write error: Broken pipe` when run
Hey, when executing the installation script generated by `cargo-dist` on a GitHub Actions runner, the following is being logged:

```txt
/usr/bin/ldd: line 41: printf: write error: Broken pipe
/usr/bin/ldd: line 43: printf: write error: Broken pipe
```

You can see it in action here: https://github.com/PyO3/pyo3/actions/runs/7093258442/job/19306303498?pr=3627#step:8:52

It pollutes the output log unnecessarily, but It does not impact the execution of the script though.

I was able to narrow it down to the following line:
https://github.com/axodotdev/cargo-dist/blob/41e79b10533c10c85a0c0bc5fb556e096c2c6357/cargo-dist/templates/installer/installer.sh.j2#L515-L515

By doing some more testing, it seems that running `ldd --version | head -1` in a GitHub Actions runner will work fine but outputs some "Broken pipe" lines to stderr. See https://github.com/CodSpeedHQ/action/actions/runs/7099719565/job/19324314497?pr=90#step:3:439 for an example.

Is there a way to not have those anymore?

Thanks!

rotu:
Fixes #624
Previously, there would be benign broken pipe error due to the interaction between `ldd` and `head`. Now that error is no more.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -512,7 +512,7 @@ get_architecture() {
         else
             # Parsing version out from line 1 like:
             # ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35
-            _local_glibc="$(ldd --version | head -1 | awk -F' ' '{ print $NF }')"
+            _local_glibc="$(ldd --version | awk -F' ' '{ if (FNR<=1) print $NF }')"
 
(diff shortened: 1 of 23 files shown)


mistydemeo (approved):
Thanks! Nice catch. Removing an extra process from the pipe is nice anyway.

PR #635, opened 2023-12-09
Fix broken links

LukeMathWalker:


mistydemeo (approved):
Thanks! Although those links still work, we definitely don't advertise those URLs anymore.

PR #636, opened 2023-12-11
new: Always run `rustup target add` during build steps.

milesj:
This is to support the recent custom runners feature and to also ensure non-builtin (aarch64) targets exist.

Here's an example running `cargo dist plan` and debugging the local build steps list.

Screenshot 2023-12-10 at 5 16 16 PM

### DIFF
--- a/cargo-dist/src/cargo_build.rs
+++ b/cargo-dist/src/cargo_build.rs
@@ -31,7 +31,6 @@ impl<'a> DistGraphBuilder<'a> {
         let mut builds = vec![];
         for (target, binaries) in targets {
             let mut rustflags = std::env::var("RUSTFLAGS").unwrap_or_default();
-            let mut added_rustup_step = false;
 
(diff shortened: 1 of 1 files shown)


cargo-dist/src/cargo_build.rs
```diff
@@ -60,35 +60,30 @@ impl<'a> DistGraphBuilder<'a> {
...
-            if target.ends_with("linux-musl")
-                && self.inner.tools.cargo.host_target.ends_with("linux-gnu")
-            {
-                if let Some(rustup) = self.inner.tools.rustup.clone() {
-                    builds.push(BuildStep::Rustup(RustupStep {
-                        rustup,
-                        target: target.clone(),
-                    }));
-                } else {
-                    warn!("You're trying to cross-compile for musl from glibc, but I can't find rustup to ensure you have the rust toolchains for it!")
```
milesj, line 90: moved this up to the other linux-musl check

PR #638, opened 2023-12-11
feat: add aarch64 custom runner and builds

Gankra:


Cargo.toml
```diff
@@ -14,7 +14,7 @@ publish = false
 # Config for 'cargo dist'
 [workspace.metadata.dist]
 # The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
-cargo-dist-version = "0.6.0-prerelease.1"
+cargo-dist-version = "0.6.0-github-main"
```
mistydemeo, line 17: Do we want `main` and not this branch itself?

mistydemeo (approved):

Cargo.toml
Gankra: "main" here is almost exactly prerelease.2 so we're essentially keeping the bootstrap chain in tact with this (except for like one tokio update).

PR #642, opened 2023-12-11
feat: add package manager source to library

mistydemeo:
This makes it easier to track which package manager a package comes from for non-Homebrew packages, which could be located in multiple places. I've also added a new method to get a flat list of package name strings from the linkage data.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -626,6 +626,17 @@ pub struct Linkage {
     pub frameworks: SortedSet<Library>,
 }
 
+/// Represents the package manager a library was installed by
+#[derive(
+    Clone, Copy, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq, PartialOrd, Ord,
+)]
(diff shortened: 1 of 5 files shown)


ashleygwilliams:
@Gankra and i discussed, want to circle up with @mistydemeo when back from vacation to get this in

Gankra:
Whoops i rebased the wrong PR, don't mind me.

mistydemeo:
This passes `cargo test` and `cargo clippy`, so we can get it in any time.

ashleygwilliams:
can we make a call to get this in or close this? it's been open for a while and it'd be good to make a decision on it.

mistydemeo (commented):
We don't have a use in mind for this at the moment, but I do think it'd be useful to track. I can get it updated with the recent changes if we agree on getting it in.

ashleygwilliams:
@mistydemeo i think it's a really good idea so the next extra second we have, let's try to get this in!

ashleygwilliams (approved):

PR #644, opened 2023-12-12
feat(dockerimage): add a docker image installer

Gankra:
This creates a Local (as in one-per-platform) docker installer (which only runs for x64 gnu linux) that builds a docker image that vendors the binaries, exporting to a tarball.

This does not currently setup docker buildx in CI (so might fail). This does not currently support publishing the docker image anywhere.

### DIFF
--- a/cargo-dist/src/backend/installer/docker.rs
+++ b/cargo-dist/src/backend/installer/docker.rs
@@ -2,7 +2,7 @@
 //!
 //! In the future this may get split up into submodules.
 
-use std::process::Command;
+use std::{io::Write, process::Command};
 
 use axoasset::LocalAsset;
 use camino::Utf8PathBuf;
@@ -151,6 +151,7 @@ impl DockerInstallerInfo {
         let Some(docker) = &dist.tools.docker else {
             unreachable!("docker didn't exist, tasks.rs was supposed to catch that!");
         };
+        eprintln!("building docker image ({})", self.target);
         // 1. check if the cargo-dist docker backend is setup
         // 2. create it if not
         // 3. do a build with the cargo-dist backend
@@ -182,16 +183,25 @@ impl DockerInstallerInfo {
                 .arg("--name")
                 .arg(DIST_BUILDER)
                 .arg("--driver=docker-container")
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::inherit())
                 .current_dir(&self.package_dir);
-            let status = cmd.status().map_err(|cause| DistError::CommandFail {
+            let output = cmd.output().map_err(|cause| DistError::CommandFail {
                 command_summary: "configure docker to use the docker-container driver with 'docker buildx create'".to_owned(),
                 cause,
             })?;
 
-            if !status.success() {
+            if !output.stdout.is_empty() {
+                eprintln!();
+                eprintln!("docker stdout:");
+                std::io::stderr()
+                    .write_all(&output.stdout)
+                    .expect("failed to write to stdout");
+            }
+            if !output.status.success() {
                 return Err(DistError::CommandStatus {
                     command_summary: "configure docker to use the docker-container driver with 'docker buildx create'".to_owned(),
-                    status,
+                    status: output.status,
                 });
             }
         }
@@ -205,17 +215,26 @@ impl DockerInstallerInfo {
                 .arg(format!("--builder={DIST_BUILDER}"))
                 .arg("--output")
                 .arg(format!("type=docker,dest={}", self.file_path))
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::inherit())
                 .current_dir(&self.package_dir);
-            let status = cmd.status().map_err(|cause| DistError::CommandFail {
+            let output = cmd.output().map_err(|cause| DistError::CommandFail {
                 command_summary: "build your docker image with 'docker buildx build'".to_owned(),
                 cause,
             })?;
 
-            if !status.success() {
+            if !output.stdout.is_empty() {
+                eprintln!();
+                eprintln!("docker stdout:");
+                std::io::stderr()
+                    .write_all(&output.stdout)
+                    .expect("failed to write to stdout");
+            }
+            if !output.status.success() {
                 return Err(DistError::CommandStatus {
                     command_summary: "build your docker image with 'docker buildx build'"
                         .to_owned(),
-                    status,
+                    status: output.status,
                 });
             }
         }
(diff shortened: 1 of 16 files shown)


cargo-dist/src/backend/installer/docker.rs
```diff
@@ -0,0 +1,203 @@
...
+            .filter(|(_, package)| package.0.stage_wanted(&DependencyKind::Run))
+            .map(|(name, spec)| {
+                if let Some(version) = spec.0.version {
+                    format!("{name}={version}")
+                } else {
+                    name
+                }
+            })
+            .collect();
+        packages
```
mistydemeo, line 110: You can probably just return directly instead of assigning and returning, right?

cargo-dist/src/backend/installer/docker.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Base image (e.g. some debian distro)
+    image: String,
+    /// Apt deps to install
+    apt_deps: Vec<String>,
+    /// Additional files/dirs to copy into the image (e.g. examples)
+    includes: Vec<Utf8PathBuf>,
+}
+
+/// As of this rewriting this is the latest debian release
+const RUNNER_IMAGE: &str = "debian:bookworm-slim";
```
mistydemeo, line 68: Might be worth trying to sync this with the builder at this point, instead of using a different Debian flavour. In particular, since you're making use of apt dependency versions, there's a very real chance the specific versions won't line up between the Ubuntu builder and this version of Debian.

cargo-dist/src/backend/installer/docker.rs
```diff
@@ -0,0 +1,203 @@
...
+        //
+        // > ERROR: Docker exporter feature is currently not supported for docker driver.
+        // > Please switch to a different driver (eg. "docker buildx create --use")
+        //
+        // So for now just use `save` which Does work
+        {
+            let mut cmd = Command::new(&docker.cmd);
+            cmd.arg("build")
+                .arg(".")
+                .arg("-t")
```
mistydemeo, line 167: Let's use long flags here to make the code more readable.

cargo-dist/src/backend/installer/docker.rs
```diff
@@ -0,0 +1,203 @@
...
+        let Some(docker) = &dist.tools.docker else {
+            unreachable!("docker didn't exist, tasks.rs was supposed to catch that!");
+        };
+        // FIXME: in theory we can do "build" and "save" in one shot with
+        //
+        //    docker buildx build --output type=docker,dest=/path/to/output.tar
+        //
+        // which is what the Github Action seems to do, but locally (in WSL) i get:
+        //
+        // > ERROR: Docker exporter feature is currently not supported for docker driver.
```
mistydemeo, line 159: I haven't tested, but I believe if you specify the driver to use in the `buildx` command, you can specify `docker-container` and get something that works in one command.

cargo-dist/src/tasks.rs
```diff
@@ -2184,6 +2188,93 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            return Ok(());
+        }
+
+        // Clone info we need from the release to avoid borrowing across the loop
+        let release = self.release(to_release);
+        let variants = release.variants.clone();
+        let checksum = release.checksum;
+        let app_name = &release.app_name;
+        let app_version = &release.version;
+        let tag = format!("{app_name}/{app_version}");
```
mistydemeo, line 2208: `:` is the common separator between image name and version. eg `ubuntu:22.04`

mistydemeo (commented):
Generally looking good; left a few comments.

cargo-dist/src/backend/installer/docker.rs
Gankra: can you elaborate?

cargo-dist/src/config.rs
```diff
@@ -602,6 +602,8 @@ pub enum InstallerStyle {
     Homebrew,
     /// Generate an msi installer that embeds the binary
     Msi,
+    /// Generate a docker image for x86_64-unknown-linux-gnu
```
mistydemeo, line 605: Is the only limit what's available in CI right now?

cargo-dist/src/tasks.rs
```diff
@@ -2184,6 +2188,93 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+
+        // Clone info we need from the release to avoid borrowing across the loop
+        let release = self.release(to_release);
+        let variants = release.variants.clone();
+        let checksum = release.checksum;
+        let app_name = &release.app_name;
+        let app_version = &release.version;
+        let tag = format!("{app_name}:{app_version}");
+
+        // Make an msi for every windows platform
```
mistydemeo, line 2210: Copy/pasted comment?

mistydemeo (approved):

cargo-dist/src/config.rs
Gankra: kinda a semantic awkwardness that i'm not sure how to resolve:

* want to be able to generate docker images for multiple platforms
* don't want to have "turning on the docker installer" generate a million docker images

simplest solution i could think of was "it's per-platform but only one platform supports it until we have better answers"...

cargo-dist/src/backend/installer/docker.rs
```diff
@@ -0,0 +1,224 @@
...
+        // We do this because the kind of build+export we want to do requires a
+        // "docker-container" buildx builder, which doesn't exist by default. So
+        // we need to create one, and then use it. Unfortunately this is persistent
+        // and non-idempotent, so we need to check if we've already done it first.
+        //
+        // FIXME: is there any way to do this that doesn't require persistently modifying
+        // the current docker install?
+        let has_dist_builder = {
+            let mut cmd = Command::new(&docker.cmd);
+            cmd.arg("buildx")
```
mistydemeo, line 167: Missed earlier - one or more of these commands is going to write to stdout, which interferes with the build manifest that we *also* write to stdout. We ignore it in cargo builds (because `cargo build` exclusively outputs to stderr), but in generic builds I'm very cautious about avoiding writing to stdout. What I do is this:

https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/generic_build.rs#L102-L103

In other words, I allow stderr to get streamed directly to the parent process's stdout, as normal, but capture stdout to a pipe. At the end of the build, I capture and then print that stdout:

https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/generic_build.rs#L135-L139

https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/generic_build.rs#L159-L160

You'll likely want to do something similar for every command you run here.

PR #647, opened 2023-12-13
fix: skip git tarball when necessary

mistydemeo:
It's possible to run cargo-dist in something that's not a git repo; we should skip generating a tarball automatically in that case. I ran into this when creating a new test project, and noticed that `cargo build` was trying to run `git archive` even though it wasn't a git repo yet.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -1201,6 +1201,32 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         if !self.global_artifacts_enabled() {
             return;
         }
+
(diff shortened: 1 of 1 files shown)


Gankra (approved):

Gankra:
This won't work well in situations where rust isn't in the root of a git repo *but* cargo-dist is also terrible in that situation so *shrug*.

mistydemeo:
I guess we could do something like run `git rev-parse --show-toplevel` and check if git thinks we're in a repo...?

mistydemeo:
Pushed a commit to call out to git to see if we're in a git repo or not.

PR #651, opened 2023-12-14
fix(powershell): replace errors with throw

### LINKED ISSUE #484: Generated PowerShell installer prints confusing error
Hi there! Love cargo-dist!

I primarily develop on macOS and Linux but for the first time I packaged something with the PowerShell option enabled, and tested it out on a little Windows machine I have. I received the following message:

Screenshot 2023-10-16 195053

(this screenshot was taken from a freshly launched Windows Sandbox as the Windows machines I have access to now all have their ExecutionPolicy set appropriately)

Having run into ExecutionPolicy stuff before, I knew what to do, but this isn't a particularly helpful error message, and doesn't really make it clear to someone who's not knowledgeable about PowerShell what the remediation steps are. It also feels a little unexpected that it's reporting it can't "run" my app, despite this being an installer for it.

I notice, reading through the code of the generated installer, that it _is_ attempting to provide additional context, but this is never displayed, at least on a the Windows 10 installations I've tested:

```powershell
  # show notification to change execution policy:
  $allowedExecutionPolicy = @('Unrestricted', 'RemoteSigned', 'ByPass')
  If ((Get-ExecutionPolicy).ToString() -notin $allowedExecutionPolicy) {
    Write-Error "PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name."
    Write-Error "For example, to set the execution policy to 'RemoteSigned' please run :"
    Write-Error "'Set-ExecutionPolicy RemoteSigned -scope CurrentUser'"
    break
  }
```

I'm no PowerShell dev, so I don't really know what the solution is here - it does seem from a glance at Microsoft's docs like `Write-Error` is not supposed to immediately exit, but the first line is all I see across multiple machines, and now in Windows Sandbox, which shouldn't be affected by any of my system's configuration changes. Perhaps concatenating the messages into one message with embedded newlines?

In any case, thanks for the great tool! Keep it up :)

mistydemeo:
I don't know if this is correct, and I can't test myself, but this is my interpretation of the proposed fix in #484.

Fixes #484.

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -52,13 +52,11 @@ function Install-Binary($install_args) {
     Get-Help $PSCommandPath -Detailed
     Exit
   }
-  $old_erroractionpreference = $ErrorActionPreference
-  $ErrorActionPreference = 'stop'
 
   Initialize-Environment
 
   # Platform info injected by cargo-dist
-  $platforms = @{ 
+  $platforms = @{
   {%- for artifact in artifacts %}
     "{{ artifact.target_triples[0] }}" = @{
       "artifact_name" = "{{ artifact.id }}"
@@ -73,8 +71,6 @@ function Install-Binary($install_args) {
   $fetched = Download "$ArtifactDownloadUrl" $platforms
   # FIXME: add a flag that lets the user not do this step
   Invoke-Installer $fetched "$install_args"
-
-  $ErrorActionPreference = $old_erroractionpreference
 }
 
 function Get-TargetTriple() {
@@ -272,26 +268,35 @@ function Add-Path($OrigPathToAdd) {
 
 function Initialize-Environment() {
   If (($PSVersionTable.PSVersion.Major) -lt 5) {
-    Write-Error "PowerShell 5 or later is required to install $app_name."
-    Write-Error "Upgrade PowerShell: https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell"
-    break
+    throw @"
+Error: PowerShell 5 or later is required to install $app_name.
+Upgrade PowerShell:
+
+    https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell
+
+"@
   }
 
   # show notification to change execution policy:
   $allowedExecutionPolicy = @('Unrestricted', 'RemoteSigned', 'ByPass')
   If ((Get-ExecutionPolicy).ToString() -notin $allowedExecutionPolicy) {
-    Write-Error "PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name."
-    Write-Error "For example, to set the execution policy to 'RemoteSigned' please run :"
-    Write-Error "'Set-ExecutionPolicy RemoteSigned -scope CurrentUser'"
-    break
+    throw @"
+Error: PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name. For example, to set the execution policy to 'RemoteSigned' please run:
+
+    Set-ExecutionPolicy RemoteSigned -scope CurrentUser
+
+"@
   }
 
   # GitHub requires TLS 1.2
   If ([System.Enum]::GetNames([System.Net.SecurityProtocolType]) -notcontains 'Tls12') {
-    Write-Error "Installing $app_name requires at least .NET Framework 4.5"
-    Write-Error "Please download and install it first:"
-    Write-Error "https://www.microsoft.com/net/download"
-    break
+    throw @"
+Error: Installing $app_name requires at least .NET Framework 4.5
+Please download and install it first:
+
+    https://www.microsoft.com/net/download
+
+"@
   }
 }
 
@@ -307,5 +312,11 @@ function New-Temp-Dir() {
 $Null = $ArtifactDownloadUrl, $NoModifyPath, $Help
 # Make Write-Information statements be visible
 $InformationPreference = "Continue"
-Install-Binary "$Args"
 
+# The default interactive handler
+try {
+  Install-Binary "$Args"
+} catch {
+  Write-Information $_
+  exit 1
+}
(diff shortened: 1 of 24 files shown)


cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -280,9 +280,11 @@ function Initialize-Environment() {
...
   If ((Get-ExecutionPolicy).ToString() -notin $allowedExecutionPolicy) {
-    Write-Error "PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name."
-    Write-Error "For example, to set the execution policy to 'RemoteSigned' please run :"
-    Write-Error "'Set-ExecutionPolicy RemoteSigned -scope CurrentUser'"
+    throw @"
+PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name.
+For example, to set the execution policy to 'RemoteSigned' please run :
+'Set-ExecutionPolicy RemoteSigned -scope CurrentUser'
+"@
     break
```
Gankra, line 288:
```suggestion
    throw @"
Error: PowerShell requires an execution policy in [$($allowedExecutionPolicy -join ", ")] to run $app_name. For example, to set the execution policy to 'RemoteSigned' please run:

    Set-ExecutionPolicy RemoteSigned -scope CurrentUser

"@
```

Is more readable, but the interactive error handler eats the whitespace, so see other comments

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -280,9 +280,11 @@ function Initialize-Environment() {
   # show notification to change execution policy:
```
Gankra, line 280: The other Write-Errors in this function should similarly be rewritten to throws

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -58,7 +58,7 @@ function Install-Binary($install_args) {
   Initialize-Environment
```
Gankra, line 58: ErrorActionPreference no longer should be overriden above and below in this function

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -308,4 +310,3 @@ $Null = $ArtifactDownloadUrl, $NoModifyPath, $Help
 # Make Write-Information statements be visible
 $InformationPreference = "Continue"
 Install-Binary "$Args"
```
Gankra, line 312:
```suggestion

# The default interactive handler
try {
  Install-Binary "$Args"
} catch {
  Write-Host $_
}
```

Gankra: thing I need to investigate: does this make the script think it succeeded (probably), and if so, what's the least noisy way to fix that.

Gankra (changes requested):

cargo-dist/templates/installer/installer.ps1.j2
Gankra: add `exit 1` to the end of the catch to set status properly

mistydemeo: Does powershell have the same exit status concept? Can we just `exit 1` or do something similar?

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -272,25 +268,36 @@ function Add-Path($OrigPathToAdd) {
...
-    Write-Error "PowerShell 5 or later is required to install $app_name."
-    Write-Error "Upgrade PowerShell: https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell"
+    throw @"
+PowerShell 5 or later is required to install $app_name.
+Upgrade PowerShell:
+
+    https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell
+
+"@
     break
```
Gankra, line 278:
```suggestion

```

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -272,25 +268,36 @@ function Add-Path($OrigPathToAdd) {
...
-    Write-Error "Please download and install it first:"
-    Write-Error "https://www.microsoft.com/net/download"
+    throw @"
+Installing $app_name requires at least .NET Framework 4.5
+Please download and install it first:
+
+    https://www.microsoft.com/net/download
+
+"@
     break
```
Gankra, line 301:
```suggestion
```

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -272,25 +268,36 @@ function Add-Path($OrigPathToAdd) {
...
+"@
   }
 
   # GitHub requires TLS 1.2
   If ([System.Enum]::GetNames([System.Net.SecurityProtocolType]) -notcontains 'Tls12') {
-    Write-Error "Installing $app_name requires at least .NET Framework 4.5"
-    Write-Error "Please download and install it first:"
-    Write-Error "https://www.microsoft.com/net/download"
+    throw @"
+Installing $app_name requires at least .NET Framework 4.5
```
Gankra, line 295:
```suggestion
Error: Installing $app_name requires at least .NET Framework 4.5
```

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -272,25 +268,36 @@ function Add-Path($OrigPathToAdd) {
 
 function Initialize-Environment() {
   If (($PSVersionTable.PSVersion.Major) -lt 5) {
-    Write-Error "PowerShell 5 or later is required to install $app_name."
-    Write-Error "Upgrade PowerShell: https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-windows-powershell"
+    throw @"
+PowerShell 5 or later is required to install $app_name.
```
Gankra, line 272:
```suggestion
Error: PowerShell 5 or later is required to install $app_name.
```

Gankra (approved):

PR #662, opened 2023-12-20
release: 0.6.1

mistydemeo:


ashleygwilliams (approved):
i think strictly speaking i would call the liblzma thing is proably a feature, but i think this handles it well and i appreciate the thoughtful explanation- so i'm fine with us calling it a fix :)

PR #665, opened 2023-12-21
fix: properly upload the "final" dist-manifest to axo releases 

Gankra:
We were just uploading the global one, making us drop all local artifacts
(they still got *uploaded*, they just weren't in the dist-manifest).
Github Releases were still getting the right manifest because the Right
manifest would get written to disk immediately after uploading to Axo Releases.


mistydemeo (approved):
Tested and confirmed working.

PR #682, opened 2024-01-02
chore(ci): replace crates.io publish with custom job

mistydemeo:
This switches us to using custom hooks in order to publish to crates.io as a part of the normal release process instead of as a manual job.


ashleygwilliams (approved):
you love to see it!

mistydemeo:
Drafted and rebased on top of #683.

mistydemeo:
Migrated this to a publish job. Should be ready for use now!

PR #686, opened 2024-01-03
docs(ci): document custom jobs

mistydemeo:
This adds the new custom jobs to our existing documentation, with explanations of where each job runs.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -17,6 +17,7 @@
   - [symbols](./artifacts/symbols.md)
 - [CI](./ci/index.md)
   - [github](./ci/github.md)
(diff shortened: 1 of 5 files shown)


Gankra (approved):

ashleygwilliams (changes requested):
so i think there's a few things we may want to change:

- it's strange to me that in the reference the jobs are not all together, i think it'd be better if they were
- it's not clear to me what each stage has access to and/or what it passes to the next stage from the docs
- i think we want at least each stage to have a "jobs" option and then we could ideally have pre- or post- be modifiers for that

ashleygwilliams (approved):
nice! i think this is a huge improvement and will help people find the feature

PR #694, opened 2024-01-05
release: 0.7.1

Gankra:


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 Nothing Yet!
 
 
+# Version 0.7.1 (2024-01-05)
+
+This release lowers the MSRV for people who install with cargo-install.
```
ashleygwilliams, line 8: do we know what version it lowers it too? my suspicion is that we lower it so low that the specific version is not relevant but figured i'd ask

ashleygwilliams (approved):

CHANGELOG.md
Gankra: we do not

PR #697, opened 2024-01-08
Fix Linux Homebrew installation for binaries.

dsully:
In my last PR I missed adding the bin install section that is OS/platform specific.

### DIFF
--- a/cargo-dist/templates/installer/homebrew.rb.j2
+++ b/cargo-dist/templates/installer/homebrew.rb.j2
@@ -68,15 +68,29 @@ class {{ formula_class }} < Formula
 
   def install
     {#- Like the URL case above, write out a single install line in the case that the binary artifacts are the same across architectures #}
-    {%- if arm64_macos.binaries == x86_64_macos.binaries %}
-    bin.install {% for binary in arm64_macos.binaries %}"{{ binary }}"{{ ", " if not loop.last else "" }}{% endfor %}
-    {%- else %}
-    if Hardware::CPU.type == :arm
+    on_macos do
+      {%- if arm64_macos.binaries == x86_64_macos.binaries %}
       bin.install {% for binary in arm64_macos.binaries %}"{{ binary }}"{{ ", " if not loop.last else "" }}{% endfor %}
-    else
-      bin.install {% for binary in x86_64_macos.binaries %}"{{ binary }}"{{ ", " if not loop.last else "" }}{% endfor %}
+      {%- else %}
+      if Hardware::CPU.type == :arm
+        bin.install {% for binary in arm64_macos.binaries %}"{{ binary }}"{{ ", " if not loop.last else "" }}{% endfor %}
+      else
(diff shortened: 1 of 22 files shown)


cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -68,15 +68,29 @@ class {{ formula_class }} < Formula
 
   def install
     {#- Like the URL case above, write out a single install line in the case that the binary artifacts are the same across architectures #}
-    {%- if arm64_macos.binaries == x86_64_macos.binaries %}
-    bin.install {% for binary in arm64_macos.binaries %}"{{ binary }}"{{ ", " if not loop.last else "" }}{% endfor %}
-    {%- else %}
-    if Hardware::CPU.type == :arm
+    on_macos do
```
mistydemeo, line 71: If we're going to introduce per-platform blocks, I think we need to guard against adding the block at all if the platform is missing. Eg if there are only Linux binaries, we shouldn't have an `on_macos` block, and vice versa.

mistydemeo (changes requested):
Good catch. I do think this is starting to get complex enough I wonder if we want to start pulling some of this logic out of the template, but this is fine for now. There is one change I think should make it in this PR, though.

mistydemeo:
Looking at the tests, there's a couple thing we want to take care of:

1. It's possible to generate an invalid `bin.install`:

```ruby
on_linux do
  if Hardware::CPU.type == :arm
    bin.install 
  else
    bin.install "axolotlsay"
  end
end
```

I believe that's a preexisting bug, so the fix can happen in another PR.

2. The test fixtures need to be regenerated via `cargo insta review`.

cargo-dist/templates/installer/homebrew.rb.j2
mistydemeo: It's not actually about different names; rather, it's possible for a project with *multiple* binaries to only build some of them on certain platforms. It's a niche usecase, however.

dsully: Thanks. It certainly makes this much harder. Do you have a preferred way to approach fixing this?

In my testing, having an empty stanza (say for ARM64 Linux when on x86_64 Linux) was not an issue.

dsully:
Ok, I simplified the template, which will eliminate the empty binary case. It's a bit more verbose in the rendered formula, but I think that's ok.

dsully:
Ping

ashleygwilliams:
hey @dsully apologies for the delay, we've been heads down on a few other things- we will get this in and out next week!

dsully:
> hey @dsully apologies for the delay, we've been heads down on a few other things- we will get this in and out next week!

Thanks!

mistydemeo (approved):

Gankra:
@dsully published in https://github.com/axodotdev/cargo-dist/releases/tag/v0.8.1

Issue #650, opened 2023-12-14
Feature idea: updater


mistydemeo:
Although we provide several forms of installers, we don't currently have any kind of updaters aside from the Homebrew package manager output. We should look into providing an updater of some kind for users who've installed a tool via the shell installers.

A few different ways we could look at exposing this:

* Have the installer install an updater script alongside the program itself, and the user can call that to update;
* Develop a crate that provides update-query and update installation mechanics, and allow users of cargo-dist to embed it in their crates for dist-enabled builds

Since we have one package manager-based install already, and will likely have more in the future, we need to make sure that these *don't* get updated by our updater. The package manager needs to be able to handle updating these. This could involve installing a metadata file that the updater uses to tell the *method* the package was installed by, and therefore whether it should try to update it or not. Another possibility would be to avoid installing any updater-related files in package manager-based installs.

mistydemeo:
This shipped!

PR #705, opened 2024-01-12
feat(installers): add install receipt

mistydemeo:
This adds install receipts, letting us track information about where a program was installed from and what its current version is. We'll use that information eventually in the updater, #650; getting the data in place now will help us be prepared for it in the future.

This is still a draft since it's got a few issues I want to fix up on Monday:

- [x] The list of binaries is wrong; I'm using the wrong source.
- [x] The install prefix isn't filled out yet, and probably has to be filled out in shell instead of in cargo-dist.
- [x] Powershell isn't done yet.

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -47,6 +47,11 @@ param (
 $app_name = '{{ app_name }}'
 $app_version = '{{ app_version }}'
 
+$receipt = @"
+{{ receipt | tojson }}
+"@
+$receipt_home = "${env:LOCALAPPDATA}\{{ app_name }}"
+
 function Install-Binary($install_args) {
   if ($Help) {
     Get-Help $PSCommandPath -Detailed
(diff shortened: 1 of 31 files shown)


mistydemeo:
Fully ready now. Powershell isn't tested yet though.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -22,6 +22,11 @@ ARTIFACT_DOWNLOAD_URL="${INSTALLER_DOWNLOAD_URL:-{{ base_url }}}"
 PRINT_VERBOSE=${INSTALLER_PRINT_VERBOSE:-0}
 PRINT_QUIET=${INSTALLER_PRINT_QUIET:-0}
 NO_MODIFY_PATH=${INSTALLER_NO_MODIFY_PATH:-0}
+read -r RECEIPT <<EORECEIPT
+{{ receipt | tojson }}
+EORECEIPT
+# Are we happy with this same path on Linux and Mac?
+RECEIPT_HOME="${HOME}/.config/{{ app_name }}"
```
Gankra, line 29: this should probably be scoped to each platform in The Grand Switch, since in principle each platform can have its own binaries (and presumably good to know `.exe` vs not?).

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -47,6 +47,11 @@ param (
 $app_name = '{{ app_name }}'
 $app_version = '{{ app_version }}'
 
+$receipt = @"
+{{ receipt | tojson }}
+"@
+$receipt_home = "${env:LOCALAPPDATA}\{{ app_name }}"
+
```
Gankra, line 54: ditto on this wanting to be platform-scoped

Gankra (approved):
the basic approach seems simple, my only issue is that some stuff feels like it wants to be platform-scoped, and that's... Very Messy.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: When you say "platform", do you mean architecture? EXE shouldn't be relevant here in the shell installer.

Gankra: hmm you're right that's true right now, but it feels like a tenuous truce.

i mean per-"target-triple". The Grand Switch on the detected target-triple already encodes a list of binaries that is Conceptually free to vary per-platform.

mistydemeo: Now that I realize you're talking about binaries - yeah, this is a bit of a problem should those ever vary per-platform.

I've pushed a commit that puts a placeholder bin array into the template, then updates it with what's actually calculated later in the file. This feels... messy though? Hm.

mistydemeo:
I think I've fixed Powershell. Mind giving it another test, and seeing if the resulting receipt looks correct? In local testing it's producing valid UTF-8 JSON, but worth checking if that's true for you too.

Gankra (approved):
nice!!!

PR #711, opened 2024-01-12
fix: ignore ldd status

Gankra:


cargo-dist/src/linkage.rs
```diff
@@ -393,8 +393,13 @@ fn do_otool(path: &Utf8PathBuf) -> DistResult<Vec<String>> {
 fn do_ldd(path: &Utf8PathBuf) -> DistResult<Vec<String>> {
     let mut libraries = vec![];
 
+    // We ignore the status here because for whatever reason arm64 glibc ldd can decide
+    // to return non-zero status on binaries with no dynamic linkage (e.g. musl-static).
+    // This was observed both in arm64 ubuntu and asahi (both glibc ldd).
+    // x64 glibc ldd is perfectly fine with this and returns 0, so... *shrug* compilers!
     let output = Cmd::new("ldd", "get linkage info from ldd")
         .arg(path)
+        .check(false)
```
ashleygwilliams, line 402: i will ask just for completions sake- can this command return an error we would ever want to be a hard fail? thinking it through it briefly on my own, i think no- but feel like we should have the conversation just to be sure.

ashleygwilliams (approved):

cargo-dist/src/linkage.rs
Gankra: so i'm actually restoring old behaviour (we were ignoring these statuses until i landed axoprocess, because i noticed we were and misty thought it wasn't intentional). we could use non-zero as the same thing we do for "dynamically linked", but at this point i don't trust this exit code, so i'm inclined for us to just always try to parse the output.

Gankra:
For monday: it's possible we can indeed check the status, but rather than error just act like we got "not a dynamic executable" and quietly bail (also, aside, i'm pretty confident the check for "not a dynamic executable" might be useless rn given ldd adds some extra padding to the line).

Gankra:
Various notes from folks investigating this on discord, thanks everyone!!!

-----

So basically on x64 the impl of ld (which ldd shells out to) kinda says "yes i can handle this" and indeed prints "statically linked"

On arm64 (aarch64) the impl of ld says "no way" and we hit the "standard" ldd error path, which sets exit status to 1:

```
    case $ret in
    1)
      # This can be a non-ELF binary or no binary at all.
      nonelf "$file" || {
    echo $" not a dynamic executable" >&2
    result=1
      }
      ;;
 ```

-----

More precise notes from re:fi.64 on discord

> ah so
>
> on arm64, `/lib/ld-linux-aarch64.so.1 --verify path/to/static/bin` returns exit code 1
> on x64, `/lib64/ld-linux-x86-64.so.2 --verify path/to/static/bin` returns exit code 2 
>
> exit code 2 is interpreted by ldd as telling it that it should try to run the dynamic linker on the binary anyway
>
> lol so indeed if I manually run `LD_TRACE_LOADED_OBJECTS=1 /lib64/ld-linux-x86-64.so.2 path/to/static/bin` it says statically linked
>
> if I do that with `ld-linux-aarch64` it...segfaults
>
> the reason it was segfaulting for me is that
> [arm64 static] binaries don't have a PT_DYNAMIC section
> glibc assumes that static binaries do have PT_DYNAMIC
> but the x64 static binaries do

----

Notes from Sierra and I on discord testing out a few platforms:

> ok so `ldd some/static/musl/bin`
> 
> on x64 ubuntu (`ldd (Ubuntu GLIBC 2.31-0ubuntu9.14) 2.31)`),  we get 
> 
> ```
> statically linked
> 
> echo $?
> 0
> ```
> 
> but on aarch64 asahi (`ldd (GNU libc) 2.35`), we get
> 
> ```
> not a dynamic executable
> 
> echo $?
> 1
> ```

File: cargo-dist/src/linkage.rs [hunk 2]
```diff
@@ -393,8 +393,13 @@ fn do_otool(path: &Utf8PathBuf) -> DistResult<Vec<String>> {
 fn do_ldd(path: &Utf8PathBuf) -> DistResult<Vec<String>> {
     let mut libraries = vec![];
 
+    // We ignore the status here because for whatever reason arm64 glibc ldd can decide
```
workingjubilee, line 396: strictly speaking, this should name ld as the primary culprit, because afaik a virtually-identical ldd script is run on both platforms (with the primary difference being in the default linker name selection... mine names some x86-specific options to try if `/usr/lib/ld-linux.so.2` doesn't work out), but it's the ld that ldd is wrapping that is going to cause it to provide amusingly different results.

ashleygwilliams: i am sad it caused a bug in our code but this stuff is so interesting to me. thanks for the additional detail!!

mistydemeo (approved):

PR #714, opened 2024-01-15
chore: update own cargo-dist

mistydemeo:
Should be safe now that #711 is in. I'll test in this PR before merging.


mistydemeo:
Confirmed this fixes our own builds.

Gankra (approved):

PR #722, opened 2024-01-17
fix: elevate priviledges for custom publish jobs

Gankra:


Gankra:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings

* `id-token: write` is basically "get an oauth token for authenticating with package managers like pypi"
* `packages: write` is basically "publish to github packages" (for docker images)

Both of these are reasonable to provide to all publish jobs, and it's not really a security risk to do so, imo.

PR #738, opened 2024-01-23
chore: update axoasset to get zstd fix

Gankra:


ashleygwilliams (approved):
oh dang, no interface update?

Gankra:
yeah we entirely promised zstd, and thought we were doing zstd, and then, just, didn't do that in the impl.

PR #739, opened 2024-01-23
fix: more literally disable host --check

Gankra:
We were using 'is from a fork' as a proxy for 'should have the secret' but evidently dependabot can make branches on your repo and trigger PR CI that doesn't have secrets. This regresses our ability to do validation that you've set things up right but it's more reliable.

### DIFF
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -113,10 +113,11 @@ jobs:
       - id: plan
         run: |
           cargo dist
-          {{%- if dispatch_releases %}} ${{ inputs.tag && (inputs.tag != 'dry-run' && format('host --steps=create --tag={0}', inputs.tag) || format('host --steps=create', inputs.tag))
-          {{%- else %}} ${{ !github.event.pull_request && format('host --steps=create --tag={0}', github.ref_name)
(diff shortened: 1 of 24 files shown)


cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -113,10 +113,11 @@ jobs:
       - id: plan
         run: |
           cargo dist
-          {{%- if dispatch_releases %}} ${{ inputs.tag && (inputs.tag != 'dry-run' && format('host --steps=create --tag={0}', inputs.tag) || format('host --steps=create', inputs.tag))
-          {{%- else %}} ${{ !github.event.pull_request && format('host --steps=create --tag={0}', github.ref_name)
+          {{%- if dispatch_releases %}} ${{ (inputs.tag && inputs.tag != 'dry-run' && format('host --steps=create --tag={0}', inputs.tag))
```
mistydemeo, line 116: I feel like the conditionals here are getting complex enough it might be nice to comment on it; if we run into bugs here it'll help us remember our reasoning a little better.

mistydemeo (approved):

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: that's a good call, i'll add some comments in a follow up (rn just wanna get the fix out)

PR #741, opened 2024-01-24
feat(lies): add --artifacts=lies

Gankra:
`--artifacts=lies` is equivalent to `--artifacts=all` except instead of being "something doomed to fail with `cargo dist build`" it just lies through its teeth and fakes the builds to generate all the files needed to upload.

This would let us have a staging upload workflow you can run on anyone's machine:

```
STAGE_INTO_THE_ABYSS=1
AXO_RELEASES_STAGING_TOKEN="...."
cargo dist host --steps=create
cargo dist build --artifacts=lies
cargo dist host --steps=upload --steps=release --steps=announce
```

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -16,7 +16,7 @@ use axoasset::LocalAsset;
 use axoprocess::Cmd;
 use backend::{
     ci::CiInfo,
-    installer::{self, InstallerImpl},
+    installer::{self, msi::MsiInstallerInfo, InstallerImpl},
 };
 use camino::Utf8Path;
 use cargo_build::{build_cargo_target, rustup_toolchain};
@@ -75,7 +75,11 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
 
     // Run all the local build steps first
     for step in &dist.local_build_steps {
-        run_build_step(&dist, step, &manifest)?;
+        if dist.local_builds_are_lies {
+            build_fake(&dist, step, &manifest)?;
+        } else {
+            run_build_step(&dist, step, &manifest)?;
+        }
     }
 
     // Compute linkage data now that we're done all builds
@@ -83,7 +87,11 @@ pub fn do_build(cfg: &Config) -> Result<DistManifest> {
 
     // Next the global steps
     for step in &dist.global_build_steps {
-        run_build_step(&dist, step, &manifest)?;
+        if dist.local_builds_are_lies {
+            build_fake(&dist, step, &manifest)?;
+        } else {
+            run_build_step(&dist, step, &manifest)?;
+        }
     }
 
     Ok(manifest)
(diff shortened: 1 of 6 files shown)


mistydemeo:
Pushed an implementation. The bulk of the logic is held up by `build_fake`, a variation of `run_build_step` which dispatches to stub builds for a bunch of the implementations. These stub builds check for the files cargo-dist looking for and basically just does `touch` calls, skipping any and all builds. The skipped steps are:

* Generic builds
* Cargo builds
* Extra artifacts
* Source tarballs
* Rustup

Other global builds, such as copying files and generating tarballs, *are* run currently. I think that's fine, and getting real tarballs with zero-byte stub binaries is probably better for testing.

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,88 @@ fn run_build_step(
...
+        }) => copy_file_or_dir(src_path, dest_path),
+        // The remainder of these are mostly safe to run as fake steps
+        BuildStep::Zip(ZipDirStep {
+            src_path,
+            dest_path,
+            zip_style,
+            with_root,
+        }) => zip_dir(src_path, dest_path, zip_style, with_root.as_deref()),
+        BuildStep::GenerateInstaller(installer) => {
+            generate_installer(dist_graph, installer, manifest)
```
Gankra, line 184: so this one is tricky because the msi installers *can't* be built on all platforms. probably means just expanding out and applying this pattern on the match inside this function

cargo-dist/src/lib.rs
```diff
@@ -214,6 +304,17 @@ fn generate_source_tarball(
     Ok(())
 }
 
+fn generate_fake_source_tarball(
+    _graph: &DistGraph,
+    _committish: &str,
+    _prefix: &str,
+    target: &Utf8Path,
+) -> DistResult<()> {
+    std::fs::File::create(target)?;
```
Gankra, line 313: is there any reason this should be faked? i thought source tarballs were trivial to generate on any platform.

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,88 @@ fn run_build_step(
...
+    if !dist.target_dir.exists() {
+        std::fs::create_dir(&dist.target_dir).into_diagnostic()?;
+    }
+
+    for idx in binaries {
+        let binary = dist.binary(*idx);
+
+        for exe_dest in &binary.copy_exe_to {
+            std::fs::File::create(exe_dest).into_diagnostic()?;
+        }
```
Gankra, line 231: we never actually populate it today but there's `copy_symbols_to` or something as well, it would be nice to mock that as well for future proofing on us re-enabling it.

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,88 @@ fn run_build_step(
...
+}
+
+fn build_fake_generic_artifacts(dist: &DistGraph, target: &GenericBuildStep) -> Result<()> {
+    generate_fake_artifacts(dist, &target.expected_binaries)
+}
+
+fn run_fake_extra_artifacts_build(dist: &DistGraph, target: &ExtraBuildStep) -> Result<()> {
+    for artifact in &target.expected_artifacts {
+        let path = dist.dist_dir.join(artifact);
+        std::fs::File::create(&path).into_diagnostic()?;
```
Gankra, line 215: I'd prefer we use axoasset for these still, if we can.

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,88 @@ fn run_build_step(
...
+    }
+
+    Ok(())
+}
+
+fn generate_fake_artifacts(dist: &DistGraph, binaries: &[BinaryIdx]) -> Result<()> {
+    if !dist.target_dir.exists() {
+        std::fs::create_dir(&dist.target_dir).into_diagnostic()?;
```
Gankra, line 223: ditto axoasset

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,88 @@ fn run_build_step(
...
+    }
+
+    for idx in binaries {
+        let binary = dist.binary(*idx);
+
+        for exe_dest in &binary.copy_exe_to {
+            std::fs::File::create(exe_dest).into_diagnostic()?;
```
Gankra, line 230: ditto axoasset

cargo-dist/src/lib.rs
mistydemeo: Aha, I didn't consider MSI - I was only testing wish shell, powershell, Homebrew.

cargo-dist/src/lib.rs
mistydemeo: Source tarballs require git to be installed and for you to be running the process inside a git checkout. Faking this makes it possible for us to simulate it in a few extra circumstances.

cargo-dist/src/lib.rs
mistydemeo: Oh huh, I forgot about that one. OK, will add.

cargo-dist/src/lib.rs
mistydemeo: I'll look into it.

cargo-dist/src/lib.rs
Gankra: this is kinda overkill but i'll defer to your judgement :)

cargo-dist/src/lib.rs
mistydemeo: OK, special-cased MSI while generating real installers for the other paths.

cargo-dist/src/lib.rs
mistydemeo: For context, this fixes a test for nix - I'd planned to write a fix for this in other contexts, and it turned out this let me do it as part of this PR.

cargo-dist/src/lib.rs
mistydemeo: Swapped to axoasset.

cargo-dist/src/lib.rs
```diff
@@ -144,6 +152,100 @@ fn run_build_step(
...
+            checksum,
+            src_path,
+            dest_path,
+        }) => Ok(generate_and_write_checksum(checksum, src_path, dest_path)?),
+        // Except source tarballs, which are definitely not okay
+        BuildStep::GenerateSourceTarball(SourceTarballStep {
+            committish,
+            prefix,
+            target,
+        }) => Ok(generate_fake_source_tarball(
```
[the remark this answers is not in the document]
mistydemeo: Added.

mistydemeo (approved):
Self-approving based on Gankra's approval.

PR #744, opened 2024-01-28
chore: be more permissive in vec args

SaraVieira:
Makes most vecs args accept also --arg=A,B

### DIFF
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -286,6 +286,7 @@ impl GenerateMode {
 #[derive(Args, Clone, Debug)]
 pub struct GenerateArgs {
     /// Which type of configuration to generate
+    #[clap(long, value_delimiter(','))]
     pub mode: Vec<GenerateMode>,
 
     /// Check if the generated output differs from on-disk config without writing it
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -54,7 +54,7 @@ pub struct Cli {
     ///
     /// If left unspecified we will use the values in [workspace.metadata.dist],
     /// except for `cargo dist init` which will select some "good defaults" for you.
-    #[clap(long, short)]
+    #[clap(long, short, value_delimiter(','))]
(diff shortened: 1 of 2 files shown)


cargo-dist/src/cli.rs
```diff
@@ -54,23 +54,23 @@ pub struct Cli {
...
     pub installer: Vec<InstallerStyle>,
 
     /// CI we want to support
     ///
     /// If left unspecified we will use the value in [workspace.metadata.dist].
     /// `cargo dist init` will persist the values you pass to that location.
-    #[clap(long, short)]
+    #[clap(long, short, value_delimiter(','))]
     #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
     pub ci: Vec<CiStyle>,
```
Gankra, line 75: unfortunately not entirely sure if we can get away with these, given `generate` supports global args [and has positional args](https://opensource.axo.dev/cargo-dist/book/reference/cli.html#cargo-dist-generate)... but maybe `generate` shouldn't take global args, now that I think about it? cc @mistydemeo 

context for misty: value_delimiter unlocks two new arg passing modes, in addition the one we get by default with Vec:

```
--my-vals=a --my-vals=b
--my-vals=a,b
--my-vals a b
```

The last one is problematic if your Command takes any positional args (is "b" part of `--my-vals` or a positional arg?). This is why that arg-passing mode is disabled by default in modern clap.

We largely don't use positional args anymore, except in `generate`, but generate doesn't really have a reason to take the global args anyway (filtering it down to a subset of platforms would almost surely be an error). (Alternatively it's possible it's "fine" anyway because maybe clap can disambiguate based on a GenerateStyle never parsing as any of these..?)

Gankra (approved):
tentatively approved, want to check in with misty on a detail

cargo-dist/src/cli.rs
mistydemeo: `generate`'s args are largely not used by other people, just us, right? I think we can adjust its behaviour without impacting our users.

PR #755, opened 2024-01-31
feat(github): migrate to artifacts-v4

### LINKED ISSUE #653: Update for v4 of the upload- & download-artifact actions
The GitHub Actions workflow generated by `cargo-dist` currently uses v3 of [`actions/upload-artifact`](https://github.com/actions/upload-artifact) and [`actions/download-artifact`](https://github.com/actions/download-artifact).  v4 of these workflows was released on December 14 (["What's New" documentation](https://github.com/actions/toolkit/tree/main/packages/artifact#v2---whats-new)), so you'll probably want to update the workflow at some point.

mistydemeo:
Migrates us from v3 to v4, with a refactor that has us using multiple artifact buckets instead of a single staging bucket.

This *doesn't* migrate the signing job yet. This one is awkward since the old behaviour actually did rely on overwriting a specific artifact. Do we have only one Windows target triple at the moment, or is there more than one that's possible? If there's only one, I can hardcode what we fetch and write to.

Sample build: https://github.com/mistydemeo/cargodisttest/actions/runs/7729932089

Sample release: https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.92

Fixes #653

### DIFF
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -117,15 +117,15 @@ jobs:
           {{%- else %}} ${{ (!github.event.pull_request && format('host --steps=create --tag={0}', github.ref_name))
           {{%- endif %}}
           {{%- if "axodotdev" in hosting_providers %}} || (env.AXO_RELEASES_TOKEN && 'host --steps=check') {{%- endif %}}
-          {{{- " || 'plan' }} --output-format=json > dist-manifest.json" | safe }}}
+          {{{- " || 'plan' }} --output-format=json > plan-dist-manifest.json" | safe }}}
           echo "cargo dist ran successfully"
-          cat dist-manifest.json
-          echo "manifest=$(jq -c "." dist-manifest.json)" >> "$GITHUB_OUTPUT"
+          cat plan-dist-manifest.json
+          echo "manifest=$(jq -c "." plan-dist-manifest.json)" >> "$GITHUB_OUTPUT"
       - name: "Upload dist-manifest.json"
         uses: actions/upload-artifact@v4
         with:
-          name: artifacts-dist-manifest
-          path: dist-manifest.json
+          name: artifacts-plan-dist-manifest
+          path: plan-dist-manifest.json
 
 {{%- for job in plan_jobs %}}
 
@@ -435,7 +435,6 @@ jobs:
           # Overwrite the previous copy
           name: artifacts-dist-manifest
           path: dist-manifest.json
-          overwrite: true
 
 
 {{%- for job in host_jobs %}}
--- a/cargo-dist/templates/ci/github_ci.yml.j2
+++ b/cargo-dist/templates/ci/github_ci.yml.j2
@@ -122,9 +122,9 @@ jobs:
           cat dist-manifest.json
           echo "manifest=$(jq -c "." dist-manifest.json)" >> "$GITHUB_OUTPUT"
       - name: "Upload dist-manifest.json"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: artifacts
+          name: artifacts-dist-manifest
           path: dist-manifest.json
 
 {{%- for job in plan_jobs %}}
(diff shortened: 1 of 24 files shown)


cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -424,10 +427,12 @@ jobs:
...
           echo "manifest=$(jq -c "." dist-manifest.json)" >> "$GITHUB_OUTPUT"
       - name: "Upload dist-manifest.json"
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: artifacts
+          # Overwrite the previous copy
+          name: artifacts-dist-manifest
           path: dist-manifest.json
+          overwrite: true
```
Gankra, line 435: wait what the fuck, there's an overwrite flag? god damnit github why was your announcement so scary when the API is way more usable. but anyway, i think it's probably safer for us to actually make plan and host use different artifacts/names. there's no situation where we want "plan OR host". either you definitely want plan, or definitely want host, yeah..? (happy to be corrected here)

cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -314,7 +316,7 @@ jobs:
     steps:
       # Get all the artifacts for the signing tasks to use
       - name: Fetch local artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: artifacts
           path: target/distrib/
```
Gankra, line 322: missing merge-multiple/pattern (let's just be safe here).

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: tbc on my proposal: give host the glorious "dist-manifest.json" name, and downgrade plan to "plan-dist-manifest.json"

Gankra (approved):

Gankra:
> This doesn't migrate the signing job yet. This one is awkward since the old behaviour actually did rely on overwriting a specific artifact. Do we have only one Windows target triple at the moment, or is there more than one that's possible? If there's only one, I can hardcode what we fetch and write to.

The windows signing stuff is totally bitrotten at this point, but it looks like we could use `overwrite: true` in a pinch. We'll re-eval that when someone needs it.

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: I feel like the issue is, if they have two different names, we might end up with the wrong one landing in the final release after we merge artifacts later?

I think the thing with `overwrite` is that it doesn't mutate the artifact: it replaces the old artifact with a *new* artifact with the same name. So we can't just treat it like we treated the old mutable artifacts.

mistydemeo:
> The windows signing stuff is totally bitrotten at this point, but it looks like we could use overwrite: true in a pinch. We'll re-eval that when someone needs it.

Right, yeah, the issue is knowing the name of the Windows artifacts job. Is that predictable? Do we only have a single Windows target triple? If so, then yes, I can predict the artifact name and it'll be fine.

cargo-dist/templates/ci/github_ci.yml.j2
Gankra: that's why i'm suggesting the plan one be renamed -- so it *can't* end up being used (github releases rm's all prefixed dist-manifests, and uploads the one non-prefixed one).

mistydemeo: I missed at first you were talking about prefixing the *manifest*. That's a good idea; pushed a commit which does that, and it's working happily. 

https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.95

https://github.com/mistydemeo/cargodisttest/actions/runs/7733291447

cargo-dist/templates/ci/github_ci.yml.j2
mistydemeo: Yeah, I left this out because this is the Windows signing job I was asking about

mistydemeo:
Pushed a commit to handle the Windows signing issue.

PR #767, opened 2024-02-02
feat(github): swap back to macos-12 for arm64

mistydemeo:
For consistency with Intel and older OS support, this may be a better choice for the time being. We'll reconsider this in the future. In the meantime, provide some more information in the docs about how to opt into macos-14.


Gankra (approved):
rip to nice things

PR #776, opened 2024-02-06
chore: add tests for install-receipts from sh

### LINKED ISSUE #773: sed error in install script
I'm using `cargo-dist` in a project, and was testing out the install script and saw an error message.

```
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/ethankhall/scope/releases/download/v2024.2.4/dev-scope-installer.sh | sh
sed: 1: "s,"CARGO_DIST_BINS","sc ...": bad flag in substitute command: '"'
```

The install looks to have worked, but I would expect to not see this error message.

## System Details

```
❯ echo $SHELL
/bin/zsh
❯ zsh --version
zsh 5.9 (x86_64-apple-darwin23.0)
❯ sh --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)
Copyright (C) 2007 Free Software Foundation, Inc.
```

Gankra:
fixes #773

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -191,7 +191,7 @@ function Invoke-Installer($bin_paths) {
   $dest_dir = Join-Path $root "bin"
 
   # The replace call here ensures proper escaping is inlined into the receipt
-  $receipt = $receipt.Replace('AXO_CARGO_HOME', $root.replace("\", "\\"))
+  $receipt = $receipt.Replace('AXO_CARGO_HOME', $dest_dir.replace("\", "\\"))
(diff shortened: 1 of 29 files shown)


Gankra:
TODO: 

* [ ] check if we really want `~/.cargo/` and not `~/.cargo/bin/` in the install receipt
* [ ] pick a "good" seperator for the sed (currently stubbed to `~`) ((or better escaping?))

mistydemeo:
>  check if we really want ~/.cargo/ and not ~/.cargo/bin/ in the install receipt

We do, yes. While we're likely to only install binaries, in the future we'll be installing other kinds of files - libraries, config files, etc. These are installed in other subdirectories inside the install prefix.

PR #779, opened 2024-02-06
feat(tag-namespace): add tag-namespace config

### LINKED ISSUE #758: allow user to configure tag "namespace"
some users deploy from a workspace and namespace tags to coordinate different release trains for different parts of the workspace. this usually means they want a member-v0.0.0 style tag instead of the v0.0.0 tag that cargo-dist expects.

i propose the following config to generate the following workflow:

```toml
# Cargo.toml
tag-namespace = "customstring"
```

generates
```yml
# release.yml
on:
  push:
    tags:
      - 'customstring-**[0-9]+.[0-9]+.[0-9]+*'
  pull_request:
```

Gankra:
fixes #758


ashleygwilliams (approved):
looks good but would of course like to see docs. does this also namespace the workflow file name?

PR #790, opened 2024-02-13
feat: produce updater on request

mistydemeo:
This is the first part of installing the updater for those who plan to use it. In this phase, a new setting allows users to opt into creating the installer and adding it to the artifacts list; if requested, it'll be added to the GitHub artifacts along with the software itself, with one copy per target triple.

In the next phase, the shell and PowerShell installers will be updated to check for and install the updater if present, renaming it to the appropriate `appname-updater` filename.

Note that right now this *only* produces the updater via `cargo install`ing from `HEAD` in the repo, but once we have the first stable release there will be prebuilt binaries in the axoupdater repo; in that case, we can simply pull down the latest binary for the target triple and only fall back to building from source if there's no binary for the requested triple.

### DIFF
--- a/cargo-dist/Cargo.toml
+++ b/cargo-dist/Cargo.toml
@@ -61,6 +61,7 @@ mach_object = "0.1"
 goblin = "0.8.0"
 similar = "2.4.0"
 tokio = { version = "1.36.0", features = ["full"] }
+temp-dir = "0.1.12"
 
 [dev-dependencies]
 insta = { version = "1.35.1", features = ["filters"] }
--- a/cargo-dist/Cargo.toml
+++ b/cargo-dist/Cargo.toml
@@ -36,10 +36,11 @@ axocli = { version = "0.2.0", optional = true }
 axotag = "0.1.0"
 cargo-dist-schema = { version = "=0.11.1", path = "../cargo-dist-schema" }
 
-axoasset = { version = "0.7.0", features = ["json-serde", "toml-serde", "toml-edit", "compression"] }
+axoasset = { version = "0.8.0", features = ["json-serde", "toml-serde", "toml-edit", "compression"] }
 axoprocess = { version = "0.2.0" }
 axoproject = { version = "0.7.0", default-features = false, features = ["cargo-projects", "generic-projects"] }
 gazenot = { version = "0.3.0" }
+octocrab = "0.34.2"
 
 comfy-table = "7.0.1"
 miette = { version = "7.0.0" }
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -333,6 +333,10 @@ pub struct DistMetadata {
     /// cargo-dist can co-exist with other release workflows in complex workspaces
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tag_namespace: Option<String>,
+
+    /// Whether to install an updater program alongside the software
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub install_updater: Option<bool>,
 }
 
 impl DistMetadata {
@@ -380,6 +384,7 @@ impl DistMetadata {
             extra_artifacts: _,
             github_custom_runners: _,
             tag_namespace: _,
+            install_updater: _,
         } = self;
         if let Some(include) = include {
             for include in include {
@@ -436,6 +441,7 @@ impl DistMetadata {
             extra_artifacts,
             github_custom_runners,
             tag_namespace,
+            install_updater,
         } = self;
 
         // Check for global settings on local packages
@@ -506,6 +512,9 @@ impl DistMetadata {
         }
         if tag_namespace.is_some() {
             warn!("package.metadata.dist.tag-namespace is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
+        }
+        if install_updater.is_some() {
+            warn!("package.metadata.dist.install-updater is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
         }
 
         // Merge non-global settings
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -513,9 +513,6 @@ impl DistMetadata {
         if tag_namespace.is_some() {
             warn!("package.metadata.dist.tag-namespace is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
         }
-        if install_updater.is_some() {
-            warn!("package.metadata.dist.install-updater is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
-        }
 
         // Merge non-global settings
         if installers.is_none() {
@@ -568,6 +565,9 @@ impl DistMetadata {
         }
         if github_custom_runners.is_none() {
             *github_custom_runners = workspace_config.github_custom_runners.clone();
+        }
+        if install_updater.is_none() {
+            *install_updater = workspace_config.install_updater;
         }
 
         // This was historically implemented as extend, but I'm not convinced the
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -167,11 +167,9 @@ pub fn fetch_updater(dist_graph: &DistGraph, updater: &UpdaterStep) -> Result<()
 
     // Install to a temporary path before moving it to the destination
     cmd.arg("install")
+        .arg("axoupdater-cli")
         .arg("--root")
         .arg(tmpdir.path())
-        .arg("--git")
-        .arg("https://github.com/axodotdev/axoupdater")
-        .arg("--all-features")
         .arg("--bin")
         .arg("axoupdater");
 
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -27,6 +27,7 @@ use config::{
 use console::Term;
 use generic_build::{build_generic_target, run_extra_artifacts_build};
 use semver::Version;
+use temp_dir::TempDir;
 use tracing::info;
 
 use errors::*;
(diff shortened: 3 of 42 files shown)


cargo-dist/src/config.rs
```diff
@@ -503,6 +509,9 @@ impl DistMetadata {
         if tag_namespace.is_some() {
             warn!("package.metadata.dist.tag-namespace is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
         }
+        if install_updater.is_some() {
+            warn!("package.metadata.dist.install-updater is set, but this is only accepted in workspace.metadata (value is being ignored): {}", package_manifest_path);
+        }
```
Gankra, line 514: *thinks* this seems like it should be an app-scoped config, yeah? since each app gets its own installer-sh (can ship it as top-level, but, it *can* be app-scoped, at least).

cargo-dist/src/tasks.rs
```diff
@@ -1393,6 +1415,30 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
     }
 
+    fn add_updater(&mut self, release_idx: ReleaseIdx, variant_idx: ReleaseVariantIdx) {
+        if !self.local_artifacts_enabled() {
+            return;
+        }
+
+        let release = self.release(release_idx);
+        let variant = self.variant(variant_idx);
+        let filename = format!("{}-{}-update", release.app_name, variant.target);
```
Gankra, line 1425: can't this just be `{variant.id}-{update}`?

Gankra: also is there a reason we need to generate N copies of this? We could just make one universal/shared one and the installer could just rename it on fetch..?

(tbh i expected the design here to be putting it inside the archives... hmm... interesting)

cargo-dist/src/tasks.rs
```diff
@@ -1654,6 +1701,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             }
 
             artifacts.push(fragment);
+
+            // Create the "pretend" updaters similar to the above for exezips
+            if self.inner.install_updater {
+                let filename = format!("{}-{}-update", release.app_name, variant.target);
```
Gankra, line 1707: once again variant.id is for this

Gankra: actually you just shouldn't be computing this twice. one of these two should set the other.

cargo-dist/src/tasks.rs
```diff
@@ -1985,6 +2045,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
                     .map(|(_, dest_path)| dest_path.file_name().unwrap().to_owned())
                     .collect(),
             });
+
+            // Create the "pretend" updaters similar to the above for exezips
+            if self.inner.install_updater {
+                let filename = format!("{}-{}-update", release.app_name, variant.target);
```
Gankra, line 2051: once again, centralize this

cargo-dist/src/tasks.rs
mistydemeo: Keeping it separate makes it easier to avoid fetching and installing it in the Homebrew and npm installers; that way there's no risk of it leaking into inappropriate places.

> We could just make one universal/shared one and the installer could just rename it on fetch..?

The updater is native code; we need one per target. We choose which version to fetch and install in the installers.

cargo-dist/src/tasks.rs
mistydemeo: Fixed; swapped to something similar to `make_executable_zip_for_variant`.

cargo-dist/src/config.rs
mistydemeo: Oh yes, good call. Changed.

mistydemeo:
Made the changes you suggested and pushed. Tests are passing, and I've tested this with my cargodisttest repo using the shell installer; it's installing the updater, and it runs fine. I think we're about ready to merge this, assuming we're ready to cut an initial release of the updater itself.

mistydemeo:
Tested actually running the update, and it's working fine:

```
$ axolotlsay-update

Checking for updates...
downloading axolotlsay 0.2.106 x86_64-apple-darwin
installing to /Users/mistydemeo/.cargo/bin
  axolotlsay
  axolotlsay-update
everything's installed!
New release installed!
```

cargo-dist/src/lib.rs
```diff
@@ -149,9 +150,44 @@ fn run_build_step(
...
+
+    // Install to a temporary path before moving it to the destination
+    cmd.arg("install")
+        .arg("--root")
+        .arg(tmpdir.path())
+        .arg("--git")
+        .arg("https://github.com/axodotdev/axoupdater")
+        .arg("--all-features")
+        .arg("--bin")
+        .arg("axoupdater");
```
Gankra, line 176: hmm in an ideal world we can fetch the latest release's prebuilt binaries? (with fallback to this for unbuilt platforms)

cargo-dist/src/lib.rs
```diff
@@ -149,9 +150,44 @@ fn run_build_step(
...
+
+    cmd.run()?;
+
+    // OK, now we have a binary in the tempdir
+    let mut source = tmpdir.path().join("bin").join("axoupdater");
+    if updater.target_triple.contains("windows") {
+        source.set_extension("exe");
+    }
```
Gankra, line 184: i'm fairly certain we have this kind of logic somewhere else, and should reuse/factor it.

cargo-dist/src/tasks.rs
```diff
@@ -1399,6 +1421,33 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
 
+    fn add_updater(&mut self, variant_idx: ReleaseVariantIdx) {
+        if !self.local_artifacts_enabled() {
+            return;
+        }
+
+        let artifact = self.make_updater_for_variant(variant_idx);
+
+        self.add_local_artifact(variant_idx, artifact);
+    }
```
Gankra, line 1432: might be worth adding a comment here being like "yep this makes an updater for every app, it could be shared between apps, but for an mvp it's fine".

cargo-dist/src/tasks.rs
```diff
@@ -1703,6 +1753,16 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             }
 
             artifacts.push(fragment);
+
+            // Create the "pretend" updaters similar to the above for exezips
+            if self.inner.install_updater {
+                let artifact = self.make_updater_for_variant(variant_idx);
+                updaters.push(UpdaterFragment {
```
Gankra, line 1760: lolsob (it's fine, just, oh no my crimes)

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -73,6 +73,13 @@ function Install-Binary($install_args) {
   {%- endfor %}
   }
 
+  {%- for updater in updaters %}
+  $platforms["{{ updater.target_triple }}"]["updater"] = @{
+    "artifact_name" = "{{ updater.id }}"
+    "bin" = "{{ updater.binary }}"
+  }
+  {%- endfor %}
```
Gankra, line 81: ...i suppose that is the easiest way. makes me wonder if this should just be in the "artifacts" input above though

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -173,6 +180,16 @@ function Download($download_url, $platforms) {
...
   }
+
+  if ($null -ne $info["updater"]) {
+    $updater_id = $info["updater"]["artifact_name"]
+    $updater_url = "$download_url/$updater_id"
+    $out_name = "$tmp\{{ app_name}}-update.exe"
+
+    $wc.downloadFile($updater_url, $out_name)
+    $bin_paths += $out_name
+  }
```
Gankra, line 191: do we want to add printouts for "unpacked updater"?

Gankra (approved):
Looks reasonable...

cargo-dist/src/lib.rs
mistydemeo: Yes! Right now it's not done just because those binaries don't exist yet. This is the fallback until that happens. Likewise, it'll be built from the crates.io release instead of git.

ashleygwilliams: ignore me if i'm way off here, but i'm curious why we'd prefer the crates.io release over the git one (i think we often can't be sure there is a crates.io release, whereas we always know there is a git one)

mistydemeo: This is of axoupdater, rather than the user's own software. Because we publish axoupdater ourselves, we can know that we'll always have a stable version on crates.io.

cargo-dist/src/lib.rs
mistydemeo: There is another place, but it operates on strings of base filenames: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/tasks.rs#L1150-L1151

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: The list of binaries will already include a thing that ends in `-update`, but I could add an extra message on top of that if you think it'd help legibility.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Yeah, I wondered about that. In practice doing that had turned out to be a bit less convenient than I'd like, so this was easier for now.

cargo-dist/Cargo.toml
```diff
@@ -36,14 +36,11 @@ axocli = { version = "0.2.0", optional = true }
 axotag = "0.1.0"
 cargo-dist-schema = { version = "=0.11.1", path = "../cargo-dist-schema" }
 
-axoasset = { version = "0.7.0", features = ["json-serde", "toml-serde", "toml-edit", "compression"] }
+axoasset = { git = "https://github.com/axodotdev/axoasset", branch = "decompress", features = ["json-serde", "toml-serde", "toml-edit", "compression"] }
```
ashleygwilliams, line 39: i assume we won't merge this until we publish the axo asset version with this feature?

cargo-dist/Cargo.toml
```diff
@@ -36,14 +36,11 @@ axocli = { version = "0.2.0", optional = true }
...
 axoprocess = { version = "0.2.0" }
 axoproject = { version = "0.7.0", default-features = false, features = ["cargo-projects", "generic-projects"] }
 gazenot = { version = "0.3.0" }
 octocrab = "0.34.2"
-tar = "0.4.40"
```
ashleygwilliams: honestly love this

ashleygwilliams (commented):
just looking at the last commit all i can say is i love moving logic into axoasset haha, really cleans things up!

cargo-dist/Cargo.toml
mistydemeo: Made that release and swapped over to it!

PR #791, opened 2024-02-13
feat(formula): add config

### LINKED ISSUE #777: Allow publishing under bin name instead of package name
Hello!

We want to publish our binary on homebrew under the binary name, but it is published under the crate name. Is there a way to do that?

ashleygwilliams:
fixes #777

### DIFF
--- a/book/src/installers/homebrew.md
+++ b/book/src/installers/homebrew.md
@@ -11,6 +11,26 @@ tap = "axodotdev/homebrew-formulae"
 publish-jobs = ["homebrew"]
 ```
 
+Since 0.11.0, cargo-dist can, optionally, also customize your Homebrew formula name.
+By default, your formula will be named using the app name (in Rust, this is the crate
+name). If you are overriding the bin name, you may want to make your Homebrew formula
+match- you can do so with config like this:
+
+```toml
+[package]
+name = "myappname"
+version = "0.666.0"
+default-run = "mybinname"
+
+[[bin]]
+name = "mybinname"
+path = "src/main.rs"
+
+tap = "axodotdev/homebrew-formulae"
+publish-jobs = ["homebrew"]
+formula = "mybinname"
+```
+
 In order to write to a tap GitHub repository, cargo-dist needs a [personal access token](https://github.com/settings/tokens/new?scopes=repo) with the `repo` scope exposed as `HOMEBREW_TAP_TOKEN`. For more information on GitHub Actions secrets, [consult this documentation](https://docs.github.com/en/actions/security-guides/encrypted-secrets).
 
 Limitations/Caveats:
--- a/book/src/reference/config.md
+++ b/book/src/reference/config.md
@@ -345,6 +345,17 @@ For instance for packages that are a library and a CLI binary, some developers p
 If you use this you *probably* want to set it on `[package.metadata.dist]` and
 not `[workspace.metadata.dist]`. See ["inferring precise-builds"](#inferring-precise-builds) for details.
 
+### formula
+
+> since 0.11.0
+
+Example: `formula = "axolotlbrew"`
+
+Specifies a string to override the default Homebrew formula name (the app name). This works
+well specifically for folks who are customizing their bin name and would like the Homebrew
+formula to match the bin name as opposed to the app name (which, in Rust, is the crate name).
+
+You must set this on `[package.metadata.dist]` and not `[workspace.metadata.dist]`.
 
 ### github-custom-runners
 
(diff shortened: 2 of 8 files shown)


cargo-dist/tests/integration-tests.rs
```diff
@@ -62,6 +62,42 @@ path-guid = "BFD25009-65A4-4D1E-97F1-0030465D90D6"
...
+fn axolotlsay_custom_formula() -> Result<(), miette::Report> {
+    let test_name = _function_name!();
+    AXOLOTLSAY.run_test(|ctx| {
+        let dist_version = ctx.tools.cargo_dist.version().unwrap();
+        ctx.patch_cargo_toml(format!(r#"
+[workspace.metadata.dist]
+cargo-dist-version = "{dist_version}"
+installers = ["homebrew"]
+tap = "axodotdev/homebrew-packages"
+formula = "AXOLOTLSHOUT"
```
mistydemeo, line 75: We probably want to test with a lowercase name, not an uppercase name. Capitalization is significant and has some implications.

We could also consider normalizing the name based on other factors in a later update, but I think it also just makes sense to link people to the Homebrew docs about formula names when they're picking one.

mistydemeo (dismissed):

mistydemeo:
I'm seeing a couple things here we want to make sure we fix:

* The snapshots aren't capturing the formula if the name is specified - not sure what's going on yet.
* Running locally, I'm seeing that the wrong class name is generated in the formula; the culprit is this line, which is still keyed off app name: https://github.com/axodotdev/cargo-dist/blob/d19b47f37467cdc4b5040c7dfa59c9e7964782b9/cargo-dist/src/tasks.rs#L1888

mistydemeo:
Realized that the snapshots are missing the formula because the name to add to the snapshot is hardcoded: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/tests/gallery/dist.rs#L214

It's probably easiest to turn this into a glob that grabs all `*.rb` files.

MarinPostma (commented):
Looks like exactly what I need, thanks @ashleygwilliams

ashleygwilliams:
add docs

cargo-dist/tests/gallery/dist.rs
```diff
@@ -209,21 +209,51 @@ impl<'a> TestContext<'a, Tools> {
...
+        assert!(
+            number_found <= 1,
+            "found {} files with the extension {}, expected 1 or 0",
+            number_found,
+            extension
+        );
+        files.first().cloned()
+    }
+
+    fn load_files_with_extension(dirname: Utf8PathBuf, extension: &str) -> Vec<Utf8PathBuf> {
```
Gankra, line 240: I suggest

```suggestion
    fn find_files_with_suffix(dirname: &Utf8Path, suffix: &str) -> Vec<Utf8PathBuf> {
```

(and a bunch of fallout changes from that)

Gankra:
condolences, merged changes need this pr rebased and the snapshot regen'd

Gankra (commented):
test harness changes look good

i was sure runtest_homebrew_installer was gonna need edits but nope, it was already perfectly robust to only assuming it had the path it was given!

book/src/reference/config.md
```diff
@@ -345,6 +345,17 @@ For instance for packages that are a library and a CLI binary, some developers p
...
+
+> since 0.11.0
+
+Example: `formula = "axolotlbrew"`
+
+Specifies a string to override the default Homebrew formula name (the app name). This works
+well specifically for folks who are customizing their bin name and would like the Homebrew
+formula to match the bin name as opposed to the app name (which, in Rust, is the crate name).
+
+You must set this on `[package.metadata.dist]` and not `[workspace.metadata.dist]`.
```
Gankra, line 358: heh, good catch.

Gankra (approved):

book/src/installers/homebrew.md
```diff
@@ -9,6 +9,27 @@ cargo-dist can, optionally, publish your formula to a tap repository for you on
 ```toml
 tap = "axodotdev/homebrew-formulae"
 publish-jobs = ["homebrew"]
+
+Since 0.11.0, cargo-dist can, optionally, also customize your Homebrew formula name.
+By default, your formula will be named using the app name (in Rust, this is the crate
+name). If you are overriding the bin name, you may want to make your Homebrew formula
+match- you can do so with config like this:
+
+```toml
```
Gankra, line 18: uhhh did something go wrong here

ashleygwilliams: yes, apologies, forcepushed fix

Gankra (approved):

PR #793, opened 2024-02-16
wip: platform refactor

Gankra:
This started as the Big Linkage Refactor but ended up being the Big Platform Fallback Refactor. The meat here is platform.rs.

I got way in over my head when I started adding RuntimeCondition and SupportQuality to try to centralize stuff like "the glibc version has to be this".

Optimistically I think the point this refactor was at was "ok all the data is threaded, now I just need to completely rework the logic for installer.sh.js, installer.ps1.j2, installer.rb, and npm, which, was Too Much to do at once.

### DIFF
--- a/cargo-dist/src/backend/installer/mod.rs
+++ b/cargo-dist/src/backend/installer/mod.rs
@@ -6,8 +6,8 @@ use camino::Utf8PathBuf;
 use serde::Serialize;
 
 use crate::{
-    config::{JinjaInstallPathStrategy, ZipStyle},
-    TargetTriple,
+    config::ZipStyle,
+    TargetTriple, ReleaseIdx,
 };
 
 use self::homebrew::HomebrewInstallerInfo;
@@ -41,20 +41,15 @@ pub enum InstallerImpl {
 pub struct InstallerInfo {
     /// The path to generate the installer at
     pub dest_path: Utf8PathBuf,
-    /// App name to use (display only)
-    pub app_name: String,
-    /// App version to use (display only)
-    pub app_version: String,
     /// URL of the directory where artifacts can be fetched from
     pub base_url: String,
-    /// Artifacts this installer can fetch
-    pub artifacts: Vec<ExecutableZipFragment>,
     /// Description of the installer (a good heading)
     pub desc: String,
     /// Hint for how to run the installer
     pub hint: String,
-    /// Where to install binaries
-    pub install_path: JinjaInstallPathStrategy,
+    /// The release this is installing
+    #[serde(skip)]
+    pub release: ReleaseIdx,
 }
 
 /// A fake fragment of an ExecutableZip artifact for installers
--- a/cargo-dist/src/platform.rs
+++ b/cargo-dist/src/platform.rs
@@ -1,172 +1,261 @@
 #![allow(missing_docs)]
 
+use axoproject::platforms::{
+    TARGET_ARM64_MAC, TARGET_ARM64_WINDOWS, TARGET_X64_MAC, TARGET_X64_WINDOWS, TARGET_X86_MAC,
+    TARGET_X86_WINDOWS,
+};
 use cargo_dist_schema::ArtifactId;
 
-use crate::{ArtifactIdx, TargetTriple, SortedMap, DistGraph, Release, DistGraphBuilder, ReleaseIdx, backend::installer::ExecutableZipFragment};
+use crate::{config::ZipStyle, DistGraphBuilder, ReleaseIdx, SortedMap, TargetTriple};
 
-const X64_MUSL_STATIC: &str = "x86_64-unknown-linux-musl-static";
-const X64_MUSL_DYNAMIC: &str = "x86_64-unknown-linux-musl-dynamic";
+/// Suffixes of TargetTriples that refer to statically linked linux libcs.
+///
+/// On Linux it's preferred to dynamically link libc *but* because the One True ABI
+/// is actually the Linux kernel syscall interface, you *can* theoretically statically
+/// link libc. This comes with various tradeoffs but the big selling point is that the
+/// Linux kernel is a much more slowly moving target, so you can build a binary
+/// that's portable across way more systems by statically linking libc. As such,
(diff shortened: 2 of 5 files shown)


cargo-dist/src/backend/installer/mod.rs
```diff
@@ -41,20 +41,15 @@ pub enum InstallerImpl {
...
     /// Description of the installer (a good heading)
     pub desc: String,
     /// Hint for how to run the installer
     pub hint: String,
-    /// Where to install binaries
-    pub install_path: JinjaInstallPathStrategy,
+    /// The release this is installing
+    #[serde(skip)]
+    pub release: ReleaseIdx,
 }
```
Gankra, line 54: This is the part that's over-aggressive and broke me. I hadn't appreciated that InstallerInfo just gets handed directly to each jinja template. We want each installer to lookup the Release and grab the PlatformSupport off of it. The reason this is late-bound like this is because we want PlatformSupport to be where installers grab things like shasums from, which in a sufficiently unified build would come after we compute the InstallerInfo (in principle we should be able to build binaries and then generate installers that fetch them in a single run of cargo-dist).

Gankra: Oh similarly "what glibc version do we need" and "linkage" are also late-bound like this. I think that's probably how I ended up doing a platform refactor when i was trying to do a linkage refactor -- in theory stuff like linkage is just like RuntimeConditions.

Gankra: I think if one had a mind to salvage this, they'd keep this change to include the ReleaseIdx but note delete all these other fields. Then write a shim that turns PlatformInfo into this "legacy" InstallerInfo/ExecutableZipFragment format. It won't include any of the "late-bound" info but we can then iteratively migrate each installer to using the latebound info (and also make sure it's populated).

With the goal of eventually dropping InstallerInfo/ExecutableZipFragment.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// The layers of emulation are out of control.
+    Hellmulated,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RuntimeCondition {
+    MinGlibcVersion { major: u64, series: u64 },
+    MinMuslVersion { major: u64, series: u64 },
+    Rosetta2,
+}
```
Gankra, line 62: NB the idea here is that jinja templates would include a check for RuntimeConditions on a FetchableArchive, and if they see an unknown one they would crash, basically forcing anyone adding a new one to add support to all installers.

Of course an installer can individually opt to ignore it, but it should be opt-out to handle these, not opt-in.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,261 @@
...
+    pub archives: Vec<FetchableArchive>,
+    pub platforms: SortedMap<TargetTriple, Vec<PlatformEntry>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct FetchableArchive {
+    pub id: ArtifactId,
+    pub native_runtime_conditions: Vec<RuntimeCondition>,
+    pub target_triples: Vec<TargetTriple>,
+    pub sha256sum: Option<String>,
```
Gankra, line 75: this is an Option so that it can be populated post-local-builds.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// This should be treated as a last resort, but hey, it works!
+    Emulated,
+    /// The layers of emulation are out of control.
+    Hellmulated,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RuntimeCondition {
+    MinGlibcVersion { major: u64, series: u64 },
+    MinMuslVersion { major: u64, series: u64 },
```
Gankra, line 60: i never quite worked out how these would get updated to concrete values post-build. mostly these are sketches. I think maybe the idea would be that you'd rerun `PlatformSupport::new` post-builds and it would know to add these conditions with the refined info from the DistManifest or something (similar to how linkage works).

Gankra: at very least we could just default these to the values we hardcode installers now so the value is in one central location.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// The archive is still technically native to this platform, but it's in some sense
+    /// imperfect. This can happen for things like "running a 32-bit binary on 64-bit" or
+    /// "using a statically linked linux libc". This solution is acceptable, but a HostNative
+    /// (or BulkyNative) solution should always be preferred.
+    ImperfectNative,
+    /// The archive is only running by the grace of pretty heavyweight emulation like Rosetta2.
+    /// This should be treated as a last resort, but hey, it works!
+    Emulated,
+    /// The layers of emulation are out of control.
+    Hellmulated,
```
Gankra, line 54: shipping this would make the refactor worth it

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,261 @@
...
+        for (_platform, support) in &mut platforms {
+            support.sort_by(|a, b| {
+                // Sort by SupportQuality, tie break by artifact name (for stability)
+                a.quality.cmp(&b.quality).then_with(|| {
+                    let archive_a = &archives[a.archive_idx];
+                    let archive_b = &archives[b.archive_idx];
+                    archive_a.id.cmp(&archive_b.id)
+                })
+            });
+        }
```
Gankra, line 134: This is the purpose of the SupportQuality thing. Oh right, providing all options for a platform would also theoretically allow an installer to say "there's a sickrad option, but i can't check for some RuntimeConditions it requires, so i'm gonna use a worse one i'm confident in".

Nice to have the option.

cargo-dist/src/backend/installer/mod.rs
Gankra: i think another thing that overwhelmed me here was that i started convincing myself there should be huge overhauls to dist-manifest (there should) where maybe PlatformSupport replaces a ton of stuff in there (less obvious).

PR #808, opened 2024-02-20
fix: make powershell exprs more reliable

### LINKED ISSUE #458: wrap irm-iex exprs in pwsh?
See also: https://github.com/axodotdev/oranda/issues/393

We currently emit irm-iex (the windows version of curl-sh) exprs of the form:

```
irm https://github.com/axodotdev/oranda/releases/download/v0.4.1/oranda-installer.ps1 | iex
```

This expression only works in powershell, and will fail if pasted into e.g. cmd.exe. Wrapping it in "launch powershell" as follows would make it work more universally, though I don't know if there's Weird Implications:

```
pwsh -c "irm https://github.com/axodotdev/oranda/releases/download/v0.4.1/oranda-installer.ps1 | iex"
```

Gankra:
fixes #458


ashleygwilliams (commented):
is this "breaking" at all?

Gankra:
not that i could fine

ashleygwilliams (approved):

PR #816, opened 2024-02-23
fix: handle mismatched formula names in brew publish

Gankra:


cargo-dist/templates/ci/github_ci.yml.j2
```diff
@@ -487,16 +487,19 @@ jobs:
...
       - name: Commit formula files
         run: |
           git config --global user.name "${GITHUB_USER}"
           git config --global user.email "${GITHUB_EMAIL}"
 
-          for release in $(echo "$PLAN" | jq --compact-output '.releases[]'); do
-            name=$(echo "$release" | jq .app_name --raw-output)
+          for release in $(echo "$PLAN" | jq --compact-output '.releases[] | select([.artifacts[] | endswith(".rb")] | any)'); do
+            filename=$(echo "$release" | jq '.artifacts[] | select(endswith(".rb"))' --raw-output)
+            name=$(echo "$filename" | sed "s/.rb//")
```
Gankra, line 499: in theory this should have some kind of ends-with anchor

cargo-dist/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cargo-dist"
 description = "Shippable application packaging for Rust"
-version = "0.11.0"
+version = "0.11.1-prerelease.1"
```
ashleygwilliams, line 4: is this intentional?

ashleygwilliams (approved):
lgtm modulo the version bumps but maybe those are on purpose

PR #818, opened 2024-02-26
fix: formula files pass brew lint

ucodery:


ucodery:
These changes don't actually fully satisfy test-bot. Dependnig on what exactly gets published, the tests can still fail because related branches aren't collapsed. This is the diff for my project between what cargo-dist did and what brew style accepts https://github.com/ucodery/homebrew-draft/compare/8171f0b..651c4ef

cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -68,29 +68,29 @@ class {{ formula_class }} < Formula
 
   def install
     {%- if arm64_macos.binaries %}
-    on_macos do
-      on_arm do
+    if OS.mac?
```
mistydemeo, line 71: Like the audit you ran suggested, if we're making changes here anyway, we might as well combine these conditions onto a single line.

mistydemeo (changes requested):
Thank you for this! I'm happy for us to make these changes, but just to set expectations: while we can make improvements wherever they come up, we can't guarantee that the Homebrew formulas we produce will always pass Homebrew's `brew audit`. It's a bit of a moving target, and when it comes to stylistic notes, it's more about Homebrew's own style guide rather than a strict set of requirements that third party taps need to abide by.

cargo-dist/templates/installer/homebrew.rb.j2
ucodery: makes sense. I combined the ifs but left it as a statement. It could be all one line like what the audit did, but I'm not sure how long the install lines are expected to get in practice.

mistydemeo (approved):

PR #824, opened 2024-02-29
Remove rust-toolchain.yml

mistydemeo:
Seeing if we can safely remove this now that it's been a bit.


mistydemeo:
Nope, looks like the Buildjet runners have 1.72.

PR #825, opened 2024-03-02
Run "rustup target add" if crossing building to Windows ARM64

AustinWise:
Similar to what macOS and musl do. The target is not installed by default on GitHub Actions runners, unlike x86 and x64.

Perhaps we should always run rustup when the host and target are different? It does not hurt: it only takes a couple of seconds and `rustup` exits successfully when the target is already installed. The only downside is you would get a warning when rustup is missing.

### DIFF
--- a/cargo-dist/src/cargo_build.rs
+++ b/cargo-dist/src/cargo_build.rs
@@ -91,6 +91,19 @@ impl<'a> DistGraphBuilder<'a> {
                 }
             }
 
+            if target == "aarch64-pc-windows-msvc"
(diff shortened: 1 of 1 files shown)


Gankra:
i agree we should just do it unconditionally for all crosses, although it *does* incur costs on a user if they do this too much (every toolchain update thinks it needs to fetch all of them again).

Gankra (approved):
approved but lmk if you wanna do the unified impl (we already have logic for pretty platform names in the changelog code)

AustinWise:
I posted a unified version here: #846

PR #835, opened 2024-03-06
Enable fallback to x64 in Invoke-Installer

### LINKED ISSUE #834: Installing on ARM64 Windows machines fails to fall back to x64 correctly
In `install.ps1`, the `Download` method contains code to fall back to x64 if ARM64 isn't in the set of supported platforms, but the corresponding logic is missing from `Invoke-Installer`.

Concretely, this causes installation of `uv` to fail on my Surface Pro 9 (5G), with the message:

> Downloading uv 0.1.15 (x86_64-pc-windows-msvc)
> Installing to C:\Users\kebatt\.cargo\bin
> &nbsp;&nbsp;uv.exe
> Cannot index into a null array.

(see https://github.com/astral-sh/uv/issues/2246)

kbattocchi:
Fixes #834.

Right now, if `platforms` doesn't contain an entry for ARM64, `Download` falls back to x64, but `Invoke-Install` is missing matching fallback behavior, leading to failure.

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -82,7 +82,7 @@ function Install-Binary($install_args) {
 
   $fetched = Download "$ArtifactDownloadUrl" $platforms
   # FIXME: add a flag that lets the user not do this step
-  Invoke-Installer $fetched $platforms "$install_args"
+  Invoke-Installer -bin_paths $fetched -platforms $platforms "$install_args"
 }
 
 function Get-TargetTriple() {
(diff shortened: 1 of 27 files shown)


mistydemeo:
Thank you! This is a good catch.

mistydemeo:
@Gankra I'm having some trouble satisfying PSScriptAnalyzer here - any thoughts? Or know where we'd put a PSScriptAnalyzer config file?

Gankra (approved):
ah i see, this fallback logic should be all the way back up in tasks.rs, but this existing trick already is present, so it's fine to replicate it here.

PR #837, opened 2024-03-07
feat: allow overriding install path

mistydemeo:
This provides a way to override any configuration on the installer and install to a location specified by a new environment variable. It provides a more predictable way to insist that it gets installed to a specific place, regardless of how the installer was built.

Possible fix for axodotdev/axoupdater#32, where we'd like to be able to install a new version to exactly the same path as the previous version.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -318,8 +318,6 @@ install() {
     else
         err "could not find your CARGO_HOME or HOME dir to install binaries to"
     fi
-    # Replace the temporary cargo home with the calculated one
-    RECEIPT=$(echo "$RECEIPT" | sed "s,AXO_CARGO_HOME,$_install_dir,")
 {% elif install_path.kind == "HomeSubdir" %}
     # Install to this subdir of the user's home dir
     # In this case we want to be late-bound, as $HOME is reliable/nice.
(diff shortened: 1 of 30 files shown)


Gankra:
is there any reason to do this an env-var and not a cli flag..? i suppose for curl-sh it's harder

mistydemeo:
This also strikes me as a power user/internal feature, which feels better-suited to being an environment variable instead of a CLI flag.

Gankra:
It *is* a power-user feature, but it's also an *end-power-user* feature, right? like they won't even know what "cargo-dist" is?

mistydemeo:
I don't think end users should use this; this is a feature for the updater which needs fine-grained control over installation location. That other people can use it is kind of a bonus.

Gankra:
oh I thought this was also motivated by some other astral asks to let the end-user install to a custom dir...

mistydemeo:
Oh, I think I totally missed that side of the conversation.

mistydemeo:
One more factor: the shell installer errors on unrecognized flags. An environment variable helps us guard against the possibility of using an installer version that doesn't recognize the new flag, because it'll silently ignore the instruction instead of hard-erroring and not performing the install. eg

```sh
$ ./axolotlsay-installer.sh --force-install-prefix /tmp/axolotlsay
ERROR: unknown option --force-install-prefix
```

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -345,6 +345,16 @@ install() {
...
 {% endif %}
+    # ...ignoring all of the above, if the user asked us to completely override
+    # those choices and use a specified directory, then pick that now
+    if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then
+        _install_home="$CARGO_DIST_FORCE_INSTALL_DIR"
+        _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
+        _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
+        _install_dir_expr='$CARGO_DIST_FORCE_INSTALL_DIR/bin'
+        _env_script_path_expr='$CARGO_DIST_FORCE_INSTALL_DIR/env'
+    fi
```
Gankra, line 356: the powershell code seems to be carefully making sure the install receipt gets the right value, but not so for shell? ` RECEIPT=$(echo "$RECEIPT" | sed "s,AXO_CARGO_HOME,$_install_dir,")` occurs much higher up...

Gankra (approved):
one issue but basic idea seems fine

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Good call, updated.

PR #846, opened 2024-03-11
Always invoke "rustup target add" when crosscompiling

AustinWise:
On GitHub actions, adding that target seems to take a about 2 seconds. It's about the same amount of time whether or not the target was already installed.

### DIFF
--- a/cargo-dist/src/cargo_build.rs
+++ b/cargo-dist/src/cargo_build.rs
@@ -64,8 +64,7 @@ impl<'a> DistGraphBuilder<'a> {
 
             // If we're trying to cross-compile, ensure the rustup toolchain
             // is setup!
-            if target != self.inner.tools.cargo.host_target
-            {
+            if target != self.inner.tools.cargo.host_target {
                 if let Some(rustup) = self.inner.tools.rustup.clone() {
                     builds.push(BuildStep::Rustup(RustupStep {
                         rustup,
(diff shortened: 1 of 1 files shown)


AustinWise:
In this comment:

https://github.com/axodotdev/cargo-dist/pull/825#pullrequestreview-1923430668

A comment was made about "pretty platform names". I was not sure how to incorporate that into this change. Should the warning use pretty platform names? Should the check for platforms use the pretty names??

cargo-dist/src/cargo_build.rs
```diff
@@ -62,32 +62,17 @@ impl<'a> DistGraphBuilder<'a> {
...
             }
 
-            // If we're trying to cross-compile on macOS, ensure the rustup toolchain
+            // If we're trying to cross-compile, ensure the rustup toolchain
             // is setup!
-            if target.ends_with("apple-darwin")
-                && self.inner.tools.cargo.host_target.ends_with("apple-darwin")
-                && target != self.inner.tools.cargo.host_target
+            if target != self.inner.tools.cargo.host_target
             {
```
mistydemeo, line 68: Mind moving this onto the previous line? `rustfmt` flags this

mistydemeo (approved):
Thank you! Makes sense to me. There's just one tweak I'd like to see before we merge

cargo-dist/src/cargo_build.rs
AustinWise: Opps, I did not notice that CI failure. I ran `cargo fmt` and pushed.

mistydemeo: Thanks!

mistydemeo (approved):

Issue #843, opened 2024-03-08
dist-manifest 1.0


Gankra:
The cargo-dist [dist-manifest.json format](https://opensource.axo.dev/cargo-dist/book/reference/schema.html) is an amalgam of bolted on concepts and features and needs a major cleanup.

## per-machine info

dist-manifest is lacking a proper notion of "per-machine" output info, where things like linkage (what packages you dynamically linked, sdk/libc versions), toolchain-info (compiler version, os version), and unpredictable values (debuginfo, hashes) can be transferred. This info could be consumed by installers (checksums, sysdeps), checkers (complain about dynamic openssl), and receipts (SBOMs, attestation, reproducibility).

Linkage has a partial implementation but it's more "over there" and doesn't really scale.

The totally-global "system_info" subsection is a useless farce because of this.

## remove changelogs

changelogs, like READMEs are a classic example of "arbitrary user input that can get infinitely long and cause serious issues to a metadata system". it should ideally not be contained in the dist-manifest. This raises some questions about where if anywhere oranda should get "clean" changelogs from, but at worst we can just teach it how to "strip" the autogen tables in a github release.

Gankra:
I almost started doing this in #793 and then quickly backed off as the abyss of that PR grew deeper and deeper... but like doing this would make that change a lot cleaner.

ashleygwilliams:
@Gankra is this "done" - i know that #848 at "worst" made a ton of progress here

mistydemeo:
Checking in again - we made substantive changes in #848. Do you see us needing additional changes?

Gankra:
Oh hmm! We should probably do a look over it at some point but I think yeah the main issues have been fixed. Although in a 1.0 we could breaks some parser compat and delete some lingering deprecated fields.

Gankra:
Remaining work is deleting deprecated fields:

* system_info: can just do it
* linkage: needs us to immediately update oranda too
* announcement_changelog: see github_body
* announcement_github_body: need some way to get this info to announcement task in github ci (upload a file as an artifact?)

Gankra:
Oh and maybe we should write some docs about philosophy/structure/patterns to follow when adding stuff to this

PR #848, opened 2024-03-11
Manifest reform

Gankra:
Works towards #843 by introducing top-level "systems" and "assets" maps, similar to artifacts, which are indexed by other items.

The biggest challenge of this is to rework the linkage subsystem to properly refer to binaries by id. It also adds inline checksum values to artifacts, with sha256 checksums being guaranteed on archives.

* [x] add new fields to manifest (dummied)
* [x] refactor build code for managing binaries
* [x] compute linkage bottom up
* [x] flow data to manifest
  * [x] compute system ids ("{subcommand}:{artifacts_mode}:{targets.join(',')}"?)
  * [x] compute asset ids (guid generated on the system that builds it?) (annoying for snapshot stability, might need safe mode...)  
  * [x] bubble linkage up so it can be stored in dist-manifest (or pass manifest down...?)
  * [x] implement manifest merging for new fields
  * [x] make artifacts[] reference assets[]
  * [x] implement artifacts[] merging
  * [x] add new artifact_uploads[] so it's okay for artifacts[] to contain old info
  * [x] change ci.yml to use artifact_uploads[] instead of artifacts[]
* [x] stop using old linkage format
  * [x] make homebrew dep code read linkage from new manifest location
  * [x] make linkage checker read linkage from new manifest location
    * [x] honestly just remove cargo_dist's Linkage types, only use cargo_dist_schema's? 
* [x] make --artifacts=lies (build_fake) add fake linkage?


Gankra:
https://github.com/Gankra/cargodisttest/releases/tag/v0.6.4

has a dist-manifest.json with the new system (see: assets[] and systems[])

Gankra:
i've now added inline checksums to the manifest, and changed the homebrew code to use both that and New Linkage.

Gankra:
lmfao, real fake checksums are uhhhh a problem for RUIN_ME_WITH_INSTALLERS, might need to make a tweak here...

image

Gankra:
nice, new checksum/brew stuff appears to have worked: https://github.com/Gankra/cargodisttest/releases/tag/v0.6.5

cargo-dist/src/build/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+        let Some(pkg) = self.packages.get_mut(&pkg_id) else {
+            return;
+        };
+
+        // lookup the binary in the package
+        let Some(bin_name) = src_path.file_stem() else {
+            return;
+        };
+        let Some(bin_result) = pkg.binaries.get_mut(bin_name) else {
+            return;
```
mistydemeo, line 104: I like separating out/deduplicating this logic! Much cleaner.

That said, I feel like we should add an error case for if the binary it's been passed doesn't exist in the expected set it's looking at. When this logic was centralized in one place, we could be pretty sure we wouldn't end up with "bad" unexpected binaries; now that this is a method that could hypothetically get called from anywhere, that logic error is easier to make in the future.

cargo-dist/src/build/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+            determine_linkage(src_path, target)?.to_schema()
+        };
+        let bin = dist.binary(src.idx);
+        manifest.assets.insert(
+            bin.id.clone(),
+            AssetInfo {
+                id: bin.id.clone(),
+                name: bin.name.clone(),
+                system: dist.system_id.clone(),
+                linkage: Some(linkage),
```
mistydemeo, line 199: Mapping this per-asset instead of in a big array is going to be a *lot* nicer going forward, imo

cargo-dist/src/tasks.rs
```diff
@@ -137,6 +137,12 @@ pub struct BinaryIdx(pub usize);
 /// It also allows us to report what *should* happen without actually doing it.
 #[derive(Debug)]
 pub struct DistGraph {
+    /// Unique id for the system we're building on.
+    ///
+    /// Since the whole premise of cargo-dist is to invoke it once on each machine, and no
+    /// two machines have any reason to have the exact same CLI args for cargo-dist, we
```
mistydemeo, line 143: This makes sense but we should definitely validate it.

cargo-dist/src/build/fake.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// build fake binaries, and emulate the build process as much as possible
+fn build_fake_binaries(
+    dist: &DistGraph,
+    manifest: &mut DistManifest,
+    binaries: &[BinaryIdx],
+) -> DistResult<()> {
+    // Shove these in a temp dir inside the dist dir, where it's safe for us to do whatever
+    let tempdir = dist.dist_dir.join("_fake_tmp");
```
mistydemeo, line 42: Can we make this a call to `tempdir` instead of shoving something inside the dist dir?

mistydemeo (commented):
In general I like the way this is going! I'm going to need to kick the tires a bit to get a feel for what the data looks like in practice, but I think the approach is valid! I'll do some testing on this soon and look at the resulting manifests.

cargo-dist/src/tasks.rs
Gankra: The big place where this falls over right now is `all:` which ends up being every invocation of `plan/host`. in my original initial plan i was going to have it be `{cmd}:{artifacts}:{targets}` but this is *slightly* annoying since the calling command isn't really passed down. in practice it doesn't matter since plan/host makes no artifacts... hmm

cargo-dist/src/build/fake.rs
Gankra: part of me felt like it was "right" for the build outputs to appear in the build dir, but... yeah...

cargo-dist/src/build/mod.rs
Gankra: So it's actually Intended Design that this API gets handed a bunch of junk paths, because this API is specifically processing the output from rustc/cargo, which reports tons and tons of intermediate/uninteresting results.

Previously it was the job of build_cargo to filter those out, but I specifically moved the logic into the generic driver so that it could be more "head empty", and so other future backends could potentially similarly be like "i see a bunch of binaries in this dir, did we need any of these, oh wise BuildExpectations?"

mistydemeo: That makes sense; can we document that somewhere explicitly? I think I'm going to forget this and become surprised again in the future if this gets handed a bunch of unexpected files.

cargo-dist/src/tasks.rs
mistydemeo: If I understand correctly, inserting something else with an identical system ID is going to lead to confusing issues down the line and we'll read inconsistent data back out of the manifest earlier. We should try to make sure the IDs are actually unique at the point where we merge manifests.

mistydemeo (approved):

PR #859, opened 2024-03-19
doc: add updater docs

mistydemeo:


book/src/installers/updater.md
```diff
@@ -0,0 +1,16 @@
+# Self-updater
+
+> since 0.12.0
+
+NOTE: This feature is currently experimental.
+
+Ordinarily, your users will need to visit your website and download an installer for the latest release in order to upgrade. Users who installed your software via a package manager, like Homebrew and npm, can use that package manager to upgrade to the latest release. For users of the [shell] and [PowerShell] installers, you can provide your users with a standalone installation program to upgrade more conveniently.
```
ashleygwilliams, line 7: as a tool author i would probably want to know a little more about the user experience of the installer. if my tool is called bob- can my users run `bob update`? i think elaborating on the end user experience here and potentially showing an example would be great

ashleygwilliams (approved):
i think these are good but i suspect that we will quickly learn folks want more in depth docs- esp with a focus on what the end user experience is as well as probably an example of how to use the library (we should probably mention the library is in Rust)

book/src/installers/updater.md
mistydemeo: I've updated with a longer description, a sample user session, and a reference to the axoupdater documentation.

PR #862, opened 2024-03-20
chore: add ruin_me powershell tests

Gankra:

PR #863, opened 2024-03-20
fix: unify handling of install prefix

mistydemeo:
I previously tried to inline the env subdir reference, but this didn't work on PowerShell. Instead, using the same logic as CARGO_HOME should make sure it gets referenced properly *and* cleaned up so it's JSON-valid.

### DIFF
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -3018,7 +3018,7 @@ pub struct Provider {
 #[derive(Clone, Debug, Serialize)]
 pub struct InstallReceipt {
     /// The location on disk where this app was installed
-    pub install_prefix: Utf8PathBuf,
+    pub install_prefix: String,
     /// A list of all binaries installed by this app
     pub binaries: Vec<String>,
     /// Information about where to request information on new releases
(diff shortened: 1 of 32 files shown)


cargo-dist/src/tasks.rs
```diff
@@ -3047,11 +3047,12 @@ impl InstallReceipt {
...
+            InstallPathStrategy::CargoHome
+            | InstallPathStrategy::EnvSubdir {
+                env_key: _,
+                subdir: _,
+            } => "AXO_INSTALL_PREFIX".to_owned(),
             InstallPathStrategy::HomeSubdir { subdir } => format!("$HOME/{}", subdir),
-            InstallPathStrategy::EnvSubdir { env_key, subdir } => {
-                format!("${}/{}", env_key, subdir)
-            }
         };
```
Gankra, line 3056: i would love if this PR was just making us unconditionally put AXO_INSTALL_PREFIX in and then all the paths in the shell and powershell installer do the find-replace. future-proof for all installer path features (as-is i'm not sure this would handle the env-var that forces the install dir).

Gankra (changes requested):
needs to at least be rebased on #862 and to address the FIXME in it

cargo-dist/src/tasks.rs
mistydemeo: Seems reasonable; let me look into that.

Gankra (approved):

PR #868, opened 2024-03-21
chore: loosen deps

Gankra:


Gankra:
this is overly agressive but a fun experiment

PR #880, opened 2024-03-26
fix: fix tests

mistydemeo:


ashleygwilliams (approved):

mistydemeo:
The snapshots were left in an inconsistent state because of the order of PR merges yesterday; the actual tests were fine.

PR #894, opened 2024-04-02
fix: use mv instead of cp to copy files to their final destination

Gankra:
mv can do the replace as a nice clean fs transaction, while cp does a bunch of messy writes over time. when having a binary invoke its own installer to replace itself, the former works fine on linux, while the latter errors out about modifying an executing binary.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -363,7 +363,7 @@ install() {
     local _bins="$2"
     for _bin_name in $_bins; do
         local _bin="$_src_dir/$_bin_name"
-        ensure cp "$_bin" "$_install_dir"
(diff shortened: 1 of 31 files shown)


Gankra:
just to clarify a bit, I think the precisely important thing is that the old file is morally "deleted" and "recreated" as far as linux is concerned. linux is fine with open files getting deleted, because everyone who has it open is holding an opaque handle to it that can be kept alive even if it's no longer reachable on the file system.

after deletion, creating a new file at that location creates a file that is effectively unrelated to that handle, so the executing binary is effectively unmodified.

however cp presumably has no interest in deleting an old file and creating a new one, and probably like, truncates the old file to empty before starting to write its data. anything holding an open handle to the file *would* observe that and rightfully should freak out.

PR #898, opened 2024-04-02
chore: add cargo-dist self tests

Gankra:


Gankra:
hmm no i don't think this is quite what i want... cli test might be more right...

Issue #897, opened 2024-04-02
dogfood updater (library)


Gankra:
Add `cargo dist update` command.

I think it should be slightly special, in that the vast majority of cases people will want to run this command is when they're about to do `init`. For the sake of pedagogy (and making people Update Update Update), it would be nice if `update` just ran `init` for you by default. As such I propose the following slightly weird semantics:

* By default this should assume it should only be run in a dir you could run `cargo dist init` in
  * Add an `--only-bins` flag or something to opt out of this requirement
* If installed (or if no updates needed), re-exec as the new binary and run `init`
* Add a `--reinstall` flag or something to force the updater to assume the application is out of date and should be refetched (useful for testing)
* Add a `--version` flag to pick a version to install (useful for testing)
* Add a `--???` flag to say "hey generate an installer script and use that instead of fetching"

PR #899, opened 2024-04-02
Self update command

mistydemeo:
An alternative to #896; this provides a `cargo dist update` command using axoupdater's lib. It also, optionally (and by default) runs `cargo dist update` from the new release after performing a successful upgrade, as suggested in #897.

### DIFF
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -174,6 +174,10 @@ pub enum Commands {
     /// Host artifacts
     #[clap(disable_version_flag = true)]
     Host(HostArgs),
+
+    /// Performs a self-update, if a new version is available
+    #[clap(disable_version_flag = true)]
+    Update(UpdateArgs),
 }
 
 #[derive(Args, Clone, Debug)]
@@ -323,6 +327,9 @@ pub struct LinkageArgs {
 
 #[derive(Args, Clone, Debug)]
 pub struct HelpMarkdownArgs {}
+
+#[derive(Args, Clone, Debug)]
+pub struct UpdateArgs {}
 
 /// A style of CI to generate
 #[derive(ValueEnum, Clone, Copy, Debug)]
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -329,7 +329,12 @@ pub struct LinkageArgs {
 pub struct HelpMarkdownArgs {}
 
 #[derive(Args, Clone, Debug)]
-pub struct UpdateArgs {}
+pub struct UpdateArgs {
+    /// Run `cargo dist init` after performing an upgrade
+    #[clap(long)]
+    #[clap(default_value_t = true)]
+    pub run_init: bool,
+}
 
 /// A style of CI to generate
 #[derive(ValueEnum, Clone, Copy, Debug)]
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -330,10 +330,10 @@ pub struct HelpMarkdownArgs {}
 
 #[derive(Args, Clone, Debug)]
 pub struct UpdateArgs {
-    /// Run `cargo dist init` after performing an upgrade
+    /// Skip running `cargo dist init` after performing an upgrade
     #[clap(long)]
-    #[clap(default_value_t = true)]
-    pub run_init: bool,
+    #[clap(default_value_t = false)]
+    pub skip_init: bool,
 }
 
 /// A style of CI to generate
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -44,7 +44,7 @@ pub enum DistError {
 
     #[cfg(cli)]
     #[error(transparent)]
-    AxopupdaterError(#[from] axoupdater::AxoupdateError),
+    AxoupdaterError(#[from] axoupdater::AxoupdateError),
 
     /// A problem with a jinja template, which is always a cargo-dist bug
     #[error("Failed to render template")]
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -42,10 +42,6 @@ pub enum DistError {
     #[error(transparent)]
     FromUtf8Error(#[from] std::string::FromUtf8Error),
 
-    #[cfg(cli)]
-    #[error(transparent)]
-    AxoupdaterError(#[from] axoupdater::AxoupdateError),
-
     /// A problem with a jinja template, which is always a cargo-dist bug
     #[error("Failed to render template")]
     #[diagnostic(help("this is a bug in cargo-dist, let us know and we'll fix it: https://github.com/axodotdev/cargo-dist/issues/new"))]
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -42,6 +42,10 @@ pub enum DistError {
     #[error(transparent)]
     FromUtf8Error(#[from] std::string::FromUtf8Error),
 
+    #[cfg(cli)]
+    #[error(transparent)]
+    AxopupdaterError(#[from] axoupdater::AxoupdateError),
+
     /// A problem with a jinja template, which is always a cargo-dist bug
     #[error("Failed to render template")]
     #[diagnostic(help("this is a bug in cargo-dist, let us know and we'll fix it: https://github.com/axodotdev/cargo-dist/issues/new"))]
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -309,6 +309,11 @@ pub enum DistError {
         /// Name of binary
         bin_name: String,
     },
+
+    /// Error during `cargo dist update`
+    #[error("`cargo dist update` failed; the new version isn't in the place we expected")]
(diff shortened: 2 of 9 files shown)


cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
+    let mut updater = AxoUpdater::new_for("cargo-dist");
+    // Do we want to treat this as an error?
+    // Or do we want to sniff if this was a Homebrew installation?
+    if updater.load_receipt().is_err() {
+        println!("Unable to load install receipt to check for updates.");
+        println!("If you installed this via `brew`, please `brew upgrade cargo-dist`!");
+        return Ok(());
+    }
+
+    if !updater.check_receipt_is_for_this_executable()? {
```
Gankra, line 533: why only handle the boolean failure and not the error failure? why are these materially different?

cargo-dist/src/cli.rs
```diff
@@ -324,6 +328,14 @@ pub struct LinkageArgs {
 #[derive(Args, Clone, Debug)]
 pub struct HelpMarkdownArgs {}
 
+#[derive(Args, Clone, Debug)]
+pub struct UpdateArgs {
+    /// Run `cargo dist init` after performing an upgrade
+    #[clap(long)]
+    #[clap(default_value_t = true)]
+    pub run_init: bool,
```
Gankra, line 336: the polarity of this seems like messed up UX for the user. `--run-init=false`?

cargo-dist/src/errors.rs
```diff
@@ -42,6 +42,10 @@ pub enum DistError {
     #[error(transparent)]
     FromUtf8Error(#[from] std::string::FromUtf8Error),
 
+    #[cfg(cli)]
+    #[error(transparent)]
+    AxopupdaterError(#[from] axoupdater::AxoupdateError),
```
Gankra, line 47: typo of the variant

Gankra (changes requested):

cargo-dist/src/main.rs
mistydemeo: The boolean isn't an error, it's an answer to the question: is this receipt for this executable? The error is if we encountered an error trying to calculate an answer to that question.

cargo-dist/src/errors.rs
mistydemeo: Whoops, the `cfg` weirdly hid it. Will fix.

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
 }
+
+async fn cmd_update(_config: &Cli, args: &cli::UpdateArgs) -> Result<(), miette::ErrReport> {
+    let mut updater = AxoUpdater::new_for("cargo-dist");
+    // Do we want to treat this as an error?
+    // Or do we want to sniff if this was a Homebrew installation?
+    if updater.load_receipt().is_err() {
+        println!("Unable to load install receipt to check for updates.");
+        println!("If you installed this via `brew`, please `brew upgrade cargo-dist`!");
+        return Ok(());
```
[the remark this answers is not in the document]
mistydemeo: Right yes, I'll switch to `eprintln`. At this point axoupdater as a lib no longer prints anything outside its `main` as a CLI, and provides flags to shut up the underlying updater.

cargo-dist/src/cli.rs
mistydemeo: Right, you want to make this `skip_init` instead of `run_init=false`?

Gankra: yeah

mistydemeo: Cool, will do

cargo-dist/src/errors.rs
```diff
@@ -305,6 +305,11 @@ pub enum DistError {
         /// Name of binary
         bin_name: String,
     },
+
+    /// Error during `cargo dist update`
+    #[error("`cargo dist update` failed; the new version isn't in the place we expected")]
```
ashleygwilliams, line 310: is this the only way it can fail? it may be useful to give more info on "place" e.g. place on the web? place on the local machine?

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
     Ok(())
 }
+
+async fn cmd_update(_config: &Cli, args: &cli::UpdateArgs) -> Result<(), miette::ErrReport> {
+    let mut updater = AxoUpdater::new_for("cargo-dist");
+    // Do we want to treat this as an error?
+    // Or do we want to sniff if this was a Homebrew installation?
+    if updater.load_receipt().is_err() {
+        eprintln!("Unable to load install receipt to check for updates.");
+        eprintln!("If you installed this via `brew`, please `brew upgrade cargo-dist`!");
```
ashleygwilliams, line 529: do we want to make this generic such as "via a package manager, like `brew`"

also- could we detect if it was installed via brew and report that? that'd be *nice*

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
+        return Ok(());
+    }
+
+    if !updater.check_receipt_is_for_this_executable()? {
+        eprintln!("This installation of cargo-dist wasn't installed via a method that `cargo dist update` supports.");
+        eprintln!("Please upgrade manually.");
```
ashleygwilliams, line 535: the words update and upgrade are interchanged here- we probably want to keep it consistent

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
+    if let Some(result) = updater.run().await? {
+        eprintln!(
+            "Update performed: {} => {}",
+            env!("CARGO_PKG_VERSION"),
+            result.new_version
+        );
+
+        // At this point, we've either updated or bailed out;
+        // we can proceed with the init if the user would like us to.
+        if !args.skip_init {
```
ashleygwilliams, line 548: i think it might be useful to print something to the screen so the user knows what's happening

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,57 @@ fn cmd_manifest_schema(
...
+            }
+
+            let mut cmd = Cmd::new(new_path, "cargo dist init");
+            cmd.arg("dist").arg("init");
+            cmd.run()?;
+
+            return Ok(());
+        }
+    } else {
+        eprintln!("No update necessary; up to date");
```
ashleygwilliams, line 570: it might be nice to print the current version here

cargo-dist/tests/snapshots/long-help.snap
```diff
@@ -18,6 +18,7 @@ Commands:
   manifest  Generate the final build manifest without running any builds
   plan      Get a plan of what to build (and check project status)
   host      Host artifacts
+  update    Performs a self-update, if a new version is available
```
ashleygwilliams, line 21: given what we may do in the future do we want this to be "self update" ?

cargo-dist/src/errors.rs
mistydemeo: Ooh yeah, good call.

cargo-dist/src/main.rs
mistydemeo: I think we can sniff that it was installed via brew and provide more detailed info, yeah.

cargo-dist/tests/snapshots/long-help.snap
mistydemeo: Seems reasonable!

cargo-dist/src/main.rs
mistydemeo:
```
$ /usr/local/bin/cargo-dist dist update
Your copy of `cargo-dist` seems to have been installed via Homebrew.
Please run `brew upgrade cargo-dist` to update this copy.
```

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,84 @@ fn cmd_manifest_schema(
...
+    // Or do we want to sniff if this was a Homebrew installation?
+    if updater.load_receipt().is_err() {
+        eprintln!("Unable to load install receipt to check for updates.");
+        eprintln!("If you installed this via `brew`, please `brew upgrade cargo-dist`!");
+        return Ok(());
+    }
+
+    if !updater.check_receipt_is_for_this_executable()? {
+        eprintln!("This installation of cargo-dist wasn't installed via a method that `cargo dist update` supports.");
+        eprintln!("Please upgrade manually.");
```
Gankra, line 562:
```suggestion
        eprintln!("Please update manually.");
```

cargo-dist/src/main.rs
```diff
@@ -516,3 +519,84 @@ fn cmd_manifest_schema(
...
+            }
+
+            let mut cmd = Cmd::new(new_path, "cargo dist init");
+            cmd.arg("dist").arg("init");
+            cmd.run()?;
+
+            return Ok(());
+        }
+    } else {
+        eprintln!("No update necessary; up to date");
```
Gankra, line 597:
```suggestion
        eprintln!("No update necessary; {} is up to date.", env!("CARGO_PKG_VERSION"));
```

Gankra (approved):

PR #903, opened 2024-04-04
release: 0.12.2

mistydemeo:
This updates our cargo-dist and ensures cargo-dist's own installer incorporates the fix we shipped in 0.12.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 Nothing Yet!
 
 
+# Version 0.12.2 (2024-04-04)
+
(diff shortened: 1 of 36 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,21 @@
 Nothing Yet!
 
 
+# Version 0.12.2 (2024-04-04)
+
+This is a minor release which regenerates the installer using the fix from 0.12.1. It shouldn't affect most users.
```
ashleygwilliams, line 8: we may want to be careful with this messaging- i think it can always be helpful to be a little more specific about who it will affect (it only affects people who XYZ) and let that couching make it clear that it is a narrow effect

mistydemeo: That's a good call; I'll update to note that it affects users who try to run the installer while a cargo-dist instance is already running.

PR #906, opened 2024-04-05
prepare 0.13.0 release

Gankra:


cargo-dist/src/errors.rs
```diff
@@ -310,6 +307,37 @@ pub enum DistError {
     #[error("`cargo dist update` failed; the new version isn't in the place we expected")]
     #[diagnostic(help("This is probably not your fault, please file an issue!"))]
     UpdateFailed {},
+
+    /// Trying to run cargo dist update in a random dir
+    #[error("`cargo dist update` needs to be run in a project")]
```
mistydemeo, line 312:
```suggestion
    #[error("`cargo dist selfupdate` needs to be run in a project")]
```

cargo-dist/tests/snapshots/markdown-help.snap
```diff
@@ -155,8 +155,8 @@ Automatically accept all recommended/default values
 
 This is equivalent to just mashing ENTER over and over during the interactive prompts.
 
-#### `--no-generate`
-Don't automatically invoke 'cargo dist generate' at the end
+#### `--skip-generate`
```
mistydemeo, line 158: Are we okay with renaming this commandline option? Consistency is good, just want to avoid confusing people.

mistydemeo (approved):
Generally looking good, bar a couple minor comments.

This has one quirk: if you `cargo dist selfupdate` in a repo that's on an older cargo-dist than the one you're running, *and* the one you're running is the latest version, it won't go on to run `init`. We can pretty easily always run `init` in the case that we don't upgrade, just in case; that's probably safe. Being smarter about only running it if the config was outdated, copying the same mechanics from the update we tried to perform, would be a bit more work.

I've pushed a commit that handles the simpler case.

cargo-dist/tests/snapshots/markdown-help.snap
Gankra: There's aliases for the old name (alas not exposed in docs).

PR #917, opened 2024-04-09
fix: prepend extra newline to profiles

Gankra:


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -480,6 +480,8 @@ add_install_dir_to_path() {
             # (This will also create the rcfile if it doesn't exist)
             if [ -f "$_env_script_path" ]; then
                 say_verbose "adding $_robust_line to $_target"
+                # prepend an extra newline in case the user's file is missing a trailing one
+                ensure echo "" >> "$_target"
```
mistydemeo, line 484: This'll add a second newline if the file *does* end with one, which we might not be what we want. Do we want to try to detect if it ends in a newline and only add one if it doesn't?

mistydemeo (approved):

cargo-dist/templates/installer/installer.sh.j2
Gankra: So specifically we'll under any normal circumstances prepend a newline, producing

```
. "whatever/came/before"

. "$HOME/.cargo/env"

```

which I consider acceptable.

mistydemeo: That seems fair, yeah.

PR #918, opened 2024-04-09
fix: re-add HOME to paths in updater

Gankra:
I'm not sure if this is the ideal solution, because there's a few cases where we actually wouldn't have $HOME (EnvSubdir or the user actually setting $CARGO_HOME)... ideally we check for both forms.


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -348,8 +348,10 @@ install() {
         _install_home="$CARGO_DIST_FORCE_INSTALL_DIR"
         _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
```
Gankra, line 349: also this looks like another case where we're incorrectly assuming appending `/bin` is correct -- this won't work for HOME_SUBDIR users.

mistydemeo (approved):
Looks good - this is what I was planning to do too.

Gankra:
obsoleted by #928

PR #928, opened 2024-04-11
feat: clean up force install behaviour

mistydemeo:
Instead of always overriding after setting the path, this places it directly inside each of the branches. This lets us customize the values to be more appropriate to each branch, fixing an issue where `/bin` would be added to paths that shouldn't have it.

In addition, it adds a new check to several branches for if `$HOME` is contained within and substitutes the literal value for the variable in the `expr` values, and avoids overwriting the configured values if the forced values are already equivalent to the defaults.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -290,13 +290,11 @@ install() {
 {% if install_path.kind == "CargoHome" %}
     # first try CARGO_HOME, then fallback to HOME
     if [ -n "${CARGO_HOME:-}" ]; then
-        _install_home="$CARGO_HOME"
         _install_dir="$CARGO_HOME/bin"
         _env_script_path="$CARGO_HOME/env"
 
         # If we need to override the above
         if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ] && [ "${CARGO_DIST_FORCE_INSTALL_DIR}" != "$CARGO_HOME" ]; then
-            _install_home="$CARGO_DIST_FORCE_INSTALL_DIR"
             _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
             _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
         fi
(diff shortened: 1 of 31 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -336,21 +364,24 @@ install() {
...
         _env_script_path_expr="$_env_script_path"
+
+        # Override if necessary
+        if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then
+            _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
+            # Note that it's possible for the `env` file to have
+            # been configured to a totally different path, but we don't
+            # (yet) have the appropriate value to force that to here, so
+            # we use this instead.
+            _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
```
Gankra, line 375: Could you elaborate? the previous lines seem to always put these together

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -314,6 +322,17 @@ install() {
         _env_script_path="$HOME/.cargo/env"
         _install_dir_expr='$HOME/.cargo/bin'
         _env_script_path_expr='$HOME/.cargo/env'
+
+        # If we need to override the above
+        if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ] && [ "${CARGO_DIST_FORCE_INSTALL_DIR}" != "$HOME/.cargo" ]; then
+            _install_home="$CARGO_DIST_FORCE_INSTALL_DIR"
```
Gankra, line 328: ....i can't find any code that reads _install_home... can we just delete it from everywhere, or do you want to keep this around for future work?

Gankra (approved):

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: I think you're right, I was misreading the condition and thought it branched further. I'll remove the comment.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: It's funny, when I started I thought `$_install_home` was used to construct further paths, but it isn't. It's probably safe to remove it for now. It might get used at some point when we start doing more complex path structures (eg libs in `$_install_home/lib`), but we're not there yet so it can get removed until that happens. I'll do that in a followup commit.

PR #941, opened 2024-04-16
fix: check for rosetta in shell

mistydemeo:
We inherited a check for the old-style Mac `uname -m` reporting on Intel from rustup, but we're missing a case to check for Rosetta under aarch64. Makes sense to me to add one and always install the aarch64 binary even if the installer's being run from within an Intel Rosetta shell.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -627,6 +627,13 @@ get_architecture() {
         fi
     fi
 
+    if [ "$_ostype" = Darwin ] && [ "$_cputype" = x86_64 ]; then
+        # Rosetta on aarch64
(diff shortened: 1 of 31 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -627,6 +627,13 @@ get_architecture() {
         fi
     fi
 
+    if [ "$_ostype" = Darwin ] && [ "$_cputype" = x86_64 ]; then
+        # Rosetta on aarch64
+        if [ "$(sysctl -n hw.optional.arm64 2>/dev/null)" = "1" ]; then
```
mistydemeo, line 632: I used `sysctl -n` instead of piping into grep for a bit of a cleaner comparison.

Gankra (approved):

mistydemeo:
Confirmed this works when running in both rosetta and non-rosetta shells on arm64.

PR #944, opened 2024-04-17
feat: migrate from flat to hierarchical install strategy 

### LINKED ISSUE #934: normalize install dir structure
Because we initially regarded the CARGO_HOME structure as "overkill" the `~/some/path` and `$MY_ENV_VAR/some/path` installdir strategies do not make a separate `bin/` subfolder, which limits the scalability of the approach somewhat, and adds a ton of special case paths to installers/updaters.

Also the MSI installer really does its own thing altogether when installing to ProgramFiles.

We should change everything to be shaped like CARGO_HOME, with `./bin/` and `./env` (on unix), with room to expand to having libs dirs as well.

This will require migrating existing users of `~/some/path`, but will be good longterm.

mistydemeo:
Before this commit, we had two different installation layouts:

* Cargo home produced an FHS-style installation layout, with binaries installed into the `bin` directory within an install prefix.
* Env subdir and home subdir produced a flat installation layout, with binaries and all other files placed in a single directory.

Going forward, it will be advantageous for us to have a unified strategy where we can predict the layout of an install directory no matter which path we're following.

This doesn't just change the default, but also performs a migration: any binaries that this installer would install that exist in the old default path are migrated to the new path before the installation itself begins.

We'll want to make sure this gets some testing before merging, though the "ruin me" env var tests in CI should also give us some active testing.

Fixes #934.

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -120,6 +120,18 @@ function Get-TargetTriple() {
   }
 }
 
+# Provides a way to migrate from an old "flat" layout to the new
+# FHS-style organized layout.
+function MigrateInstallDir($old_root, $new_root, $bin_paths) {
+  foreach ($bin_name in $bin_paths) {
+    $bin_path = "$old_root\$bin_name"
+      if (Test-Path -Path "$bin_path") {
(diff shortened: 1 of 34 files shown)


Gankra:
closing this for now

PR #947, opened 2024-04-18
feat: multiple install paths with fallbacks

mistydemeo:
This adds a new feature which allows specifying multiple paths in `install-path` instead of a single path. If this is set, then we use the next paths in the set as fallbacks in case the earlier options aren't available. As a result, path errors for `HomeSubdir` and `EnvSubdir` that were formerly fatal instead cause us to fall back to the next case; the only actual fatal error is if we exhaust the array of options and find nothing viable.

I've added tests for two fallback options and confirmed they work, along with a test for what happens if the first path in the array is valid and we *don't* perform a fallback.

I haven't fully massaged the spacing in the jinja templates to keep it from adding extra newlines in places that don't need them. I did add a comment for people reading the generated script who might be confused by what looks like a pointless `if` statement.

This isn't yet backwards compatible with existing `Cargo.toml`s for users already using `install-path`. I've temporarily adjusted `install-path` to always be an array, which will cause errors for users already using it as a string. We'll want to adjust the parsing so that a string parses to a one-length array.


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -34,6 +34,49 @@ RECEIPT_HOME="${HOME}/.config/{{ app_name }}"
...
+# this looks strange to you, it's because it's possible for there
+# to be multiple generated paths here. If you're only seeing one
+# $install_path assignment, it's because the app you're looking at
+# has been configured with a single install path.
+{% for install_path in install_paths %}
+if [ -z "$INSTALL_PATH_STRATEGY" ]; then
+{% if install_path.kind == "CargoHome" %}
+    # This strategy can't fail; always assign it
+    INSTALL_PATH_STRATEGY="CargoHome"
+    INSTALL_PATH_HINT="\$CARGO_HOME/bin (\$HOME/.cargo/bin)"
```
Gankra, line 54: Yeah i still think the install-hints shouldn't be this dynamic. If you ask the script what it does, it should report its actual configurable behaviour, and not just what it will do with the current set of env-vars. That said if we're printing out a list of options, I suppose I wouldn't complain if the printout showed the expression *and* its computed value.

Like

```text
then unpacks the binaries and installs them to the first defined option:
   $MY_COMPANY (unset)
   $MY_COMPANY_LEGACY/bin = /home/gankra/corpcrud/bin
   $HOME/.mycompany/bin = /home/gankra/.mycompany/bin
```

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,7 +323,7 @@ install() {
     # Potentially-late-bound version of env_script_path to write to rcfiles like $HOME/.profile
     local _env_script_path_expr
 
-{% if install_path.kind == "CargoHome" %}
+if [ "$INSTALL_PATH_STRATEGY" = "CargoHome" ]; then
```
Gankra, line 326: I'm not a huge fan of us unconditionally adding chunks of likely-dead-code to each install script.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -34,6 +34,49 @@ RECEIPT_HOME="${HOME}/.config/{{ app_name }}"
...
+        INSTALL_PATH_HINT="\$$INSTALL_PATH_ENV_KEY/$INSTALL_PATH_SUBDIR"
+    fi
+{% else %}
+    {{ error("unimplemented install_path format: " ~ install_path.kind) }}
+{% endif %}
+fi
+{% endfor %}
+
+if [ -z "$INSTALL_PATH_STRATEGY" ]; then
+    err "could not find a valid path to install to; please check the installation instructions"
```
Gankra, line 77: oh yeah this would also be especially cruel with dynamic help text, because trying to get the installation instructions would just get this message again ☠️ 

(we should factor out the subroutine of usage() that prints the options and show it here? or maybe just fully invoke usage()?)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -332,48 +368,41 @@ install() {
...
-        _install_dir="${{ install_path.env_key }}{% if install_path.subdir | length %}/{% endif %}{{ install_path.subdir }}"
-        _env_script_path="${{ install_path.env_key }}{% if install_path.subdir | length %}/{% endif %}{{ install_path.subdir }}/env"
+    _install_dir_expr="$_install_dir"
+    _env_script_path_expr="$_env_script_path"
+
+    # Override if necessary
+    if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then
+        _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
+        _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
         _install_dir_expr="$_install_dir"
```
Gankra, line 402: I think this is the fatal issue with this implementation: the install strategy that resolves *today* is completely unrelated to the install strategy that resolved when the app was first installed.

And I don't mean because a developer can change their cargo-dist config. I mean because and end-user's env-vars have to be regarded as ephemeral. We *specifically* allow $HOME to remain dynamic, but $MYCOMPANY is fully "substituted" during installation, on the assumption that it was temporarily set and won't be around for long.

As a result, the updater needs to trust the install script to behave correctly if it (re)runs an install script with different env-vars set. So all handling of CARGO_DIST_FORCE_INSTALL_DIR has to be unconditional, and not nested within any particular INSTALL_PATH_STRATEGY.

This is why I was pushing for the normalization stuff before doing this change -- it would make that trivially true.

This is also why I suggested a partial compromise: only allow CARGO_HOME in install-strategy if it's the *only* option. That way there's two cases for CARGO_DIST_FORCE_INSTALL_DIR:

* either it's *just* CARGO_HOME and we have to append /bin
* or it's an arbitrary pile of $ENVVAR and $HOME checks which use the same flat structure

in either case we get to emit uniform CARGO_DIST_FORCE_INSTALL_DIR logic

(Note: my proposal *will* result in us substituting $HOME more aggressively on re-installs if the install was done with an env-var, which in the current implementation will add a duplicate entry to rcfiles. This is not fatal just ugly. We could fix this by more checking for more variations when checking to see if we should edit rcfiles.)

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -12,7 +12,9 @@ The installer for {{ app_name }} {{ app_version }}
 
 This script detects what platform you're on and fetches an appropriate archive from
 {{ base_url }}
-then unpacks the binaries and installs them to {% if install_path.kind == "CargoHome" -%}
+then unpacks the binaries and installs them to{% if install_paths|length > 1 %} one of the following locations{% endif %}
+{% for install_path in install_paths %}
+{%- if install_path.kind == "CargoHome" -%}
     $env:CARGO_HOME\bin ($HOME\.cargo\bin)
 {%- elif install_path.kind == "HomeSubdir" -%}
```
Gankra, line 19: 💯

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -47,6 +50,47 @@ param (
 $app_name = '{{ app_name }}'
 $app_version = '{{ app_version }}'
 
+$install_path = $null
+# If you're reading this install script generated by an app, and
+# this looks strange to you, it's because it's possible for there
+# to be multiple generated paths here. If you're only seeing one
+# $install_path assignment, it's because the app you're looking at
+# has been configured with a single install path.
```
Gankra, line 58: I don't understand what this comment is trying to say.

Regardless, this implementation has the same issues I describe in the shell impl.

Gankra (changes requested):

Gankra:
Some pseudo-jinja-rust-shell code i think the impl should look like:

```rust
// start with nowhere to install
_install_dir = ""

// First off, if we get handed FORCE_INSTALL_DIR, just use that
if $CARGO_DIST_FORCE_INSTALL_DIR {
   // two major families of install-strategy, that the updater 
   %if strategies.contains("cargo_home") && strategies.len == 1
      _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
      _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
      _install_dir_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/bin" | sed "s,$HOME,\$HOME,")"
      _env_script_path_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/env" | sed "s,$HOME,\$HOME,")"
   %elseif !strategies.contains("cargo_home")
      _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
      _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
      _install_dir_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR" | sed "s,$HOME,\$HOME,")"
      _env_script_path_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/env" | sed "s,$HOME,\$HOME,")"
   %else
      jinjaerror "cannot combine CARGO_HOME with other strategies"
   %endif
}

// Try to find an appropriate place to install
%for strategy in strategies
  if _install_dir.is_empty() {
    %if strategy.kind == "cargo_home"
      if $CARGO_HOME {
        _install_dir = "..."
      } else if $HOME {
         _install_dir = "..."
      } else {
        // error i guess
      }
    %elseif strategy.kind == "home"
      if $HOME {
        _install_dir = "..."
      }
    %elseif strategy.kind == "env"
      if ${strategy.envvar} {
        _install_dir = "..."
      }
    %else
      jinjaerror "unimplemented strategy"
    %endif
  }
%endfor

if _install_dir.is_empty() {
  error "oopsie whoopsie no install dir"
}

// ok install paths handled, now do the rest of setup
...
```

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: That's reasonable, yeah. By comparison the Windows version *does* do that, so it would be better to have it be identical.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: By installation instructions I meant to refer people to the homepage, since our own usage text doesn't provide *that* much beyond "where would this be installed to", but we could still just reprint the usage text.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Take a look at the generated script this comment is hanging off: https://github.com/axodotdev/cargo-dist/pull/947/files#diff-dd947bc4ad8e756b1f8021216e9c9888f3e1635b526b7bfb289febdad0c6b889R54

If there's *only one* path set, then you get a weird looking condition where we assign an empty string, immediately check if that string is empty and do something, then after that check again if the string is empty. This only looks "normal" in scripts with multiple strategies assigned. It serves as an explanation that there's a reason for this to make sense even if the specific generated file they've seen looks odd.

mistydemeo:
I think your suggestion makes sense, and I did write this with the consideration we might have some more ideas about #944. In the absence of that, yeah, we have to avoid mixing strategies. I wonder if it makes more sense to consider making this an error in Rust at the place we produce the installer stuff before feeding it into jinja, or even when loading the user's config, in order to make sure we get a nicer-looking error.

Gankra:
> I wonder if it makes more sense to consider making this an error in Rust at the place we produce the installer stuff before feeding it into jinja, or even when loading the user's config, in order to make sure we get a nicer-looking error.

oh yes absolutely, i just decided to fold it into the jinja to avoid making the idea more complicated to express

Gankra (changes requested):

cargo-dist/tests/snapshots/install_path_no_fallback_taken.snap
```diff
@@ -89,8 +86,8 @@ The installer for axolotlsay 0.2.1
 
 This script detects what platform you're on and fetches an appropriate archive from
 https://github.com/axodotdev/axolotlsay/releases/download/v0.2.1
-then unpacks the binaries and installs them to
-$INSTALL_PATH_HINT
+then unpacks the binaries and installs them to one of the following locations
+\$HOME/.axolotlsay\$MY_ENV_VAR/My Axolotlsay Documents/bin
```
Gankra, line 90: looks like missing a newline

mistydemeo: Ugh yes, I get the jinja whitespace squelching stuff wrong at least 50% of the time

cargo-dist/src/errors.rs
```diff
@@ -330,6 +330,11 @@ pub enum DistError {
         #[diagnostic_source]
         cause: ProjectError,
     },
+
+    /// Trying to include CargoHome with other install paths
+    #[error("Incompatible install paths configured in Cargo.toml")]
+    #[diagnostic(help("The CargoHome `install-path` configuration can't be combined with other install path strategies."))]
+    IncompatibleInstallPathConfiguration {},
```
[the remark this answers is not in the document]
mistydemeo: I was sure it got auto-added by something at some point and I accidentally internalized it, if I can eliminate it I'm happy to

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
+        # For these two, if the new value contains $HOME, make
+        # sure to `sed` the $HOME expression back into the expr
+        _install_dir_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR" | sed "s,$HOME,\$HOME,")"
+        _env_script_path_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/env" | sed "s,$HOME,\$HOME,")"
+        {% endif %}
+    # If we're *not* overriding, go ahead with the more complex
+    # path evaluation logic
+    else
+    {% for install_path in install_paths %}
+        if [ -z "${_install_dir:-}" ]; then
```
Gankra, line 313: you don't need this else, it's implicit in the fact that `CARGO_DIST_FORCE_INSTALL_DIR` will set `_install_dir`, which nicely makes all the options look like a clean cascade :)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
+        {% if install_paths| selectattr("kind", "equalto", "CargoHome") %}
+        _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
+        _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
+        _install_dir_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/bin" | sed "s,$HOME,\$HOME,")"
+        _env_script_path_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/env" | sed "s,$HOME,\$HOME,")"
+        {%- else -%}
+        _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
+        _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
+        # For these two, if the new value contains $HOME, make
+        # sure to `sed` the $HOME expression back into the expr
```
Gankra, line 305: this comment should be shared between both paths

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
+                        _env_script_path_expr='$HOME/.cargo/env'
+                    else
+                        _install_dir_expr="$_install_dir"
+                        _env_script_path_expr="$_env_script_path"
+                    fi
+                else
+                    _install_dir_expr="$_install_dir"
+                    _env_script_path_expr="$_env_script_path"
+                fi
+            elif [ -n "${HOME:-}" ]; then
```
Gankra, line 335: can/should we restructure this so it looks like the rest of the cascade options..? (especially the final error condition)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
-        if [ -n "${HOME:-}" ]; then
-            if [ "$HOME/.cargo/bin" = "$_install_dir" ]; then
+    # Before actually consulting the configured install strategy, see
+    # if we're overriding it.
+    if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then
+        {% if install_paths| selectattr("kind", "equalto", "CargoHome") %}
+        _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR/bin"
+        _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
+        _install_dir_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/bin" | sed "s,$HOME,\$HOME,")"
+        _env_script_path_expr="$(echo "$CARGO_DIST_FORCE_INSTALL_DIR/env" | sed "s,$HOME,\$HOME,")"
```
Gankra, line 300: ooh i just realized that for all our paranoid checking of `$HOME` existing, these expressions will be messed up if it's not set (right..?)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
+                        _install_dir_expr='$HOME/.cargo/bin'
+                        _env_script_path_expr='$HOME/.cargo/env'
+                    else
+                        _install_dir_expr="$_install_dir"
+                        _env_script_path_expr="$_env_script_path"
+                    fi
+                else
+                    _install_dir_expr="$_install_dir"
+                    _env_script_path_expr="$_env_script_path"
+                fi
```
Gankra, line 334: Can this all be replaced with the sed approach?

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
-            _install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
-            _env_script_path="$CARGO_DIST_FORCE_INSTALL_DIR/env"
-            _install_dir_expr="$_install_dir"
-            _env_script_path_expr="$_env_script_path"
-        fi
-    else
-        err "could not find your {{ install_path.env_key }} dir to install binaries to"
+
+    if [ -z "$_install_dir_expr" ]; then
+        err "could not find a valid path to install to!\n\n$(usage)"
```
Gankra, line 369: wait i thought `\n` didn't work in shell

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -287,93 +290,84 @@ install() {
...
-
-        # If CARGO_HOME was set but it ended up being the default $HOME-based path,
-        # then keep things late-bound. Otherwise bake the value for safety.
-        # This is what rustup does, and accurately reproducing it is useful.
-        if [ -n "${HOME:-}" ]; then
-            if [ "$HOME/.cargo/bin" = "$_install_dir" ]; then
+    # Before actually consulting the configured install strategy, see
+    # if we're overriding it.
+    if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then
+        {% if install_paths| selectattr("kind", "equalto", "CargoHome") %}
```
Gankra, line 296:
```suggestion
        {%- if install_paths| selectattr("kind", "equalto", "CargoHome") %}
```

(you're missing these in a ton of places, lots of extra newlines)

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -47,6 +50,47 @@ param (
...
+{% for install_path in install_paths %}
+if (-Not $install_path) {
```
Gankra, line 60: ...we should really look into if we're supposed to be wrapping these in quotes like we do in shell...

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -47,6 +50,47 @@ param (
...
+{% if install_path.kind == "CargoHome" %}
+  # This strategy can't fail; always assign it
+  $install_path = @{
+    "strategy" = "CargoHome"
+  }
```
Gankra, line 65: I'd still prefer if the powershell impl had the same kind of "no dead code" approach as the shell approach. Especially since you're having to shove a bunch of random keys in here anyway, seems like you're making more logic without unifying anything?

Gankra (changes requested):
The shell is looking great, just some minor nits.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: I was wondering the same thing. It's a bit inconsistent about that atm

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: This was on purpose: setting `{%- if ` here staples it to the end of the previous line with no newline.

```sh
if [ -n "${CARGO_DIST_FORCE_INSTALL_DIR:-}" ]; then_install_dir="$CARGO_DIST_FORCE_INSTALL_DIR"
```

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Oh god right this is probably completely dependent on which `echo` is being used. I'm sure this isn't portable.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Sort of. This is slightly more restricted than the `sed` approach, so any changes we make here would change the behaviour of the installer going forward. My goal here was to minimize side changes like this other than the big one we're changing on purpose.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: It sure will! `sed: first RE may not be empty`

mistydemeo: That's why I originally still had the more explicit `$HOME` check here. I should restore it.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: This isn't trying to catch it being unset from the above case. We're generating an `if []` for each of the `for install_path in install_paths` cases, so it's letting item 2 in the array check if it was assigned in case 1, and so on.

cargo-dist/templates/installer/installer.sh.j2
Gankra: It shouldn't..? It would if you did `{%- ... -%}`, but the asymmetric version should include the newline inside the if.

cargo-dist/templates/installer/installer.sh.j2
Gankra: yes, i'm just saying we can omit the else because every case has this if-guard, making the whole thing feel a lot more natural (i applied this change)

cargo-dist/templates/installer/installer.sh.j2
Gankra: moved this up into the main comment explaining early/late

cargo-dist/templates/installer/installer.sh.j2
Gankra: punted on this. it can be done but it's wonkier

cargo-dist/templates/installer/installer.sh.j2
Gankra: Ah i see what you mean. I at least simplified it a bit.

cargo-dist/templates/installer/installer.sh.j2
Gankra: I believe the issue was the "else" here (fixed).

Gankra (approved):

PR #958, opened 2024-04-23
feat: add fish support to shell installers

mistydemeo:
This adds support for fish as a shell. Like with sh, I took inspiration from rustup's implementation. This broadly leans on the existing shell support, but has a few distinct features.

* Instead of writing to an existing configuration file like `~/.bashrc`, we write to a file in fish's `conf.d` path. This is a directory which is autoloaded on shell boot, which means we can create an app-specific path and can avoid having to edit the user's own configuration file.
* Unique to fish, since `conf.d` may not exist at the time we want to interact with it, I've added an `mkdir -p` call.
* I've used fish's [`fish_add_path`](https://fishshell.com/docs/current/cmds/fish_add_path.html) helper, which encapsulates the logic of "only prepend to the path if it's not already on the `PATH`" without us having to write it ourselves. (Note that `fish_add_path` was introduced in fish 3.2.0, which was released in 2020. Since that version is available in the current LTS release of Ubuntu and Debian stable, I believe it's old enough to be safe for us to use it.)
* I updated our existing shell writing code to make it aware of what shell we're writing so I could branch which form of envfile to write, and whether to output `source` or `.` statements. (fish strongly prefers `source` and has stated `.` may be removed in the future.)

I gave this a test locally; it appears to be working fine, and, like with the sh side, doesn't overwrite cargo's configuration for `CargoHome` installs.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -372,6 +372,12 @@ install() {
         err "could not find a valid path to install to!"
     fi
 
+    # Identical to the sh version, just with a .fish file extension
+    # We place it down here to wait until it's been assigned in every
+    # path.
+    _fish_env_script_path="${_env_script_path}.fish"
+    _fish_env_script_path_expr="${_env_script_path_expr}.fish"
+
     # Replace the temporary cargo home with the calculated one
     RECEIPT=$(echo "$RECEIPT" | sed "s,AXO_INSTALL_PREFIX,$_install_dir,")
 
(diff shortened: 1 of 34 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -530,6 +553,18 @@ esac
...
 
+write_env_script_fish() {
+    # write this env script to the given path (this cat/EOF stuff is a "heredoc" string)
+    # Note that `fish_add_path` avoids re-adding a path that already
+    # exists in the PATH, so we don't need the complex logic from above.
+    local _install_dir_expr="$1"
+    local _env_script_path="$2"
+    ensure cat <<EOF > "$_env_script_path"
+fish_add_path --path "$_install_dir_expr"
+esac
```
Gankra, line 564: ...what's up with this trailing `esac`? copy-paste error?

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -499,10 +516,16 @@ add_install_dir_to_path() {
             # If the script now exists, add the line to source it to the rcfile
             # (This will also create the rcfile if it doesn't exist)
             if [ -f "$_env_script_path" ]; then
-                say_verbose "adding $_robust_line to $_target"
+                local _line
+                if [ "$_shell" = "fish" ]; then
+                    _line="$_pretty_line"
+                else
+                    _line="$_robust_line"
+                fi
```
Gankra, line 524: Worth adding a comment here noting that fish has deprecated `. file` and explicitly wants you to use `source file`.

mistydemeo: Yeah, good call, I'll add a comment

PR #962, opened 2024-04-25
chore: complete DistError migration

Gankra:
Noticed we were close and just decided to finish it off.


ashleygwilliams (approved):
Unsure if we want to just get this merged, but I generally try to avoid transparent errors unless I really know the message is good- might be worth quickly wrapping those now

Gankra:
Yeah I agree we overuse transparent here, but I think that's gonna need to be an ongoing project.

PR #968, opened 2024-04-26
feat: add release-branch config

Gankra:
This lets you specify pushing to a branch at all should trigger a release

* [x] fill in TODOs in template
* [x] write docs
* [x] come up with some way for the "plan" step to decide on a tag/version that all other steps have to agree on (in particular with a `-alpha+{timestamp}` suffix).
  * possibly this extra detail is an extra config?
  * [x] come up with a way to force-rewrite the Cargo.tomls to use that version 
    * refuse to support non-unified-version workspaces..? (cargo-dist's CI refuses to acknowledge the idea of multiple tags coming out of a single workflow run) 
* [x] debounce the workflow??? (using the gitops queue thing we do for our own production branches)
  * ostensibly users of this don't want one tag/release per commit, and just want to release "the latest"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -390,7 +390,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-dist"
-version = "0.15.0-prerelease.1"
+version = "0.15.0-prerelease.2"
 dependencies = [
  "axoasset",
  "axocli",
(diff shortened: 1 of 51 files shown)


Gankra:
Arguably instead of two configs for `dispatch_releases` and `release_branch` we should just have an enum similar to `install-path`, but I'd rather get the impl working before thinking about that.

mistydemeo (approved):
A few comments, but looks good in general

PR #974, opened 2024-04-29
chore: rewrite npm installer

Gankra:


cargo-dist/src/backend/templates.rs
```diff
@@ -189,10 +200,31 @@ impl Templates {
...
         val: &impl Serialize,
+    ) -> DistResult<String> {
+        if file.env == ENV_NONE {
+            self.render_raw_file_to_clean_string(file)
+        } else {
+            self.render_templated_file_to_clean_string(file, val)
+        }
+    }
+
+    /// Render a non-jinja template file to a string, cleaning all newlines to be unix-y
```
mistydemeo, line 216: "Render" kind of implies actually inlining values into the file, might be worth having the comment clarify it's just copying it and redoing newlines?

cargo-dist/templates/installer/npm/binary.js
```diff
@@ -0,0 +1,134 @@
...
+const { configureProxy } = require("axios-proxy-builder");
+
+const error = (msg) => {
+  console.error(msg);
+  process.exit(1);
+};
+
+const {
+  name,
+  artifactDownloadUrl,
```
mistydemeo, line 14: Right, I see now how this made its way into `package.json`. May be worth a note where we add it that this isn't a standard field, but one we've added for ourselves.

mistydemeo (approved):
Looks good, just a few comments.

Ran locally and confirmed working:

```sh
$ npm exec axolotlsay hi
         +----+
         | hi |
         +----+
        /
≽(◕ ᴗ ◕)≼
```

Issue #457, opened 2023-09-27
bump curl-sh exprs to tls 1.3?


Gankra:
We currently emit curl-sh exprs of the form:

```
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.3.0/cargo-dist-installer.sh | sh
```

This was inherited from https://rustup.rs/ although we added the extra `-L` (iirc required to chase redirects from github).

It has been suggested we should update this to tlsv1.3 -- I'm not really a tls/interop expert, so feedback welcome.

Nilstrieb:
I tried it on debian oldoldstable (which has a too old glibc to even run cargo-dist 0.3.0 :D) and it works just fine. curl added TLS 1.3 all the way back in 2018 before it was standardized. https://daniel.haxx.se/blog/2018/03/27/play-tls-1-3-with-curl/

bengaywins:
This seems pointless to bother with. I created #978 to remove this entirely. TLS should be negotiated for the highest supported version first and downgraded when it's unsupported.

bjorn3:
`--tlsv1.2` indicates that TLSv1.2 or higher must be used. Changing this to `--tlsv1.3` will force TLSv1.3 or higher to be used, which Github is known to support. If for some reason TLSv1.2 or lower is negotiated that either means that curl didn't support TLSv1.3 (if that is the case your curl version is likely old enough to be vulnerable to known security issues anyway, so erroring out in this case is completely reasonable IMO) or there is a MITM attacker trying to downgrade to a lower TLS version to exploit a cryptographic issue, in which case you definitively want to error out. Now TLSv1.3 should have protection against downgrade attacks when the client is TLSv1.3 capable and downgrading to anything lower than TLSv1.2 shouldn't be possible as Github rejects connections with TLSv1.1 and lower. As such `--tlsv1.2` probably doesn't provide any benefit for connecting to Github or any other site which requires TLSv1.2 or higher. `--tlsv1.3` may potentially provide marginal benefit against future attacks, though probably not a whole lot given the downgrade protection mechanism of TLSv1.3.

bengaywins:
If TLS 1.3 support isn't compiled into your cURL binary, that would then fail. This should be up to the endpoint to determine which TLS version to use/support. cURL (and by standard use case, the ssl/tls library implementation) provides for a way to downgrade if necessary.

Gankra:
[...]

PR #978, opened 2024-04-30
Update release.yml to remove extraneous command line options for cURL

bengaywins:
Remove protocol switch because this is already specified in the URL with HTTPS.

Remove TLS version specificity, to allow for cURL to determine which TLS version to use based on compile-time support.

There's no real reason to have either of these in here unless you're solving for a problem that exists. To my knowledge, there is no problems with cURL determining the highest supported TLS version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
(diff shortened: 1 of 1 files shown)


Gankra:
I am going to reject this PR per discussion in #457 

* Specifying `--proto '=https'` is understood to prevent redirects from downgrading from https to http
* Specifying `--tls=1.2` similar sets a minimum threshold, preventing security downgrades

It's *ugly* for sure, but it does serve a purpose (happy to be informed I'm wrong). That said we *might* consider a flag to say "I don't care about this attack, give me the prettier thing", although that seems like an unfortunate tradeoff to give users.

PR #984, opened 2024-05-01
factor platform code into shared unit

Gankra:
This is a significantly reduced scope implementation of #793 that takes all the rich info and shoves it back into the shape of ExecutableZipFragment so all the installers can remain the same. However in doing so it introduces much better support for aliases and fallbacks, and centralizes the logic.

As a result several missing aliases are added, and fallback is much more maintainable.


Gankra:
This is draft because:

* [ ] it's gonna get (minorly) merge-conflict by my npm refactor
* [ ] it temporarily deletes the UpdaterFragment logic, because that predated the original branch and I didn't want to get bogged down.

cargo-dist/src/backend/installer/mod.rs
```diff
@@ -66,8 +66,8 @@ pub struct InstallerInfo {
 pub struct ExecutableZipFragment {
     /// The id of the artifact
     pub id: String,
-    /// The targets the artifact supports
-    pub target_triples: Vec<TargetTriple>,
+    /// The target the artifact supports
+    pub target_triples: TargetTriple,
```
Gankra, line 70: The fact that everything ignored that this can be an array sucked, it's nice that the new impl smooths this out. The value should be renamed but my rust-analyzer was bugging out so I opted not to do this.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -260,8 +255,8 @@ json_aliases() {
     local _arch="$1"
 
     case "$_arch" in {% for artifact in artifacts %}
-    "{{ artifact.target_triples[0] }}")
-        echo '{{ aliases[artifact.target_triples[0]] | tojson }}'
+    "{{ artifact.target_triple }}")
+        echo '{{ aliases[artifact.target_triple] | tojson }}'
```
mistydemeo, line 259: Just a note that this needs to be rebased on #986 after that's merged.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,414 @@
...
+                TARGET_X64_MAC.to_owned(),
+                PlatformEntry {
+                    quality: SupportQuality::ImperfectNative,
+                    runtime_conditions: archive.native_runtime_conditions.clone(),
+                    archive_idx,
+                },
+            ));
+        }
+
+        // If this is x64 macos, say it can support arm64 macos using Rosetta2
```
mistydemeo, line 344: Minor nitpick for clarity: I initially read this as meaning that x86_64 Macs could run arm64, rather than the other way around.

mistydemeo (approved):
Looking good, and I really appreciate the way this pushes so much of this logic into a central place. Just a few comments.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,351 @@
...
+                    runtime_conditions: archive.native_runtime_conditions.clone(),
+                    archive_idx,
+                },
+            ));
+        }
+
+        // x86_32 macos binaries ran fine on x86_64, but it's Imperfect compared to actual x86_64 binaries
+        // ...UP UNTIL macOS Catalina (macOS 10.15, October 2019)!
+        //
+        // FIXME: add some condition for "check the macos version" if we care about this!
```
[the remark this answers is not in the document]
Gankra: Oh good catch. When I first wrote this the idea was installers would reject any RuntimeCondition they don't know about, so it was safe to toss in "random" stuff like this but... for now we're ignoring the RuntimeConditions and making those checks Future Work.

cargo-dist/src/platform.rs
```diff
@@ -0,0 +1,351 @@
...
+///
+/// On Linux it's preferred to dynamically link libc *but* because the One True ABI
+/// is actually the Linux kernel syscall interface, you *can* theoretically statically
+/// link libc. This comes with various tradeoffs but the big selling point is that the
+/// Linux kernel is a much more slowly moving target, so you can build a binary
+/// that's portable across way more systems by statically linking libc. As such,
+/// for any archive claiming to provide a static libc linux build, we can mark this
+/// archive as providing support for any linux distro (for that architecture)
+///
+/// Currently rust takes "linux-musl" to mean "statically linked musl"a
```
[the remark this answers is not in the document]
Gankra: Thanks

PR #986, opened 2024-05-01
chore: rename aliases to bin-aliases

mistydemeo:


Gankra (changes requested):
I think there's a few places where the string "aliases" exists and needs to be changed to "bin-aliases".

mistydemeo:
Pushed additional renames, including the field in the install receipt.

Gankra (approved):

Gankra:
oh nooo.

what foul merge conflicts hath i wrought

PR #1026, opened 2024-05-08
feat: allow specifying external commit from submodule

mistydemeo:
This is the companion to [`github-releases-repo`](https://opensource.axo.dev/cargo-dist/book/reference/config.html#github-releases-repo); when targeting creating a commit in an external repo, where that repo is also checked out as a submodule of the repository in which the release is being run, this lets you consult the latest commit of that submodule and use that commit as the tag instead of whatever happens to be the latest commit in that repository.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -144,6 +144,10 @@ pub struct GithubCiInfo {
     /// What kind of job to run on pull request
     #[serde(skip_serializing_if = "Option::is_none")]
     pub pr_run_mode: Option<PrRunMode>,
+
+    /// A specific commit to tag in an external repository
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub external_repo_commit: Option<String>,
 }
 
 /// Github CI Matrix
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -269,6 +269,8 @@ pub struct DistGraph {
     pub install_updater: bool,
     /// Publish GitHub Releases to this other repo
     pub github_releases_repo: Option<config::GithubRepoPair>,
+    /// Read the commit to be tagged from the submodule at this path
+    pub github_releases_submodule_path: Option<String>,
 }
 
 /// Info about artifacts should be hosted
@@ -850,6 +852,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             all_features,
             create_release,
             github_releases_repo,
+            github_releases_submodule_path,
             pr_run_mode: _,
             allow_dirty,
             msvc_crt_static,
@@ -875,6 +878,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         let ssldotcom_windows_sign = ssldotcom_windows_sign.clone();
         let tag_namespace = tag_namespace.clone();
         let github_releases_repo = github_releases_repo.clone();
+        let github_releases_submodule_path = github_releases_submodule_path.clone();
 
         let mut packages_with_mismatched_features = vec![];
         // Compute/merge package configs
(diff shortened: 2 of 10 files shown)


cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -577,6 +577,7 @@ jobs:
           owner: {{{ github_releases_repo.owner }}}
           repo: {{{ github_releases_repo.repo }}}
           token: ${{ secrets.GH_RELEASES_TOKEN }}
+          commit: ${{ fromJson(needs.host.outputs.val).ci.github.external_repo_commit }}
```
Gankra, line 580: This needs to be wrapped in some kind of if, since it should only be done if the submodule config is set

cargo-dist-schema/src/lib.rs
```diff
@@ -144,6 +144,9 @@ pub struct GithubCiInfo {
     /// What kind of job to run on pull request
     #[serde(skip_serializing_if = "Option::is_none")]
     pub pr_run_mode: Option<PrRunMode>,
+
+    /// A specific commit to tag in an external repository
+    pub external_repo_commit: Option<String>,
```
Gankra, line 149:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub external_repo_commit: Option<String>,
```

cargo-dist/src/tasks.rs
```diff
@@ -2795,3 +2810,30 @@ impl InstallReceipt {
         })
     }
 }
+
+// Determines the *cached* HEAD for a submodule within the workspace.
+// Note that any unstaged commits, and any local changes to commit
+// history that aren't reflected by the submodule commit history,
+// won't be reflected here.
+fn submodule_head(submodule_path: &Utf8PathBuf) -> DistResult<String> {
+    let output = Cmd::new("git", "fetch expected latest commit for a submodule")
```
Gankra, line 2819: This isn't the format of the error string, it's an explanation of what we were trying to do (gets prefixed with "failed to"). So it should be like "get submodule commit".

cargo-dist/src/tasks.rs
```diff
@@ -2795,3 +2810,30 @@ impl InstallReceipt {
...
+        .arg("submodule")
+        .arg("status")
+        .arg("--cached")
+        .arg(submodule_path)
+        .output()
+        .map_err(|_| DistError::GitSubmoduleCommitError {
+            path: submodule_path.to_string(),
```
Gankra, line 2826: Should this wrap the axoprocess error? Oh I see, the other path doesn't have that.

cargo-dist/src/tasks.rs
```diff
@@ -2795,3 +2810,30 @@ impl InstallReceipt {
...
+        })?;
+
+    let line = String::from_utf8_lossy(&output.stdout);
+    // Format: one status character, commit, a space, repo name
+    let line = line.trim_start_matches([' ', '-', '+']);
```
Gankra, line 2831: neat!

Gankra (approved):
impl seems basically solid, some fixes needed

cargo-dist/src/tasks.rs
mistydemeo: Oh yeah, sorry, "expected" wasn't meant as an error message - it was because this fetches the cached/configured (eg "expected") commit of the submodule rather than what's currently there on disk

cargo-dist/templates/ci/github/release.yml.j2
mistydemeo: I believe if it's passed as `null`, it's functionally the same as if it weren't passed at all - that's why I didn't do any kind of "if" here. https://github.com/ncipollo/release-action/blob/1cbdc80532ba923d3d81ebd1cece305c437b8d66/src/Inputs.ts#L110-L116

Gankra: Will that still work right if we make `external_repo_commit` skip_serializing_if?

Gankra: (probably?)

mistydemeo: After we did a little looking, we've settled on that this should work as-is.

cargo-dist/src/tasks.rs
mistydemeo: Switched to the technical git term, "cached".

PR #1031, opened 2024-05-09
fix: bin_aliases should come from packages

mistydemeo:
It's possible for packages to have their own specific bin_aliases. We should always read from the package metadata version and skip the workspace version. Since we clone the workspace value into the package if the package has no version to compare to, that's safe.


cargo-dist/src/tasks.rs
```diff
@@ -1199,6 +1196,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             formula,
             system_dependencies,
             platform_support,
+            bin_aliases: BinaryAliases(package_config.bin_aliases.clone().unwrap_or_default()),
```
Gankra, line 1199: (style nit) can you compute this earlier up like the rest of the configs?

Gankra (approved):

cargo-dist/src/tasks.rs
mistydemeo: Done

cargo-dist/src/tasks.rs
```diff
@@ -1729,7 +1729,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             warn!("skipping shell installer: not building any supported platforms (use --artifacts=global)");
             return;
         };
-        let bin_aliases = self.inner.bin_aliases.for_targets(&target_triples);
```
ashleygwilliams, line 1732: where does this change come from?

mistydemeo: We used to look up the workspace value from the DistGraph (`self.inner`), but now we're reading it from `release`.

PR #1050, opened 2024-05-15
fix: repair and cleanup npm binary caching

Gankra:
* this.binaryPath no longer exists, which was causing us to refetch every time
* Package.run already calls Package.install as a subroutine, there's no need to call it again and duplicate the work (potentially we could also remove that, but hey integrity checks are nice?)
* The additional logic in run to select a binary to call was an artifact of development, and never actually ran (trivially provable by nothing ever omitting the argument) (this logic is performed by npx itself because we populated package.json's 'bins')

### DIFF
--- a/cargo-dist/templates/installer/npm/binary-install.js
+++ b/cargo-dist/templates/installer/npm/binary-install.js
@@ -57,7 +57,14 @@ class Package {
   }
 
   exists() {
-    return existsSync(this.binaryPath);
+    for (const binaryName in this.binaries) {
+      const binRelPath = this.binaries[binaryName];
+      const binPath = join(this.installDirectory, binRelPath);
+      if (!existsSync(binPath)) {
+        return false;
+      }
+    }
+    return true;
   }
 
   install(fetchOptions, suppressLogs = false) {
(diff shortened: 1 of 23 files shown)


cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -57,7 +57,14 @@ class Package {
   }
 
   exists() {
-    return existsSync(this.binaryPath);
+    for (const binaryName in this.binaries) {
+      const binRelPath = this.binaries[binaryName];
+      const binPath = join(this.installDirectory, binRelPath);
+      if (!existsSync(this.binaryPath)) {
```
mistydemeo, line 63: This wants to be `binPath`, right?

mistydemeo (changes requested):

cargo-dist/templates/installer/npm/run.js.j2
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
 
-const { run, install: maybeInstall } = require("./binary");
-maybeInstall(true).then(() => run({{ bin }}));
-
+const { run } = require("./binary");
+run({{ bin }}));
```
mistydemeo, line 4: Stray paren here:

```suggestion
run({{ bin }});
```

cargo-dist/templates/installer/npm/binary-install.js
Gankra: LMAO FUCK

mistydemeo (approved):

PR #1052, opened 2024-05-15
feat: fetch updater without GitHub API call

mistydemeo:
Instead of querying releases and using that to see if an asset is available for our architecture, just manually construct the URL and use a HEAD request to see if it exists or not.

This should improve reliability when checking for new updater releases as a part of the local build process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,12 +121,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
 
 [[package]]
-name = "arc-swap"
-version = "1.7.1"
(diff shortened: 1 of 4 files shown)


mistydemeo:
Confirmed this works: https://github.com/mistydemeo/cargodisttest/actions/runs/9102514810

ashleygwilliams (approved):
so i love this strategy. i have known APIs to be weird about HEAD requests... so we should keep an eye on it. if people keep good hygiene it should work well and fine, but people do weird stuff lol. e.g. https://community.cloudflare.com/t/when-cache-everything-head-requests-will-be-turned-into-get-requests/363090

mistydemeo:
We can definitely keep an eye on GitHub's behaviour here and hope they don't accidentally make a weird mistake 😅

PR #1055, opened 2024-05-16
feat: allow configuring global runner

mistydemeo:
We already parameterized the "global task" runner, but hardcoded it to always use a specific GitHub Linux runner. This updates it to now use the "global" key in the pre-existing `github-custom-runners` table, if it's set. It also updates `plan`, the one non-parameterized task, to use this runner too.

Updates the custom runner test to exercise this new feature.

Note that it's worth thinking about if there are any considerations on changing `plan`. I'm actually not clear why this one task was hardcoded rather than using the same value as the other calls.

### DIFF
--- a/cargo-dist/tests/integration-tests.rs
+++ b/cargo-dist/tests/integration-tests.rs
@@ -747,6 +747,7 @@ targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "x86_64-unkn
 ci = ["github"]
 
 [workspace.metadata.dist.github-custom-runners]
+global = "buildjet-8vcpu-ubuntu-2204"
 x86_64-unknown-linux-gnu = "buildjet-8vcpu-ubuntu-2204"
 x86_64-unknown-linux-musl = "buildjet-8vcpu-ubuntu-2204"
 aarch64-unknown-linux-gnu = "buildjet-8vcpu-ubuntu-2204-arm"
--- a/cargo-dist/tests/snapshots/axolotlsay_user_plan_job.snap
+++ b/cargo-dist/tests/snapshots/axolotlsay_user_plan_job.snap
@@ -3262,7 +3262,7 @@ on:
 jobs:
   # Run 'cargo dist plan' (or host) to determine what tasks we need to do
   plan:
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-20.04"
     outputs:
       val: ${{ steps.plan.outputs.manifest }}
       tag: ${{ !github.event.pull_request && github.ref_name || '' }}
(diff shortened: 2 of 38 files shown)


mistydemeo:
> I'm actually not clear why this one task was hardcoded rather than using the same value as the other calls.

Answering my own question - it looks like we parameterized the global task when splitting out local/global tasks, so not updating `plan` was just an artifact of the fact it's not what we were focused on.

https://github.com/axodotdev/cargo-dist/commit/bdc3da7ac10764d0ebf91cac237867b177fa61ed#diff-a9384fa92709b8a9451e42ca738c5ab944c285d7185af81458754c0f09139a2a

ashleygwilliams:
@mistydemeo that follows with my memory as well, i think we are safe with this PR (it's a strict improvement imo!)

cargo-dist/tests/integration-tests.rs
```diff
@@ -747,6 +747,7 @@ targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "x86_64-unkn
 ci = ["github"]
 
 [workspace.metadata.dist.github-custom-runners]
+global = "buildjet-8vcpu-ubuntu-2204"
```
ashleygwilliams, line 750: so "global" is the obvious term for this but i wonder if it's a bit too inside baseball to expose to end users? i am ok with keeping it for now, my brain kinda goes to `"default"` e.g. use this runner when the runner doesn't really matter, but im' not sure if that's actually better or not

ashleygwilliams: also: do we want to run our tests on the private runner? it's not a lot of money, but i feel like it might be good to not do that if we don't have to

cargo-dist/tests/snapshots/axolotlsay_user_plan_job.snap
```diff
@@ -3262,7 +3262,7 @@ on:
 jobs:
   # Run 'cargo dist plan' (or host) to determine what tasks we need to do
   plan:
-    runs-on: ubuntu-latest
```
ashleygwilliams, line 3265: as per previous comment, i think it's better if these remain on ubuntu-latest as i think that is the most common platform users will use

ashleygwilliams (changes requested):
i think this looks good modulo a few small things, the only thing i think that's left is docs!

cargo-dist/tests/integration-tests.rs
mistydemeo: Happy to defer to your opinions on the public-facing name for this.

This test doesn't actually run in CI; I'm just testing with this that it generates the expected CI config.

ashleygwilliams: i think we should probably ship with global for now, but when we do the config 1.0 project consider the way we talk about jobs in general!

mistydemeo:
Pushed a commit with docs.

ashleygwilliams (approved):
🚀

PR #1062, opened 2024-05-16
fix: really avoid native-tls

mistydemeo:
I'd completely forgotten that enabling rustls doesn't do anything if you don't also `default-features - false`.


ashleygwilliams (approved):
UGH i'm sorry i should have caught this!!!

PR #1066, opened 2024-05-17
fix: Add `.tar.zst` suffix for Zstd

### LINKED ISSUE #1064: Filename extension for Zstd should be `.zst`
According to [RFC 8478](https://datatracker.ietf.org/doc/html/rfc8478#autoid-51), the filename extension for Zstd should be `.zst` instead of `.zstd`.

sorairolake:
Closes #1064

### DIFF
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -941,7 +941,7 @@ pub enum CompressionImpl {
     Gzip,
     /// `.xz`
     Xzip,
-    /// `.zstd`
+    /// `.zst`
     Zstd,
 }
 impl ZipStyle {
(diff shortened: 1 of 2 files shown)


cargo-dist/src/config.rs
```diff
@@ -980,9 +980,9 @@ impl<'de> Deserialize<'de> for ZipStyle {
             ".zip" => Ok(ZipStyle::Zip),
             ".tar.gz" => Ok(ZipStyle::Tar(CompressionImpl::Gzip)),
             ".tar.xz" => Ok(ZipStyle::Tar(CompressionImpl::Xzip)),
-            ".tar.zstd" => Ok(ZipStyle::Tar(CompressionImpl::Zstd)),
```
ashleygwilliams, line 983: for backwards compatibility, we'll probably want to keep this around for at least a few releases. would you mind updating to accept both `zstd` and `zst`?

ashleygwilliams (approved):

PR #1080, opened 2024-05-22
feat: validate checksums when vendoring actions

### LINKED ISSUE #1077: Actions vendoring: check hashes for tarballs
In https://github.com/axodotdev/cargo-dist/pull/1074, we added support for vendoring actions. Currently, we fetch the action source as tarballs from GitHub. We should validate checksums for those tarballs before unpacking, and ensure this is in place before the 0.15.0 release.

mistydemeo:
Fixes #1077.

### DIFF
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -375,6 +375,18 @@ pub enum DistError {
         /// The input
         pair: String,
     },
+
+    /// Checksum mismatch when fetching vendored actions
+    #[error("Checksum mismatch when fetching vendored GitHub action {repo}")]
+    #[diagnostic(help("Expected {expected} but got {actual}"))]
+    VendoredActionHashMismatch {
+        /// The checksum we expected
+        expected: String,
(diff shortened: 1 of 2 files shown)


cargo-dist/src/errors.rs
```diff
@@ -375,6 +375,18 @@ pub enum DistError {
...
+    /// Checksum mismatch when fetching vendored actions
+    #[error("Checksum mismatch when fetching vendored GitHub action {action}")]
+    #[diagnostic(help("Expected {actual} but got {expected}"))]
+    VendoredActionHashMismatch {
+        /// The checksum we expected
+        expected: String,
+        /// The actual checksum
+        actual: String,
+        /// The action we were fetching (owner/repo format)
+        action: String,
```
ashleygwilliams, line 388: we might call this owner_repo (or just repo as it's called above?) because action was definitely confusing to me.

ashleygwilliams (approved):
a nit on var names but otherwise lets get this in

PR #1082, opened 2024-05-22
docs: add vendor-workflow-deps

mistydemeo:


CHANGELOG.md
```diff
@@ -19,6 +19,16 @@ cargo-dist has traditionally parsed the version number of new releases and used
 
 * impl @mistydemeo [feat: allow always marking releases as stable](https://github.com/axodotdev/cargo-dist/pull/1054)
 
+## Vendoring external GitHub actions
+
+This feature allows vendoring (almost) all external GitHub actions called by the release workflow. When enabled, specific known versions of all external actions (except `actions/checkout`) are copied into your repository as a part of `cargo dist init`. We don't expect most users to need this feature, but it's provided for users with specific security needs.
+
+To enable this user, set the [`vendor-workflow-deps`](https://opensource.axo.dev/cargo-dist/book/reference/config.html#vendor-workflow-deps) option in your `Cargo.toml` to `true`.
```
ashleygwilliams, line 26: i think the word `user` here is errant maybe?

CHANGELOG.md
```diff
@@ -19,6 +19,16 @@ cargo-dist has traditionally parsed the version number of new releases and used
 
 * impl @mistydemeo [feat: allow always marking releases as stable](https://github.com/axodotdev/cargo-dist/pull/1054)
 
+## Vendoring external GitHub actions
+
+This feature allows vendoring (almost) all external GitHub actions called by the release workflow. When enabled, specific known versions of all external actions (except `actions/checkout`) are copied into your repository as a part of `cargo dist init`. We don't expect most users to need this feature, but it's provided for users with specific security needs.
```
ashleygwilliams, line 24: i wonder if there is a way to say this more positively- like, for users that want stronger guarantees, or something. i think saying "we don't expect more users to need this" feels off here

book/src/reference/config.md
```diff
@@ -791,6 +791,17 @@ The supported choices are:
 By default all runs of `cargo-dist` will be trying to handle all platforms specified here at once. If you specify `--target=...` on the CLI this will focus the run to only those platforms. As discussed in [concepts][], this cannot be used to specify platforms that are not listed in `metadata.dist`, to ensure different runs agree on the maximum set of platforms.
 
 
+### vendor-workflow-deps
+
+> since 0.15.0
+
+Example: `vendor-workflow-deps = true`
+
+When enabled, `cargo dist init` will copy the complete source of (almost) all external GitHub actions into your git repository at `.github/actions` and configure your release workflow to use the local copies instead of the external copies. For most users, this is unnecessary; we expect this to only be of interest to users with specific restricted security contexts who want to ensure they're using specific known versions of actions without fetching them from third party repositories.
```
ashleygwilliams, line 800: same as above, can we wordsmith this to be more positive. i think a lot of our users are security conscious and so i think we should frame that as more normal/good

ashleygwilliams (changes requested):

CHANGELOG.md
mistydemeo: Good catch, I'll swap to `setting`.

CHANGELOG.md
mistydemeo: Yeah, I was struggling a bit with how to word this. The thing is, since it adds a million lines (not an exaggeration!) of code to the user's repo, I think we want to strike a balance here where we don't come across as *encouraging* it to users who'd be surprised by the weight of it. I think it's more just that people who 100% understand the tradeoffs can make the choice?

PR #1088, opened 2024-05-23
fix: remove over-eager warning

ashleygwilliams:


cargo-dist/src/manifest.rs
```diff
@@ -124,13 +124,6 @@ pub(crate) fn load_and_merge_manifests(
             output.ci = Some(val);
         };
 
-        // Just merge all the system-specific info
```
mistydemeo, line 127: Mind leaving this comment in, attached to the `extend` calls below?

mistydemeo (approved):
Context: this warning came out of pull request review in #848, the PR that reworked how manifests work: https://github.com/axodotdev/cargo-dist/pull/848#discussion_r1525623201

We'd noticed when writing this PR that it was possible for the same platform key to appear more than once in our array, and that, during merging, only the last would appear in the final output. We were concerned we might *accidentally* produce duplicate keys and drop data, or that it could be a sign of a more fundamental bug. However, we *also* produce such duplicate keys as a matter of course and therefore we always print this message. It's not serving a useful purpose at the moment. We should review this and decide whether to restore this warning and, if we do, only do it in a more targeted and useful way.

cargo-dist/src/manifest.rs
ashleygwilliams: oh sure, i wasn't sure if "system-specific info" applied to all 3 calls or now, i'll update that now

mistydemeo (approved):

PR #1096, opened 2024-05-25
README.md:update more links to https://axodotdev.github.io/cargo-dist

RandyMcMillan:


RandyMcMillan:
@ashleygwilliams - follow up to #1086

Gankra:
Hey there, thanks for the fix! I actually just fixed our infra so that the domains now work again, so unfortunately this isn't necessary, but I super appreciate it!

PR #1102, opened 2024-05-30
feat(installer): custom success msg

ashleygwilliams:
the snapshots had to be updated because we had mixed casing between powershell and shell and so i unified that

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -224,6 +224,7 @@ fn get_new_dist_metadata(
             rust_toolchain_version: None,
             ci: None,
             installers: None,
+            install_success_msg: None,
             tap: None,
             formula: None,
             system_dependencies: None,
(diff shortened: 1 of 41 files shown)


cargo-dist/src/init.rs
```diff
@@ -934,6 +936,13 @@ fn apply_dist_to_metadata(metadata: &mut toml_edit::Item, meta: &DistMetadata) {
         npm_package.as_deref(),
     );
 
+    apply_optional_value(
+        table,
+        "install_success_msg",
```
Gankra, line 941:
```suggestion
        "install-success-msg",
```

cargo-dist/src/tasks.rs
```diff
@@ -1153,6 +1158,10 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             .install_path
             .clone()
             .unwrap_or(vec![InstallPathStrategy::CargoHome]);
+        let install_success_msg = package_config
+            .install_success_msg
+            .clone()
+            .unwrap_or("everything's installed!".to_string());
```
Gankra, line 1164:
```suggestion
        let install_success_msg = package_config
            .install_success_msg
            .as_deref()
            .unwrap_or("everything's installed!")
            .to_owned();
```

clippy will probably complain about the unconditional to_string

Gankra (approved):

PR #1105, opened 2024-05-31
doc: cleanup changelogs and add msg config

Gankra:


CHANGELOG.md
```diff
@@ -1,48 +1,62 @@
...
+
+## Custom installer output
+
+When a shell or powershell installer runs successfully it will print out "everything's installed!", but if you want a different message that better matches your app's look-and-feel, or provides a more useful call-to-action, you can now do that with [the `install-success-msg` setting](https://opensource.axo.dev/cargo-dist/book/reference/config.html#install-success-msg).
+
+* [docs](https://opensource.axo.dev/cargo-dist/book/reference/config.html#install-success-msg)
+* impl @ashleygwilliams [feat: custom success msg](https://github.com/axodotdev/cargo-dist/pull/1102)
+
+
 ## Always treat releases as the latest stable release
```
ashleygwilliams, line 31: this is maybe a little confusing- i feel like this reads like we made this the default behavior and we definitely did not

ashleygwilliams (approved):
small suggestion on the "always latest" feature- i think we make it sound too usable... when it's really a niche thing that we probably don't want most people using

PR #1116, opened 2024-06-07
feat: generic workspaces

Gankra:
paired with https://github.com/axodotdev/axoproject/pull/136

### DIFF
--- a/cargo-dist/src/host.rs
+++ b/cargo-dist/src/host.rs
@@ -12,6 +12,7 @@ use crate::{
 use axoproject::WorkspaceInfo;
 use cargo_dist_schema::{DistManifest, Hosting};
 use gazenot::{AnnouncementKey, Gazenot};
+use tracing::warn;
 
 /// Do hosting
 pub fn do_host(cfg: &Config, host_args: HostArgs) -> DistResult<DistManifest> {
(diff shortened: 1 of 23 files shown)


cargo-dist/src/host.rs
```diff
@@ -269,16 +273,29 @@ fn announce_hosting(_dist: &DistGraph, manifest: &DistManifest, abyss: &Gazenot)
...
+        .map(|(idx, _)| *idx)
+        .collect::<Vec<_>>();
     // Either use the explicit one, or default to the CI provider's native solution
     let hosting_providers = hosting
         .clone()
         .or_else(|| Some(vec![ci.as_ref()?.first()?.native_hosting()?]))?;
-    let repo_url = workspace.web_url().unwrap_or_default()?;
+    // Check that there's a consistent repository URL, and if not, warn
+    if let Err(warning) = workspace.repository_url(Some(&package_list)) {
+        let report = miette::Report::new(warning);
```
mistydemeo, line 291: Curious why we need to construct a miette report here if we're just passing it to a debug string - better formatting?

mistydemeo (approved):

cargo-dist/src/host.rs
Gankra: warn!s are user-facing, and yeah wrapping in a Report is (and then debug formatting that report) is the ritual to get the Good miette output.

Issue #1117, opened 2024-06-08
Packages with no binaries + publish = false + dist = false are still built


arlyon:
Hi! I am setting up cargo-dist for use in my repo (https://github.com/arlyon/litehouse) but have hit an issue. I have libraries that are only intended to be bundled as wasm packages that are still included in the build when running `cargo dist build`, despite the repo only having a single binary.

I would expect from this [section in the documentation](https://opensource.axo.dev/cargo-dist/book/workspaces/workspace-guide.html#multiple-packages-in-a-workspace) that 'each Package that doesn't define binaries is wholly irrelevant and ignored' but that doesn't seem to be the case. As of right now it looks like if you have a single package that cannot be built on your target platform, the entire build fails despite it being irrelevant to the package you are actually trying to bundle.

See this build for the example: https://github.com/arlyon/litehouse/actions/runs/9428238688/job/25973328621. None of the plugins should be included but since they are, and a few of them can't build on this target, it fails. In this commit I added publish = false and dist = false since I thought that perhaps because they are cdylibs they might be included but no dice https://github.com/arlyon/litehouse/commit/001e464f231cfcfd6765bf0b865281921fef4b3a

I _think_ I am able to work around it for now, but worth bringing to your attention anyways.

Thanks for this otherwise excellent tool! Cheers

Alex

arlyon:
I _think_ we can fix this in my particular case by not doing a workspace build iff there is only one package. That seems to be a general win for everyone regardless of workspace structure. We also need a setting to be able to opt out of a workspace build if you do have packages that cannot be built.

Gankra:
I believe you want [`precise-builds = true`](https://opensource.axo.dev/cargo-dist/book/reference/config.html#precise-builds), which also documents why we default like we do.

arlyon:
Thanks for this, and your timely responses. All the best, Alex

PR #1118, opened 2024-06-08
do not apply workspace optimisation when a single package is built

arlyon:
Closes https://github.com/axodotdev/cargo-dist/issues/1117

This PR adds two commits:
- don't apply the workspace level build optimsation if there is exactly one binary to build. this should speed things up slightly 'for free' for users with a single binary who have libraries in the workspace that their binary doesn't depend on
- add a flag `force-per-package` that allows users to opt out of the workspace-level build optimisation for cases where they are building two or more binaries and still need this behaviour

I left the case where there are 0 binaries as-is, since I am not sure whether the list of binaries is guaranteed to be nonempty. If this is the case, I would add a third match: `[] => unreachable!()`

 This change resolves the above issue, where cargo-dist does not work in workspaces with unused libraries that cannot build on a given platform.

### DIFF
--- a/cargo-dist/src/build/cargo.rs
+++ b/cargo-dist/src/build/cargo.rs
@@ -98,19 +98,38 @@ impl<'a> DistGraphBuilder<'a> {
                     }));
                 }
             } else {
-                // If we think a workspace build is possible, every binary agrees on the features, so take an arbitrary one
-                let features = binaries
-                    .first()
-                    .map(|&idx| self.binary(idx).features.clone())
-                    .unwrap_or_default();
-                builds.push(BuildStep::Cargo(CargoBuildStep {
-                    target_triple: target.clone(),
-                    package: CargoTargetPackages::Workspace,
-                    features,
-                    rustflags,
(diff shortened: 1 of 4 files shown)


arlyon:
Also (and for #1119) please let me know what I need to do regarding docs.

arlyon:
I managed to run a build with the two PRs here if you'd like to test them https://github.com/arlyon/cargo-dist/releases/tag/v0.16.5

Gankra:
Does https://github.com/axodotdev/cargo-dist/issues/1117#issuecomment-2158442761 make this unnecessary for you? (The linked docs go into detail about why we in fact want to default workspaces even when shipping a single package, although it's definitely not a clear cut decision).

arlyon:
Ah yeah. Apologies, I pored through the docs for some feature like this but clearly not well enough. Kinda funny that the relevant code was immediately above where I inserted the feature ... sometimes you gotta just go for a walk 😂

I think the behaviour of defaulting to a precise build by default when building a single binary is helpful but that is down to opinion. Thanks!

PR #1119, opened 2024-06-08
add support for setting the cache to use

arlyon:
swatinem/rust-cache supports two caching backends; github actions and buildjet.

This just plumbs that setting through defaulting to leaving the field empty.

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -72,6 +72,9 @@ pub struct GithubCiInfo {
     pub hosting_providers: Vec<HostingStyle>,
     /// whether to prefix release.yml and the tag pattern
     pub tag_namespace: Option<String>,
+    /// Cache provider
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cache_provider: Option<String>,
 }
 
 impl GithubCiInfo {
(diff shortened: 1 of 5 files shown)


Gankra:
This looks useful, could you elaborate on the usecase for toggling this setting? Are you also using buildjet runners? (I'm wondering if we can maybe just notice a custom runner with "buildjet" in the name and auto-flip the setting).

arlyon:
Yeah, essentially. I wanted to release on arm, noticed buildjet supports it, but also saw that the cache performance for non-gh runners is pretty awful (plus they are twice the size and free..)

Gankra:
If you're interested in implementing the auto-detect approach (less config is always nice), I think the relevant change is to check if this value contains a string with substring "buildjet": https://github.com/axodotdev/cargo-dist/blob/57e1712d0164aa88a757fb333d7d7b95da093fe7/cargo-dist/src/backend/ci/github.rs#L145

Gankra:
Pardon, the value of interest is github_custom_runners, although maybe you just want to look at the resulting local_runs

https://github.com/axodotdev/cargo-dist/blob/57e1712d0164aa88a757fb333d7d7b95da093fe7/cargo-dist/src/backend/ci/github.rs#L153-L158

Gankra:
(Happy to take the PR over if that's a pain, but this is a great idea so I super want to land some version of it)

Gankra:
Oh neat, just realized this value can be plumbed into GithubMatrixEntry to precisely set the cache provider per-target.

Gankra:
thanks for this again! I ended up implementing the autodetect version in #1129, should be in the next release in a day or two

PR #1128, opened 2024-06-11
fix: rework extra_artifacts to properly use paths

### LINKED ISSUE #814: extra-artifacts uses the full path as the filename
in https://github.com/axodotdev/oranda/pull/667 we had

```
[[workspace.metadata.dist.extra-artifacts]]
artifacts = ["oranda-css/dist/oranda.css"]
build = ["cargo", "run", "--", "generate-css"]
```

and this made cargo-dist want to create an artifact *named* `oranda-css/dist/oranda.css`, which is, not right.

I worked around this by making it possible to make `generate-css` write to the top-level dir, but, this is a cargo-dist bug.

Gankra:
fixes #814

### DIFF
--- a/cargo-dist/src/build/generic.rs
+++ b/cargo-dist/src/build/generic.rs
@@ -3,15 +3,15 @@
 use std::{env, process::ExitStatus};
 
 use axoprocess::Cmd;
-use camino::{Utf8Path, Utf8PathBuf};
+use camino::Utf8PathBuf;
 use cargo_dist_schema::DistManifest;
 
 use crate::{
     build::{package_id_string, BuildExpectations},
     copy_file,
     env::{calculate_cflags, calculate_ldflags, fetch_brew_env, parse_env, select_brew_env},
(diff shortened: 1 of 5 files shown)


cargo-dist/src/build/generic.rs
```diff
@@ -168,31 +199,30 @@ pub fn build_generic_target(
...
+pub fn run_extra_artifacts_build(dist: &DistGraph, build: &ExtraBuildStep) -> DistResult<()> {
     eprintln!(
         "building extra artifacts target (via {})",
-        target.build_command.join(" ")
+        build.build_command.join(" ")
     );
 
-    let result = run_build(dist_graph, &target.build_command, None)?;
-    let dest = dist_graph.dist_dir.to_owned();
+    // TODO: run this in build.working_dir (conflicts with other inflight PR)
```
Gankra, line 208: this is referring to #1116 which changes the same code and similarly wants to make commands run in a working dir

Gankra: Added a second commit that adds duped work from the PR, as the tests needed it to pass.

mistydemeo (approved):

PR #1129, opened 2024-06-12
feat: autodetect buildjet runners to use their backend for rust cache

Gankra:
This is an alternative based on #1119 which instead of introducing a coarse-grained config, adds auto-detection for each individual runner based on the name containing "buildjet". This then feeds down through the github matrix, so the cache is hosted precisely per-platform.

Big thanks to @arlyon for the initial implementation and insights!

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -272,7 +272,7 @@ impl GithubCiInfo {
     }
 }
 
-/// Get the best `cache-provider`` key to use for <https://github.com/Swatinem/rust-cache>.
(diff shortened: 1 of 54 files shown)


cargo-dist/src/backend/ci/github.rs
```diff
@@ -271,6 +272,17 @@ impl GithubCiInfo {
     }
 }
 
+/// Get the best `cache-provider`` key to use for <https://github.com/Swatinem/rust-cache>.
```
mistydemeo, line 275:
```suggestion
/// Get the best `cache-provider` key to use for <https://github.com/Swatinem/rust-cache>.
```

mistydemeo (approved):

arlyon:
Nice!

PR #1133, opened 2024-06-13
chore: rust 1.79.0

mistydemeo:
Might be safe for us to actually remove the rust-toolchain version now. We had it because we have a minimum Rust version that not all of our runners were shipping by default yet, but that's probably no longer a concern.

### DIFF
--- a/cargo-dist/tests/gallery/dist/tools.rs
+++ b/cargo-dist/tests/gallery/dist/tools.rs
@@ -1,6 +1,8 @@
 use super::*;
 use crate::gallery::command::CommandInfo;
 
+// Allowing dead code because some of these are platform-specific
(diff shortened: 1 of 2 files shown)


Gankra:
What makes you say it's no longer a concern?

mistydemeo:
It happened six months ago (#654), so it's reasonable to imagine that all of the runners have at least 1.74 by now. Even older Ubuntu images are on 1.78 now: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md

mistydemeo:
Makes sense to me, yeah. Let's keep it then.

Gankra (approved):

PR #1135, opened 2024-06-17
feat: migrate axoproject into this repo

mistydemeo:
Since cargo-dist is the main app that uses axoproject, we should merge it into this tree for the sake of rapid development. We can continue publishing it as a separate crate on crates.io, but having it in this repo will make a few things easier.


cargo-dist/src/config.rs
```diff
@@ -1530,6 +1530,7 @@ pub(crate) fn parse_metadata_table_or_manifest(
             let src = SourceFile::load_local(manifest_path)?;
             parse_generic_config(src)
         }
+        WorkspaceKind::Javascript => todo!(),
```
mistydemeo, line 1533: Need to fix this - it looks like the `path` reference is ignoring the features we asked for, so we suddenly got asked about this.

Gankra: whaaat

Gankra: oh

Gankra (approved):

PR #1145, opened 2024-06-18
feat: shared workspaces

Gankra:
This introduces a new axoproject WorkspaceGraph abstraction that exists to replace the old Workspaces approach. The purpose of this new abstraction is that it supports nested workspaces, while still allowing you to work with all the packages uniformly.

Most code largely remains the same, as it looks up properties of packages by id, or queries a property of all workspaces/packages, where essentially nothing is changed.

Also all logic to inherit config from workspaces to packages is just changed to inherit from the *root* package, as we don't want to support multiple `[workspace.metadata.dist]` entries.

compute_build functions have been changed to take a workspace, so that we try to compute N cargo builds for N cargo workspaces (discarding them if unecessary).

The URL code has once again been refactored to be more of a chain instead of redoing the same work over and over.

The `package_info` subfield of a workspace has been deprecated in favour of Asking the WorkspaceGraph, as there are different tiers of childhood now.

### DIFF
--- a/axoproject/src/lib.rs
+++ b/axoproject/src/lib.rs
@@ -12,7 +12,7 @@ use std::fmt::Display;
 use axoasset::serde_json;
 use axoasset::{AxoassetError, LocalAsset};
 use camino::{Utf8Path, Utf8PathBuf};
-use errors::{AxoprojectError, Result};
+use errors::{AxoprojectError, ProjectError, Result};
 use tracing::info;
 
 #[cfg(feature = "cargo-projects")]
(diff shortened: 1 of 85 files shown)


Gankra:
This builds and passes tests locally, but there are several outstanding TODOs, and this isn't *really* tested until we change the gallery.

axoproject/src/lib.rs
```diff
@@ -33,6 +33,186 @@ mod tests;
...
+            .map(|(idx, pkg)| (PackageIdx(idx), pkg))
+    }
+
+    /// Get all workspaces
+    pub fn all_workspace_indices(&self) -> impl Iterator<Item = WorkspaceIdx> {
+        (0..self.workspaces.len()).map(WorkspaceIdx)
+    }
+
+    /// Try to get a consensus repository_url for the given packages (or all if None given)
+    pub fn repository_url(&self, packages: Option<&[PackageIdx]>) -> Result<Option<RepositoryUrl>> {
```
mistydemeo, line 204: Wondering if it also makes sense to keep a version of this on individual workspaces. Guess we're probably not looking at that individually as often.

mistydemeo (commented):
Generally, I like this. The `WorkspaceGraph` interface is nice and clean, and I appreciate how it looks / gets used within cargo-dist. I'm mainly wondering about the couple bits of pseudo-private data and what the plan is for those.

axoproject/src/lib.rs
Gankra: Yeah you can always get that by just querying all packages under the workspace.

axoproject/src/lib.rs
```diff
@@ -119,18 +313,21 @@ pub enum WorkspaceKind {
...
 pub struct WorkspaceInfo {
-    /// The kinf of workspace this is (Rust or Javascript)
+    /// The kind of workspace this is (Rust or Javascript)
     pub kind: WorkspaceKind,
     /// The directory where build output will go (generally `target/`)
     pub target_dir: Utf8PathBuf,
     /// The root directory of the workspace (where the root Cargo.toml is)
     pub workspace_dir: Utf8PathBuf,
-    /// Computed info about the packages.
+    /// Nested workspaces (exists temp, do not use outside axoproject)
```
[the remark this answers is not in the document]
Gankra: Yeah this is basically a hack I did to get things working without melting my brain...

axoproject/src/lib.rs
```diff
@@ -229,6 +413,14 @@ impl WorkspaceInfo {
 /// specify their location -- something Cargo does but Guppy (and cargo-metadata) don't.
 #[derive(Debug)]
 pub struct PackageInfo {
+    /// The unoverrideable name of the package within its own package management system.
+    ///
+    /// This may be needed when e.g. talking to cargo about the package.
+    pub true_name: String,
```
mistydemeo, line 419: Smart plan.

mistydemeo (approved):

PR #1150, opened 2024-06-19
tests: add generic workspace tests

mistydemeo:
This adds a test running on a generic workspace, specifically covering the mixed-content workspace. Right now, we expect cargo-dist to only cover the generic half and not be able to find the other half.

I duplicated the "patch Cargo.toml" logic because the structure is a bit different and I didn't feel like genericizing the original method at this point.

This doesn't actually work yet: the output indicates that it patched `dist-workspace.toml` with the expected content, but the cargo build acts like it's working on the unpatched file. I'm sure I missed something obvious I'll see in the morning.

### DIFF
--- a/cargo-dist/tests/gallery/dist.rs
+++ b/cargo-dist/tests/gallery/dist.rs
@@ -389,15 +389,43 @@ impl DistResult {
             release_type: String,
         }
 
-        let toml = axoasset::SourceFile::load_local("Cargo.toml")
-            .unwrap()
-            .deserialize_toml_edit()
-            .unwrap();
-        let metadata = toml
-            .get("workspace")
-            .and_then(|t| t.get("metadata"))
-            .and_then(|t| t.get("dist"))
-            .unwrap();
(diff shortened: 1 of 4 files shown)


cargo-dist/tests/gallery/dist.rs
```diff
@@ -354,15 +389,43 @@ impl DistResult {
...
-            .and_then(|t| t.get("metadata"))
-            .and_then(|t| t.get("dist"))
-            .unwrap();
+        let manifest = if Utf8Path::new("Cargo.toml").exists() {
+            "Cargo.toml"
+        } else if Utf8Path::new("dist-workspace.toml").exists() {
+            "dist-workspace.toml"
+        } else if Utf8Path::new("dist.toml").exists() {
+            "dist.toml"
+        } else {
```
Gankra, line 398: I think you'll want to put Cargo.toml last here

Gankra (approved):

PR #1155, opened 2024-06-19
feat: Adds flag to optionally use system certificates when fetching `axoupdater`.

### LINKED ISSUE #1154: Cannot use system certificates when fetching updater
When fetching `axoupdater` for use in the optional "update" feature, `cargo-dist` uses `reqwest` with just the default Mozilla Web PKI certificates, and does not pick up the system certificate store. There should be an option to support using the system certs.

The reason for wanting this is basically that some corporate network security software substitutes certificates with ones managed by the organization, and those substituted certs are rooted in one or more certs in the local system certificate store. If that store isn't picked up, then any request whose certs are substituted will fail certificate validation.

mchernicoff:
Resolves #1154 See that issue for details on why this patch exists.

This adds the "native-certs" Boolean option for `cargo-dist` config settings in a project's Cargo.toml. If both this setting and "install-updater" are set to true, `cargo-dist` will fetch `axoupdater` using native system certs for TLS instead of the `reqwest` crate's default Mozilla WebPKI certs.


Gankra:
I'm going to close this as implemented in the current release, thank you so much for the preliminary work on this that became the basis of the final impl!!!

PR #1158, opened 2024-06-19
feat: clamp workspace search to root of current repo

mistydemeo:
Previously, it was possible for axoproject to reach outside the root of the current repository and find things in the parent repository. This caused problems for our tests, where we might find multiple workspaces and end up preferring cargo-dist itself over the repo we're trying to run tests on.


Gankra (approved):
For the sake of future-proofing can we make the caller have to pass in the directory to run git from?

mistydemeo:
Good call, done.

PR #1160, opened 2024-06-20
feat: add experimental tls_native_roots feature

Gankra:
The main goal of this is actually updating axoasset to 1.0.0 and
enabling the tls_native_roots feature as appropriate.

This speculatively also adds a ClientSettings type which isn't really
important but ensures all the places where we make clients are flagged
for future updates.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -13,7 +13,7 @@
 use std::io::Write;
 
 use announce::{TagMode, TagSettings};
-use axoasset::{LocalAsset, RemoteAsset};
+use axoasset::LocalAsset;
 use axoprocess::Cmd;
 use backend::{
     ci::CiInfo,
@@ -48,6 +48,7 @@ pub mod host;
 mod init;
 pub mod linkage;
 pub mod manifest;
+pub mod net;
 pub mod platform;
 pub mod sign;
 pub mod tasks;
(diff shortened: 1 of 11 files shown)


cargo-dist/src/net.rs
```diff
@@ -0,0 +1,66 @@
...
+fn create_reqwest_client(
+    ClientSettings { prefer_native_tls }: &ClientSettings,
+) -> DistResult<reqwest::Client> {
+    // TODO: add a proper error instead of calling `expect`?
+    let builder = reqwest::Client::builder()
+        .tls_built_in_webpki_certs(!prefer_native_tls);
+        // .tls_built_in_native_certs(*prefer_native_tls);
+    let client = builder
+        .build()
+        .expect("failed to intitalize http client");
```
Hoverbear, line 38:
```suggestion
        .expect("failed to initialize http client");
```

cargo-dist/src/lib.rs
```diff
@@ -209,6 +209,7 @@ pub fn fetch_updater_from_source(dist_graph: &DistGraph, updater: &UpdaterStep)
     // Install to a temporary path before moving it to the destination
     cmd.arg("install")
         .arg("axoupdater-cli")
+        // .arg("--features=tls_native_roots")
```
mistydemeo, line 212: Is this meant to be uncommented once we've shipped an axoupdater change?

mistydemeo (approved):

cargo-dist/src/lib.rs
Gankra: Ack, yes! and it has shipped

PR #1162, opened 2024-06-21
chore: add axoproject to autopublishes

Gankra:
This I believe requires a change to the token used to publish


mistydemeo (approved):
Makes sense. I assume @ashleygwilliams is the one who can update the token?

Gankra:
Nah i can do it, I'm leaving the PR open until i do.

PR #1164, opened 2024-06-24
WIP

Gankra:
working on fixing the release atomicity and ensuring it targets the right commit

atomicity:

the create command just takes the artifacts to upload. weirdly the edit command doesn't but that's ok because we only use edit for draft-release-workflows where spewing some random artifacts onto a draft that can be deleted is ~fine (the undraft happens last)

targetting commit (relevant for workflow-dispatch flow):

> If a matching git tag does not yet exist, one will automatically get created
from the latest state of the default branch. Use --target to override this.
To fetch the new tag locally after the release, do git fetch --tags origin.

we also already pass --target for the submodule flow, so we just need to change that to being unconditional, and write some logic to get the current repo commit instead of the submodule

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -177,29 +177,30 @@ impl GithubCiInfo {
 
         let mut release_args = vec![];
         let action;
+        // Always need to use the tag flag
+        release_args.push("\"${{ needs.plan.outputs.tag }}\"");
+
+        // If using remote repos, specify the repo
         if github_releases_repo.is_some() {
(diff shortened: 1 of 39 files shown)


mistydemeo (approved):

mistydemeo:
For setting the commit in a same-repo context, we can probably use [`GITHUB_SHA`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables). It's the specific commit that created the actions run, resolved at the time the action began. The ref is also provided, but since that can get theoretically targeted midflight, `GITHUB_SHA` is more specifically what we want.

PR #1171, opened 2024-06-25
feat: customize GitHub release phase

mistydemeo:
This adjusts GitHub release behaviour. Previously, we always created releases during the "announce" phase, but this introduced a brief race condition where the npm and Homebrew installers could reference a release that wouldn't be created for another ~30 seconds.

* The default behaviour has changed: the phase nw branches based on whether Axo releases is enabled. If enabled, we create releases during announce, and otherwise we create them in host. (We prefer a later release if Axo releases is enabled because the release text will reference Axo URLs, which won't be live until the announce phase.)
* It's now possible to customize the phase at which it occurs, and any manual selection takes precedence over the auto behaviour.

The new option isn't yet exposed via `init`; let's talk about if we want that, or if we want this to be a manual config option.

### DIFF
--- a/book/src/reference/config.md
+++ b/book/src/reference/config.md
@@ -466,6 +466,29 @@ Allows specifying which runner to use for a target. The keys within this table a
 In addition to defining runners for a target, it's also possible to specify a runner for the global, non-target-specific tasks using the `global` key. This runner will be used for tasks like `plan`, `host`, generating installers, and so on.
 
 
+### github-release
+
+> since 0.17.0
+
+Example: `github-release = "announce"`
+
+Possible values:
+
+* `auto`: create the GitHub Release whenever is best
+* `host`: create the GitHub Release during the host step
+* `announce`: create the GitHub Release during the announce step
+
+Controls which stage of the release process the GitHub Release will be created in.
+
+By default, the GitHub Release is created during the "host" phase, as it hosts the files some installers will try to download. If axo Releases is also enabled, it will be moved back to the "announce" phase, as the files will be primarily hosted on axo Releases, and GitHub Releases will be treated like a backup and announcement of the release.
+
+**Most users should be well-served by the default setting, and changing it is likely to introduce undesirable publishing race conditions.** The only reason you might want to override this setting is if you're using [`dispatch-releases = true`](#dispatch-releases) and you really want your git tag to be the last operation in your release process (because creating a GitHub Release necessarily creates the git tag if it doesn't yet exist). In this case setting github-release = "announce" will accomplish that, but see below for what race conditions this might introduce.
+
+If using only GitHub Releases, and you force it to run during "announce", there will be a very brief window (~30 seconds) during which generated Homebrew and npm installers are live and referencing URLs that will only exist when the GitHub Release is created, causing the packages to error out when installed.
+
+However, if you're publishing only packages that don't reference hosted artifacts (such as Cargo crates, or any custom publish job that fully embeds the binaries), then there is no race, and you could consider changing the default. That said, it would be a looming footgun if you ever introduce new publish jobs and forget about this.
+
+
 ### github-releases-repo
 
 > since 0.14.0
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -12,7 +12,8 @@ use tracing::warn;
 use crate::{
     backend::{diff_files, templates::TEMPLATE_CI_GITHUB},
     config::{
-        DependencyKind, HostingStyle, JinjaGithubRepoPair, ProductionMode, SystemDependencies,
+        DependencyKind, GithubReleasePhase, HostingStyle, JinjaGithubRepoPair, ProductionMode,
+        SystemDependencies,
     },
     errors::DistResult,
     DistGraph, SortedMap, SortedSet, TargetTriple,
@@ -74,6 +75,8 @@ pub struct GithubCiInfo {
     pub tag_namespace: Option<String>,
     /// `gh` command to run to create the release
     pub release_command: String,
+    /// Which phase to create the release at
+    pub release_phase: GithubReleasePhase,
 }
 
 impl GithubCiInfo {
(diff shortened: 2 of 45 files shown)


cargo-dist/src/config.rs
```diff
@@ -845,6 +854,30 @@ impl std::fmt::Display for InstallerStyle {
...
+    /// Create release during the "host" stage, before npm and Homebrew
+    Host,
+    /// Create release during the "announce" stage, after all publish jobs
+    Announce,
+}
+
+impl std::fmt::Display for GithubReleasePhase {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let string = match self {
+            GithubReleasePhase::Auto => "auto",
```
Gankra, line 873: It would be more conventional for us to just use None to capture this but since it's a global setting this is fine, I suppose.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -116,6 +119,21 @@ impl GithubCiInfo {
             .hosts
             .clone();
 
+        let release_phase = if dist.github_release == GithubReleasePhase::Auto {
+            // If Axo is in use, we want the release to come late
+            // because the release wil lcontain links to Axo URLs
```
Gankra, line 124:
```suggestion
            // because the release will contain links to Axo URLs
```

cargo-dist/src/backend/ci/github.rs
```diff
@@ -116,6 +119,21 @@ impl GithubCiInfo {
...
+            // that won't become live until the announce phase.
```
Gankra, line 125: I was about to say this is incorrect but oh right the actual release *body* does use the prettier URLs that need an announce to go live (the actual artifacts refer to the already-live artifact set).

cargo-dist/src/backend/ci/github.rs
```diff
@@ -116,6 +119,21 @@ impl GithubCiInfo {
...
+            if hosting_providers.contains(&HostingStyle::Axodotdev) {
+                GithubReleasePhase::Announce
+            // Otherwise, it's safe for the release to occur during host.
```
Gankra, line 128: That said imo the primary framing is that we *want* to do the Github release in announce, and that axo releases makes it safe to do so. So this comment feels backwards. Putting it in Host is the "do it for safety" option.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -74,6 +75,8 @@ pub struct GithubCiInfo {
     pub tag_namespace: Option<String>,
     /// `gh` command to run to create the release
     pub release_command: String,
+    /// Which phase to create the release at
+    pub release_phase: GithubReleasePhase,
```
Gankra, line 79: I would prefer that this be called github_release_phase, as this does not e.g. affect axo releases.

book/src/reference/config.md
```diff
@@ -466,6 +466,25 @@ Allows specifying which runner to use for a target. The keys within this table a
...
+
+By default, the GitHub release is created during the "host" phase when releasing only to GitHub, and during the "announce" phase when releasing to both GitHub and Axo Releases. This can be overridden by setting this to "host" or "announce" to force those phases regardless of other circumstances. Most users should be well-served by the default setting.
+
+The choice of when to sequence the release has a few implications when using installers published to external services, like npm or Homebrew. The order of phases runs like so:
+
+host => publish (Homebrew, npm) => announce
+
+If the GitHub release is created during "announce", there will be a very brief window (30 seconds) during which the Homebrew and npm installers are live and referencing assets that don't exist yet.
+
+Likewise, Axo Releases assets don't become public until the "announce" phase, so if a GitHub release is published during the "host" phase while hosting assets in Axo Releases, there will be a short window during which the GitHub release contains links to assets which haven't yet gone live.
```
Gankra, line 485: Ok yeah this is confusing/misleading... maybe something like:

> If you're using GitHub Releases as your only hosting, setting `github-release = "announce"` will result in a brief window (~30 seconds) where your published Homebrew and npm packages are referencing assets that don't exist yet.
> 
> If you're also using axo Releases then the installers will always have live URLs, and either option is fine for your published packages. However the GitHub Release's text will contain `curl | sh` expressions that reference special prettier axo URLs that only become live during the "announce" step, so setting `github-release = "host"` still isn't advisable.

cargo-dist/templates/ci/github/partials/publish_github.yml.j2
```diff
@@ -0,0 +1,37 @@
+      # Create a GitHub Release while uploading all files to it
+      - name: "Download GitHub Artifacts"
+        uses: actions/download-artifact@v4
+        with:
+          pattern: artifacts-*
+          path: artifacts
+          merge-multiple: true
```
Gankra, line 7: I was gonna complain this is redundant (we end up downloading all the artifacts twice) but now I see we've always been doing this and there's some reason for why, so nvm.

cargo-dist/src/backend/ci/github.rs
Gankra: Actually I see we've broke this on several fields already so nvm, rather not add more churn.

Gankra (approved):
I think all my change requests are docs/comments.

cargo-dist/src/backend/ci/github.rs
mistydemeo: Fair enough, yeah - I can reverse the comments. I've been thinking about the npm releases, which made me think of "do it in host" as the ideal and announce as the safety when Axo is used.

cargo-dist/templates/ci/github/partials/publish_github.yml.j2
mistydemeo: I feel like this kinda branch always exposes those.

cargo-dist/src/config.rs
mistydemeo: Yeah I thought about this for a bit. From a user-facing config side, I always appreciate being able to explicitly specify a setting and still have the "auto" setting be an option, just for clarity.

cargo-dist/src/backend/ci/github.rs
mistydemeo: Updated comments.

PR #1173, opened 2024-06-26
release: 0.17.0-prerelease.2

mistydemeo:


axoproject/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "axoproject"
 description = "project detection logic for various axo.dev applications"
-version = "0.7.1"
+version = "0.17.0-prerelease.2"
```
ashleygwilliams, line 4: big jump! did this get published?

ashleygwilliams (approved):

axoproject/Cargo.toml
mistydemeo: No, not yet.

The bump is because we need to keep axoproject's versions in lockstep with cargo-dist's versions. There's a similarly versioned (but not published to crates.io) 0.17.0-prerelease.1

PR #1179, opened 2024-07-02
feat: add github-custom-job-permissions

Gankra:
This lets you overide the permissions we default assign to a custom task


cargo-dist/tests/integration-tests.rs
```diff
@@ -363,6 +363,7 @@ targets = ["x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows
 ci = ["github"]
 build-local-artifacts = false
 local-artifacts-jobs = ["./local-artifacts"]
+github-custom-job-permissions = {{ local-artifacts = ["packages: write"] }}
```
Gankra, line 366: example of using the feature (note the `{`'s are doubled because this is a rust format)

ashleygwilliams: appreciate pointing this out- would be good to get docs on this feature otherwise i am eager to unfork them and we have config 1.0 coming up so i say ship it

book/src/reference/config.md
```diff
@@ -449,6 +449,24 @@ Defaults false (but may become true in the future).
...
 
+### github-custom-job-permissions
+
+> since 0.18.0
+
+Example:
+
+```
+publish-jobs = ["npm", "./my-custom-publish"]
+github-custom-job-permissions = { "my-custom-publish" = ["packages: admin"] }
```
mistydemeo, line 460: I'm kind of torn on the format here. Preformatted yaml-compatible permissions expressions is easier on us to format, but it feels like it introduces new room for user error.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -164,13 +178,50 @@ impl GithubCiInfo {
...
-        let global_artifacts_jobs = dist.global_artifacts_jobs.clone();
-        let host_jobs = dist.host_jobs.clone();
+
+        let mut job_permissions = dist.github_custom_job_permissions.clone();
+        // use publish jobs default to elevated priviledges
+        for name in &dist.user_publish_jobs {
+            job_permissions.entry(name.clone()).or_insert_with(|| {
+                [
+                    ("id-token".to_owned(), "write".to_owned()),
+                    ("packages".to_owned(), "write".to_owned()),
```
mistydemeo, line 188: This drops attestations from the root permisisons, breaking any build steps that need to create attestations. eg, with default settings and attestations on:

```diff
 permissions:
-  contents: write
-  id-token: write
-  attestations: write
+  "contents": "write"
+  "id-token": "write"
+  "packages": "write"
```

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -24,12 +24,12 @@
 
 name: Release
 
+{{%- if root_permissions %}}
 permissions:
-  contents: write
-  {{%- if github_attestations %}}
```
mistydemeo, line 29: We still need an `if github_attestations` here, or to add these to the root permissions if attestations are on.

mistydemeo (changes requested):

cargo-dist/src/backend/ci/github.rs
Gankra: yikes, good catch!

cargo-dist/src/backend/ci/github.rs
```diff
@@ -164,13 +175,33 @@ impl GithubCiInfo {
...
+                    ("id-token".to_owned(), GithubPermission::Write),
+                    ("packages".to_owned(), GithubPermission::Write),
+                ])
+            });
+        }
+
+        let mut root_permissions = GithubPermissionMap::new();
+        root_permissions.insert("contents".to_owned(), GithubPermission::Write);
+        if github_attestations {
+            root_permissions.insert("attestations".to_owned(), GithubPermission::Write);
```
mistydemeo, line 193: Attestations use two permissions, not one: it also needs `id-token`. https://github.com/axodotdev/cargo-dist/blob/4b8b18d315a92d1c4aac3063ff01acd0a5e4b250/cargo-dist/templates/ci/github/release.yml.j2#L30-L31

mistydemeo (changes requested):

mistydemeo (approved):
Looks good. Most recent test worked.

PR #1180, opened 2024-07-02
docs: improve example for `bin-aliases` option

zanieb:


book/src/reference/config.md
```diff
@@ -140,8 +140,8 @@ Example:
 
 ```toml
 [package.metadata.dist.bin-aliases]
-"myrealbin" = ["somealias", "otheralias"]
-"myotherbin" = "alias"
```
zanieb, line 144: This is not valid for this option.

book/src/reference/config.md
```diff
@@ -140,8 +140,8 @@ Example:
...
+"mybin" = ["somealias"]
+"myotherbin" = ["someotheralias", "anotheralias"]
```
zanieb, line 144: I found one being "real" and the other not confusing.

PR #1182, opened 2024-07-03
Support cdylibs from Rust/generic projects

mistydemeo:
This is the start of the branch for cdylib support - it doesn't fully work yet.

We had cdylibs available from axoproject/`cargo metadata` until now, but we ignored them. This pipes them into the buildsystem and treats them as binaries like any others. For most of the layers of cargo-dist, we have a single set of "binaries" but we now type them so that we can keep track of which are executables and which are libraries. We separate them back out at the time that we create installers to simplify the process of deciding where they get installed to.

Because we lean on axoproject to tell us about cdylibs, which already supported them in both rust and generic projects, users don't have to make any changes to their projects in order to use this feature. As long as a project declares cdylibs, we'll start installing them. (We always built them, we just didn't include them in our tarballs.)

Installer support:

* Homebrew is opinionated about where libraries go, so we use Homebrew's native features for this. Libraries are separated out via `lib.install` so they go in the right place.
* shell/powershell: For CARGO_HOME-style installs, we separate the libraries from the binaries in order to avoid putting them on the `PATH`. For example, if we install binaries to `~/.cargo/bin`, we place the libraries in `~/.cargo/lib`. For other installs, where everything is dumped in one directory, we place libs alongside binaries.
* npm: No change needed. We don't advertise them in npm packages, but they need to be there in case a binary depends on them. Luckily, since we already unpack the entire tarball instead of just the binaries, that's already done.

### DIFF
--- a/cargo-dist/src/backend/installer/mod.rs
+++ b/cargo-dist/src/backend/installer/mod.rs
@@ -63,6 +63,8 @@ pub struct InstallerInfo {
     pub receipt: Option<InstallReceipt>,
     /// Aliases to install binaries under
     pub bin_aliases: BTreeMap<String, BTreeMap<String, Vec<String>>>,
+    /// Whether to install generated C dynamic libraries
+    pub install_cdylibs: bool,
 }
 
 /// A fake fragment of an ExecutableZip artifact for installers
@@ -72,8 +74,10 @@ pub struct ExecutableZipFragment {
     pub id: String,
     /// The target the artifact supports
     pub target_triple: TargetTriple,
-    /// The binaries the artifact contains (name, assumed at root)
-    pub binaries: Vec<String>,
+    /// The executables the artifact contains (name, assumed at root)
+    pub executables: Vec<String>,
+    /// The dynamic libraries the artifact contains (name, assumed at root)
+    pub cdylibs: Vec<String>,
     /// The style of zip this is
     pub zip_style: ZipStyle,
     /// The updater associated with this platform
--- a/cargo-dist/src/build/cargo.rs
+++ b/cargo-dist/src/build/cargo.rs
@@ -201,11 +201,8 @@ pub fn build_cargo_target(
         };
         match message {
             cargo_metadata::Message::CompilerArtifact(artifact) => {
-                let Some(new_exe) = artifact.executable else {
-                    continue;
-                };
-                // Hey we got an executable, record that fact
-                expected.found_bin(artifact.package_id.to_string(), new_exe, artifact.filenames);
+                // Hey we got some files, record that fact
+                expected.found_bins(artifact.package_id.to_string(), artifact.filenames);
             }
             _ => {
                 // Nothing else interesting?
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -378,6 +378,17 @@ pub struct Binary {
     pub copy_symbols_to: Vec<Utf8PathBuf>,
     /// feature flags!
     pub features: CargoTargetFeatures,
+    /// What kind of binary this is
+    pub kind: BinaryKind,
+}
+
+/// Different kinds of binaries cargo-dist knows about
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum BinaryKind {
+    /// Standard executable
+    Executable,
+    /// C-style dynamic library (.so/.dylib/.dll)
+    DynamicLibrary,
 }
 
 /// A build step we would like to perform
@@ -685,6 +696,15 @@ pub struct Release {
     ///
     /// The string is the name of the binary under that package (without .exe extension)
     pub bins: Vec<(PackageIdx, String)>,
+    /// C dynamic libraries that every variant should ostensibly provide
+    ///
+    /// The string is the name of the library, without lib prefix, and without platform-specific suffix (.so, .dylib, .dll)
+    /// Note: Windows won't include lib prefix in the final lib.
+    pub cdylibs: Vec<(PackageIdx, String)>,
+    /// Whether to package C dynamic libraries in the final archive
+    pub package_cdylibs: bool,
+    /// Whether to install packaged C dynamic libraries
+    pub install_cdylibs: bool,
     /// Artifacts that are shared "globally" across all variants (shell-installer, metadata...)
     ///
     /// They might still be limited to some subset of the targets (e.g. powershell scripts are
@@ -899,6 +919,8 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             bin_aliases: _,
             display: _,
             display_name: _,
+            package_cdylibs: _,
+            install_cdylibs: _,
         } = &workspace_metadata;
 
         let desired_cargo_dist_version = cargo_dist_version.clone();
@@ -1191,6 +1213,8 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             bin_aliases,
             display,
             display_name,
+            package_cdylibs,
+            install_cdylibs,
             // The rest of these are workspace-only
             precise_builds: _,
             merge_tasks: _,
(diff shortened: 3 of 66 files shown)


mistydemeo:
Current status: cargo artifacts fully piped through, and support added to Homebrew, sh, and ps1 installers. Need to ensure generic builds work.

cargo-dist/src/tasks.rs
```diff
@@ -1349,9 +1376,32 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+                    platform_lib_prefix = "lib";
+                };
 
-                let file_name = format!("{binary_name}{platform_exe_ext}");
+                let platform_lib_ext = if target_is_windows {
+                    ".dll"
+                } else if target.contains("linux") {
+                    ".so"
+                } else if target.contains("darwin") {
+                    ".dylib"
```
Gankra, line 1394: bsds also use this, fwiw

cargo-dist/src/tasks.rs
```diff
@@ -1349,9 +1376,32 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+                } else {
+                    panic!("Unknown os in {target}")
+                };
```
Gankra, line 1397: hmm I suggest making this more descriptive / a proper error

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -282,6 +294,12 @@ function Invoke-Installer($bin_paths, $platforms) {
       }
     }
   }
+  foreach ($lib_path in $artifacts["lib_paths"]) {
+    $installed_file = Split-Path -Path "$lib_path" -Leaf
+    Copy-Item "$lib_path" -Destination "$dest_dir"
+    Remove-Item "$lib_path" -Recurse -Force
+    Write-Information "  $installed_file"
```
Gankra, line 301: Probably don't want to bother informing users about installed dylibs... probably?

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -376,6 +382,7 @@ install() {
         elif [ -n "${HOME:-}" ]; then
             _receipt_install_dir="$HOME/.cargo"
             _install_dir="$HOME/.cargo/bin"
+            _lib_install_dir="$HOME/.cargo/lib"
```
Gankra, line 385: .cargo/lib doesn't exist and i don't feel like we should be creating that...

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -436,6 +447,14 @@ install() {
         done
         say "  $_bin_name"
     done
+    # Like the above, but no aliases
+    for _lib_name in $_libs; do
+        local _lib="$_src_dir/$_lib_name"
+        ensure mv "$_lib" "$_lib_install_dir"
+        # unzip seems to need this chmod
+        ensure chmod +x "$_lib_install_dir/$_lib_name"
+        say "  $_lib_name"
```
Gankra, line 456: similarly, not sure users should care about cdylibs

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -436,6 +447,14 @@ install() {
         done
```
Gankra, line 447: while we're in here wanna add the says for the aliases?

Gankra (changes requested):
This actually looks basically right, but some detail nits on installers

cargo-dist/src/build/cargo.rs
```diff
@@ -201,11 +201,26 @@ pub fn build_cargo_target(
...
+                } else {
+                    // "executable" is singled out as an artifact
+                    // type in cargo's output; for libraries, we have
+                    // to just look at the list of output files
+                    // and let found_bin sort it out.
+                    for new_exe in &artifact.filenames {
+                        expected.found_bin(
+                            artifact.package_id.to_string(),
+                            new_exe.to_owned(),
+                            artifact.filenames.clone(),
```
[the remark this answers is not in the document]
mistydemeo: Yeah, it may make more sense for a `found_bins` that takes multiple paths. Really it's just hedging on the fact here that the full list of filenames is passed for the random other stuff that might be there (symbols, etc.), but the single-filename field is where we put maybe-requested files.

cargo-dist/src/tasks.rs
mistydemeo: Oh is that where Apple got it from! That makes sense. We don't support BSDs atm, but worth keeping in mind for the future

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Probably not, yeah. We might but we also might not?

My thought is more or less:

* If the lib is an actual downstream-usable artifact, like a C library for non-Rust software, I might want to know it's there. (Say the user packaged something like glib - it could be nice to know I got `libglib-2.0.0.dylib`.)
* If the lib is just used for a binary, or a secondary artifact like Python/Ruby/etc bindings, I probably don't want to know it's there.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: I'm not tied to any specific path structure, but I do think we should be keeping the lib out of the user's `PATH`. Since libraries have executable permissions, installing them in `~/.cargo/bin` means the user will get them in tab-complete suggestions (despite not being really runnable) - which feels wrong. Since `~/.cargo` is organized, we can reserve `~/.cargo/bin` for actual binaries and add a sidecar directory for libs.

cargo-dist/src/backend/installer/mod.rs
```diff
@@ -72,8 +72,10 @@ pub struct ExecutableZipFragment {
     pub id: String,
     /// The target the artifact supports
     pub target_triple: TargetTriple,
-    /// The binaries the artifact contains (name, assumed at root)
-    pub binaries: Vec<String>,
+    /// The executables the artifact contains (name, assumed at root)
+    pub executables: Vec<String>,
+    /// The dynamic libraries the artifact contains (name, assumed at root)
+    pub libraries: Vec<String>,
```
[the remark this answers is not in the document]
mistydemeo: Good call, done for this and related fields.

mistydemeo:
All right, I've now tested on both Mac and Windows with the shell and ps1 installers. All works as expected.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: We discussed this and agreed that right now it's probably best for us to install these to begin with, and we can adjust that behaviour in future versions.

cargo-dist/src/tasks.rs
mistydemeo: Thought: now that axoproject is in-tree, maybe in the future, we try to put all "what's the file extension/prefix" logic into axoproject's `platforms` mod.

cargo-dist/src/tasks.rs
```diff
@@ -2882,6 +2985,8 @@ pub struct InstallReceipt {
     pub install_prefix: String,
     /// A list of all binaries installed by this app
     pub binaries: Vec<String>,
+    /// A list of all libraries installed by this app
+    pub libraries: Vec<String>,
```
Gankra, line 2989: should probably also be called cdylibs..?

Gankra (approved):
Sure *seems* like this all makes sense.

cargo-dist/src/tasks.rs
mistydemeo: I was wondering about how much we want to expose "cdylibs" vs "libraries" to the user, but I guess this isn't user-facing. We can be more specific about the technical details and separate other libraries into other categories here.

PR #1185, opened 2024-07-03
chore: remove unecessary cruft from create-react-app test example

Gankra:
dependabot keeps freaking out about it


mistydemeo (approved):
Always nice when we can nix 30,000 lines safely.

PR #1189, opened 2024-07-08
Fix for Bash env script injection on RPM family distros

### LINKED ISSUE #1187: Installer Bash source injection is still insufficient on some distros (Fedora, RHEL-compatibles)
Related to #547 - on at least Fedora 40 and AlmaLinux 9, the current sourcing injection logic isn't sufficient for non-login Bash shells.

This seems to be due to the RPM family (?)  having a `skel` like this:
```
.bash_logout
.bash_profile
.bashrc
```
While (for example) Debian has one like this:
```
.bash_logout
.bashrc
.profile
```
The current implementation iterates over the list `".bash_profile .bash_login .bashrc"` in order, only editing the first valid file it finds in this list.

With the latter layout, there is no `.bash_profile`, so `.bashrc` is edited and everything works as expected. (`.profile` is also edited, presumably for POSIX? At any rate, I believe Bash picks up on that too for login shells.)

With the former layout, it instead edits `.bash_profile`, which Bash only sources in a login shell. GUI terminal emulators are left high and dry.

It looks like the [intention](https://github.com/axodotdev/cargo-dist/pull/555#discussion_r1387056559) was to copy how `rustup` (which *does* work for Fedora and co.) implements it, but it seems the logic was solely based on the [ZSH handling](https://github.com/rust-lang/rustup/blob/b5d74ac69307de29cee586c779bdb6b3aee7f7af/src/cli/self_update/shell.rs#L193-L206) (which only edits the first file it finds) - the [current Bash handling](https://github.com/rust-lang/rustup/blob/b5d74ac69307de29cee586c779bdb6b3aee7f7af/src/cli/self_update/shell.rs#L147) actually seems to be aware of this pitfall and just blasts `source $HOME/.cargo/env` into `.bash_profile`, `.bash_login` *and* `.bashrc`.

First encountered this in atuinsh/atuin#2235, if that helps. I'll try to write a PR myself, but right now I need to get to bed :sweat_smile:

Colonial-Dev:
Resolves #1187 - this is a quick fix I hacked up that piggybacks on the existing `add_install_dir_to_path` function.

It works well in a Fedora container, but shell scripting is not my strong suit - I'll gladly cook up something different if there's a footgun I overlooked.


Gankra:
Oh wow, I super appreciate the research you did into this!

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -443,7 +443,7 @@ install() {
         add_install_dir_to_ci_path "$_install_dir"
         add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" ".profile" "sh"
         exit1=$?
-        add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" ".bash_profile .bash_login .bashrc" "sh"
+        shotgun_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" ".bash_profile .bash_login .bashrc" "sh"
```
Gankra, line 446: I believe this *slightly* changes behaviour here. Currently we will create `.bash_profile` if none of these files exist, and with shotgun we no long detect this case. Is that intentional?

Colonial-Dev: No, it's not - good catch, thanks! I'll fix that.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -608,17 +608,22 @@ shotgun_install_dir_to_path() {
...
             local _rcfile_abs="$_home/$_rcfile_relative"
 
             if [ -f "$_rcfile_abs" ]; then
+                _found=true
                 add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" "$_rcfile_relative" "$_shell"
             fi
         done
+
+	    if [ "$_found" = false ]; then
+            add_install_dir_to_path "$_install_dir_expr" "$_env_script_path" "$_env_script_path_expr" "$(echo "$_rcfiles" | awk '{ print $1 }')" "$_shell"
```
Gankra, line 625: What does this extra awk accomplish?

Gankra:
Hmm interesting it looks like rustup currently in fact doesn't create a bash file if they're all missing, so maybe your original commit should be desired behaviour? (It *does* unconditionally create .profile)

https://github.com/rust-lang/rustup/blob/fe99f782cf362a11609768d58ea27183ae685532/src/cli/self_update/shell.rs#L136-L156

cargo-dist/templates/installer/installer.sh.j2
Colonial-Dev: Oops - nothing, that was left over from the check I pasted from `add_install_dir_to_path` before I checked and found that `var=bool` is dash/POSIX compliant.

Colonial-Dev:
I can definitely see an argument for that. Should we make the other shells behave similarly, in that case? I did notice that the Atuin installer was creating a `.zshrc` in my home despite `zsh` not even being installed on my system

Gankra:
This is really interesting, it seems like rustup has tightened up how hard it shotguns since last i checked. It means that someone who installs zsh later won't see cargo on path, hmmm...

Gankra:
(Gonna consult with coworkers on what we wanna do here)

mistydemeo:
Thanks for this PR! It's a big help.

> Most people will (I'm assuming) be using non-login shells

macOS's terminal uses login shells by default, so we do want to plan around it. (I don't even have a `.bashrc` on my system!)

I think, reading up on what rustup does, that their logic makes some sense. Plus, if we stop creating a file by default, we can completely dodge the "login shell or not login shell" question.

The "should we create `.profile`" question is kinda sorta separate. The answer "this is the only one standardized by POSIX, so we should always handle it" makes a certain amount of sense, but in practice absolutely nothing on my system has ever interacted with it (and only cargo-dist has ever put stuff in there, so it's a graveyard of test data). It doesn't hurt to make that the default file we create, but I'm not convinced it's actually going to affect very many users.

Colonial-Dev:
Interesting, I didn't know that about macOS.

With respect to `.profile` - I think it makes sense to keep it, even if just for the token POSIX acknowledgement. Best effort is better than no effort.

mistydemeo:
Sounds good! In that case, how about we keep what you've written except we change the order so the "write to the first in the list" is `.profile`?

Colonial-Dev:
Sure! Do you think we need to alter the behavior for the other shells at all? I'd need to double check, but I think ZSH etc. all match `rustup`.

Colonial-Dev:
Double checked, and the existing logic for POSIX/`Zsh`/`fish` seems to be valid and equivalent to `rustup`'s implementation.

PR #1195, opened 2024-07-09
feat: inline default install-path

### LINKED ISSUE #1194: `cargo dist init` should explicitly mark the current install-path
We're planning to migrate to a new default install path in the future. Before we do that, the next point or major release of cargo-dist should ensure that `cargo dist init` explicitly adds the current `install-path` to the user's `Cargo.toml` in order to ensure that they keep that value in future versions unless they make the explicit choice to switch to our new default.

While alternate `install-path`s will be in the user's `Cargo.toml`, the current default, `CARGO_HOME`, is implicit.

mistydemeo:
This preps us for a future release where we change the default, making it easier for people using the previous default to keep it.

I considered migrating away from `Opt<Vec<_>>`, but this made it easier to keep `opt_string_or_vec`.

Fixes #1194.

### DIFF
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -1140,6 +1140,13 @@ pub enum InstallPathStrategy {
     },
 }
 
+impl InstallPathStrategy {
+    /// Returns the default set of install paths
+    pub fn default_list() -> Vec<Self> {
+        vec![InstallPathStrategy::CargoHome]
+    }
+}
+
 impl std::str::FromStr for InstallPathStrategy {
     type Err = DistError;
     fn from_str(path: &str) -> DistResult<Self> {
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -8,8 +8,8 @@ use serde::Deserialize;
 
 use crate::{
     config::{
-        self, CiStyle, CompressionImpl, Config, DistMetadata, HostingStyle, InstallerStyle,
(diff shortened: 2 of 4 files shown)


cargo-dist/src/config.rs
```diff
@@ -269,7 +269,10 @@ pub struct DistMetadata {
     /// All of these error out if the required env-vars aren't set. In the future this may
     /// allow for the input to be an array of options to try in sequence.
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[serde(default, with = "opt_string_or_vec")]
+    #[serde(
+        default = "InstallPathStrategy::default_list",
+        with = "opt_string_or_vec"
+    )]
     pub install_path: Option<Vec<InstallPathStrategy>>,
```
Gankra, line 276: serde defaults continue to not be the correct way to do defaults in our config system -- this will break config inheritance. Defaults need to be applied where all the other defaults are. Also we probably shouldn't write back this default unless they have a shell/powershell installer. The config is meaningless otherwise.

Gankra (changes requested):

cargo-dist/src/config.rs
mistydemeo: Ahhh, yeah, that makes sense. I hadn't considered config inheritance.

mistydemeo:
Updated to make this an `init`-time feature.

Gankra (approved):

PR #1210, opened 2024-07-16
Track glibc versions and other runtime requirements

mistydemeo:
So far, we have one runtime requirement expressed directly in our installers: the glibc version used to build. At the moment, this is hardcoded based on the value in the default runner, but we'd like to have it be determined at buildtime and then used in the installer as appropriate.

This adds that tracking. At buildtime, we now add to the `SystemInfo` field to track system environment information specific to the build environment. We should expand this in time, but at the moment we track builder glibc version and macOS version. We then collate that information in the installer to calculate the safest minimum glibc version and inject that into the installers rather than hardcoding.

At the moment this reuses some platform information from the `platform` module, but not all of it - that module is focused on covering one platform at a time, while we're trying to collate information on every platform during install time. I'm not 100% sure if we want to be reusing these in this shape, or at least we may want to rethink the way `platform` is put together down the line if we want to be able to express more of this in the installers.

Some more thoughts on build info to track:

* musl version (on musl-based systems)
* Xcode/CLT version (on macOS)

We may also want to start using the macOS version to express a version minimum in the installers.

### DIFF
--- a/cargo-dist/src/build/cargo.rs
+++ b/cargo-dist/src/build/cargo.rs
@@ -4,12 +4,13 @@ use std::env;
 
 use axoprocess::Cmd;
 use axoproject::WorkspaceIdx;
-use cargo_dist_schema::DistManifest;
+use cargo_dist_schema::{BuildEnvironment, DistManifest};
 use miette::{Context, IntoDiagnostic};
 use tracing::warn;
 
 use crate::build::BuildExpectations;
 use crate::env::{calculate_ldflags, fetch_brew_env, parse_env, select_brew_env};
+use crate::linkage::determine_build_environment;
 use crate::{errors::*, BinaryIdx, BuildStep, DistGraphBuilder, TargetTriple, PROFILE_DIST};
 use crate::{
     CargoBuildStep, CargoTargetFeatureList, CargoTargetPackages, DistGraph, RustupStep, SortedMap,
@@ -222,6 +223,16 @@ pub fn build_cargo_target(
 
     // Process all the resulting binaries
     expected.process_bins(dist_graph, manifest)?;
+
+    let system_id = format!("build:local:{}", target.target_triple);
+    // Record the build environment
+    if let Some(info) = manifest.systems.get_mut(&system_id) {
+        // Not yet set; do that now
+        if matches!(info.build_environment, BuildEnvironment::Indeterminate) {
+            let build_environment = determine_build_environment(&target.target_triple);
+            info.build_environment = build_environment;
+        }
+    };
 
     Ok(())
 }
--- a/cargo-dist/src/build/cargo.rs
+++ b/cargo-dist/src/build/cargo.rs
@@ -4,13 +4,12 @@ use std::env;
 
 use axoprocess::Cmd;
 use axoproject::WorkspaceIdx;
-use cargo_dist_schema::{BuildEnvironment, DistManifest};
+use cargo_dist_schema::DistManifest;
 use miette::{Context, IntoDiagnostic};
 use tracing::warn;
 
 use crate::build::BuildExpectations;
 use crate::env::{calculate_ldflags, fetch_brew_env, parse_env, select_brew_env};
-use crate::linkage::determine_build_environment;
 use crate::{errors::*, BinaryIdx, BuildStep, DistGraphBuilder, TargetTriple, PROFILE_DIST};
 use crate::{
     CargoBuildStep, CargoTargetFeatureList, CargoTargetPackages, DistGraph, RustupStep, SortedMap,
@@ -223,16 +222,6 @@ pub fn build_cargo_target(
 
     // Process all the resulting binaries
     expected.process_bins(dist_graph, manifest)?;
-
-    let system_id = format!("build:local:{}", target.target_triple);
-    // Record the build environment
-    if let Some(info) = manifest.systems.get_mut(&system_id) {
-        // Not yet set; do that now
-        if matches!(info.build_environment, BuildEnvironment::Indeterminate) {
-            let build_environment = determine_build_environment(&target.target_triple);
-            info.build_environment = build_environment;
-        }
-    };
 
     Ok(())
 }
--- a/cargo-dist/src/build/generic.rs
+++ b/cargo-dist/src/build/generic.rs
@@ -5,12 +5,13 @@ use std::{env, process::ExitStatus};
 use axoprocess::Cmd;
 use axoproject::WorkspaceIdx;
 use camino::{Utf8Path, Utf8PathBuf};
-use cargo_dist_schema::DistManifest;
+use cargo_dist_schema::{BuildEnvironment, DistManifest};
 
 use crate::{
     build::{package_id_string, BuildExpectations},
     copy_file,
     env::{calculate_cflags, calculate_ldflags, fetch_brew_env, parse_env, select_brew_env},
+    linkage::determine_build_environment,
     ArtifactKind, BinaryIdx, BuildStep, DistError, DistGraph, DistGraphBuilder, DistResult,
     ExtraBuildStep, GenericBuildStep, SortedMap, TargetTriple,
 };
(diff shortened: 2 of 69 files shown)


cargo-dist/src/platform.rs
```diff
@@ -79,6 +81,28 @@ pub enum RuntimeCondition {
...
+/// Runtime conditions collected for every platform
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+#[derive(Default)]
+pub struct RuntimeConditions {
+    /// Linux-specific conditions
+    pub linux: LinuxRuntimeConditions,
+    /// MacOS-specific conditions
+    pub macos: Option<RuntimeCondition>,
+}
+
```
Gankra, line 104: per discussion: we should just drop the enum entirely in favour of a struct with many Option fields. It's also not clear to me that we need this level of platform distinction, since all the options apply to the platform uniformly

cargo-dist/src/tasks.rs
```diff
@@ -3066,3 +3084,36 @@ fn submodule_head(submodule_path: &Utf8PathBuf) -> DistResult<Option<String>> {
         Ok(Some(commit.to_owned()))
     }
 }
+
+// Picks the newest glibc in the case of conflicts
+fn glibc_thunderdome(glibcs: Vec<GlibcVersion>) -> GlibcVersion {
```
Gankra, line 3089: 10/10 name

(this is a normal comment when it can/should be a doc-comment for intellisense)

cargo-dist/src/tasks.rs
```diff
@@ -1954,6 +1960,9 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             return;
         };
         let bin_aliases = release.bin_aliases.for_targets(&target_triples);
+
+        let runtime_conditions = pick_runtime_conditions(&self.manifest.systems);
```
Gankra, line 1964: Why can't this be computed centrally in platform_support?

Gankra: Specifically this info would ideally be embedded in `artifacts`

Gankra: Ah I see right there's no more code between here and the template... Ok so I propose that the "right" information be computed centrally in platform_support and then we have another "compatibility method" like .fragments() that does the flattening to runtime_conditions.

cargo-dist/src/build/cargo.rs
```diff
@@ -223,6 +224,16 @@ pub fn build_cargo_target(
     // Process all the resulting binaries
     expected.process_bins(dist_graph, manifest)?;
 
+    let system_id = format!("build:local:{}", target.target_triple);
```
Gankra, line 227: It feels like something has gone wrong if we're synthesizing this id here -- isn't the current id in the WorkspaceGraph?

Gankra: Oh is this grappling with the fact that the "system" keys are too coarse (concatenating target triples when you try to build multiple on one machine?)

cargo-dist/src/linkage.rs
```diff
@@ -434,3 +437,57 @@ pub fn determine_linkage(path: &Utf8PathBuf, target: &str) -> DistResult<Linkage
...
 }
+
+/// Determine the build environment on the current host
+/// This should be done local to the builder!
+pub fn determine_build_environment(target: &str) -> DistResult<BuildEnvironment> {
+    if target.contains("darwin") {
+        determine_macos_build_environment()
+    } else if target.contains("linux") {
+        determine_linux_build_environment()
+    } else if target.contains("windows") {
```
Gankra, line 448: Hmm errors here are hard... this is conceptually very similar to linkage which we don't consider an error to fail to collect. I think it would be correct to do the same here? I don't want to fail someone's release process on this, even if it means the installer doesn't have perfect information.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -28,11 +28,9 @@ EORECEIPT
...
 RECEIPT_HOME="${HOME}/.config/{{ app_name }}"
 
-# glibc provided by our Ubuntu 20.04 runners;
-# in the future, we should actually record which glibc was on the runner,
-# and inject that into the script.
-BUILDER_GLIBC_MAJOR="2"
-BUILDER_GLIBC_SERIES="31"
+# glibc provided by the system(s) which built these binaries
+BUILDER_GLIBC_MAJOR="{{ runtime_conditions.linux.glibc.major }}"
+BUILDER_GLIBC_SERIES="{{ runtime_conditions.linux.glibc.series }}"
```
Gankra, line 33: These are currently nullable, right? And Extremely bad/chaotic things would happen if they were null? (in all the kinds of installers)

If we don't want to handle that in this PR, we probably want somewhere in the platform code default this to 2.31 when no info is available

cargo-dist/src/tasks.rs
```diff
@@ -1094,6 +1098,8 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         let system = SystemInfo {
             id: system_id.clone(),
             cargo_version_line,
+            // Real value calculated during the build
+            build_environment: BuildEnvironment::Indeterminate,
```
Gankra, line 1102: hmm interesting... probably fine...

cargo-dist/src/build/generic.rs
```diff
@@ -216,6 +217,16 @@ pub fn build_generic_target(
...
 
+    // Record the build environment
+    let system_id = format!("build:local:{}", target.target_triple);
+    if let Some(info) = manifest.systems.get_mut(&system_id) {
+        // Not yet set; do that now
+        if matches!(info.build_environment, BuildEnvironment::Indeterminate) {
+            let build_environment = determine_build_environment(&target.target_triple)?;
+            info.build_environment = build_environment;
+        }
+    };
```
Gankra, line 228: I don't really get why you compute this in the middle of each build step -- it's a global property of the system we're running on. This could be computed in compute_build_steps or anything more universal/shared, right? (mostly don't want code duplication here)

Gankra (changes requested):

cargo-dist/src/tasks.rs
mistydemeo: The idea here is this also stands in for "we couldn't calculate it" or "it didn't get calculated for some reason". So set it as the default, and if it lands there on the end we know we had some kind of unmeasurable system (or a logic error)

cargo-dist/src/platform.rs
mistydemeo: Merged in your other PR that did this.

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Yeah, that's one of the things I was thinking about with reusing these types. Specifically, we always calculate a value here, but the type itself is an option - it's probably dangerous to rely on it always being `Some()` even if we always assign to it.

cargo-dist/src/build/generic.rs
mistydemeo: Yes, that's also true. I sort of automatically associated "build environment properties" with the build process itself, but they don't have to be coupled *that* closely. (I guess it's possible for the build to mutate the environment we capture, but it's probably *very uncommon* for the build itself to obsolete the environment we calculate prebuild.)

cargo-dist/src/tasks.rs
```diff
@@ -2112,6 +2120,8 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
             warn!("The Homebrew publish job is enabled but no tap was specified\n  consider setting the tap field in Cargo.toml");
         }
 
+        let runtime_conditions = pick_runtime_conditions(&self.manifest.systems);
```
Gankra, line 2123: This can now be replaced with 

```
release.platform_support.conflated_runtime_conditions()
```

And stuff like glibc_thunderdome I think are (alas) obsoleted?

Gankra (changes requested):
New stuff I think looks good, but some old glue can be dropped.

cargo-dist/src/linkage.rs
mistydemeo: Ensured this returns an indeterminate value instead of an error if we failed to collect information or somehow ran cross-platform.

Gankra (approved):

mistydemeo:
Ran some tests and confirmed this is working. For example, [this release](https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.214) was run on an `ubuntu-latest` runner, which is 22.04 and which bundles glibc 2.35. The shell and npm installers correctly report that as the minimum version, like we want them to. Meanwhile, [this release](https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.215) uses our default ubuntu runner with glibc 2.31, and so that version gets inlined into the installers.

PR #1213, opened 2024-07-17
fix: restore proper diagnostics for repository url

Gankra:
In various refactors this code got overly permissive, silently allowing 'no repository url set at all' in cases where it's invalid. Later code would then panic with an obtuse message about it being impossible (because it should be).

### DIFF
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -128,10 +128,22 @@ pub enum DistError {
         #[diagnostic_source]
         inner: AxoprojectError,
     },
+
     /// User tried to enable Github CI support but no url for the repo
     #[error("Github CI support requires you to specify the URL of your repository")]
-    #[diagnostic(help(r#"Set the repository = "https://github.com/..." key in your Cargo.toml"#))]
+    #[diagnostic(help(
+        r#"Set the repository = "https://github.com/..." key in the Cargo.toml of your app"#
+    ))]
     CantEnableGithubNoUrl,
+
+    /// We got a repository URL but couldn't interpret it as a GitHub repo
+    #[error("Github CI support requires your repository be hosted on Github")]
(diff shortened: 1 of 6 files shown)


Gankra:
Before:

image


After:

image

Gankra:
This also just turns the cases where we did properly warn into hard errors, because the warnings are easy to miss, and there really is no proceeding pass that point.

cargo-dist/src/errors.rs
```diff
@@ -128,10 +128,22 @@ pub enum DistError {
         #[diagnostic_source]
         inner: AxoprojectError,
     },
+
     /// User tried to enable Github CI support but no url for the repo
     #[error("Github CI support requires you to specify the URL of your repository")]
-    #[diagnostic(help(r#"Set the repository = "https://github.com/..." key in your Cargo.toml"#))]
+    #[diagnostic(help(
+        r#"Set the repository = "https://github.com/..." key in the Cargo.toml of your app"#
```
ashleygwilliams, line 135: how precise can we get here.... could we know which app they were trying to release and therefore which cargo.toml was missing the repo. given how frequent and blocking this error is i think the extra effort would have a huge impact

cargo-dist/src/errors.rs
```diff
@@ -128,10 +128,22 @@ pub enum DistError {
...
+    ))]
     CantEnableGithubNoUrl,
+
+    /// We got a repository URL but couldn't interpret it as a GitHub repo
+    #[error("Github CI support requires your repository be hosted on Github")]
```
ashleygwilliams, line 140: we allow github CI + axo hosting... i think "hosting" might be confusing here (i know you mean the repo and not the artifacts but a drive by read could confuse this)

Gankra:
afterer 
image

ashleygwilliams (approved):

Issue #179, opened 2023-03-02
Support for distributing dynamic library


nurmohammed840:

ashleygwilliams:
@nurmohammed840 would love more info on this request if you have the time to share more :)

Michael-F-Bryan:
I came here because the thing I'm working on includes both a CLI (in a `foo-cli` crate) and a C API (in a `foo-c-api` crate). 

@ashleygwilliams when distributing a C API you'll normally want to include a couple things:

- Header file (can be considered a static file in your repo for all intents and purposes)
- Dynamic library (i.e. use the `cdylib` crate type to generate `libfoo.so` or `libfoo.dll`)
- Static library (i.e. use the `staticlib` crate type to generate `libfoo.a` or `libfoo.lib`)

penberg:
I would love to use `cargo dist` to distribute a static library as well: https://github.com/penberg/limbo/issues/162

mistydemeo:
We're in the process of shipping this! It should be in the next release.

mistydemeo:
As of PR #1214, we'll support static libraries too.

penberg:
Hey @mistydemeo! I see `install-libraries` in https://opensource.axo.dev/cargo-dist/book/reference/config.html?highlight=staticlib#install-libraries. However, I don't think it's available in 0.19.0 -- a documentation bug?

Gankra:
Ah yes, documentation bug -- it's slated for 0.20.0, but it *is* largely implemented already.

The current work can be found in this prerelease, and should works as documented:

https://github.com/axodotdev/cargo-dist/releases/tag/v0.20.0-prerelease.3

----

Before we ship the proper release it we need some clarification on what installers you'd like to be using, on which platforms, and if there's any specific platform-specific behaviour you expect/require. I think this might have gotten discussed in some other places but it would be good to write them down somewhere.

Current behaviour is:

* the tarball packs all the files flat and together
* the homebrew installer unpacks each kind of file to the appropriate package manager bin/lib dirs
* the shell/powershell installers dump the files flat into the same dir on PATH
  * this is *presumably* undesirable and should get adjusted to some more appropriate per-platform/configurable behaviour, but **unclear which knobs/defaults we want**
* ??? with the msi installer unclear it even knows about the files

mistydemeo:
I've corrected the minimum version in the docs; sorry for the confusion!

penberg:
Hey @Gankra! I am just using the shell installer:

[...]

PR #1214, opened 2024-07-17
Build and install static libs

mistydemeo:
This is similar to #1182, extending that feature to support static libraries as well. It also contains a few other changes/improvements:

* The syntax for `package_cdylibs` and `install_cdylibs` has changed; this is now `package_libraries` and `install_libraries`, and it accepts arrays of strings in order to be able to specify `cdylib` and `cstaticlib` independently of each other.
* Two separate bugs have been fixed which prevented us from being able to reason about two binaries with overlapping names. That bug previously meant that a binary named `foo.exe` and a library named `libfoo.dylib` would be mistaken for each other, with one definition overwriting the other one internally. I caught it when I discovered that `libfoo.a` and `libfoo.dylib` couldn't be packaged at the same time, but it also affected executables and cdylibs.

I intentionally allowed some duplication in the installer code with how it handles staticlibs to leave ourselves open to install them to a separate path from dylibs at some point, if that's ever wanted.

Note that while Rust uses `cdylib` and `staticlib`, inconsistently prefixing with `c` or not, we use `cdylib` and `cstaticlib` to be more precise. This is true both internally and in the public-facing terminology.

Covers https://github.com/axodotdev/cargo-dist/issues/179#issuecomment-2232464351.

### DIFF
--- a/cargo-dist/src/announce.rs
+++ b/cargo-dist/src/announce.rs
@@ -34,6 +34,7 @@ pub(crate) struct ReleaseArtifacts {
     pub package_idx: PackageIdx,
     pub executables: Vec<String>,
     pub cdylibs: Vec<String>,
+    pub cstaticlibs: Vec<String>,
 }
 
 /// Settings for `select_tag`
@@ -179,14 +180,14 @@ fn check_dist_package(
         .unwrap_or(false);
 
     let package_empty = if package_cdylibs {
-        pkg.binaries.is_empty() && pkg.cdylibs.is_empty()
+        pkg.binaries.is_empty() && pkg.cdylibs.is_empty() && pkg.cstaticlibs.is_empty()
     } else {
         pkg.binaries.is_empty()
     };
 
     // Nothing to publish if there's no binaries/libraries!
     if package_empty {
-        return Some("no binaries or cdylibs".to_owned());
+        return Some("no binaries or cdylibs/cstaticlibs".to_owned());
     }
 
     // If [metadata.dist].dist is explicitly set, respect it!
@@ -395,13 +396,22 @@ fn select_packages(
                 cdylibs.push(library.to_owned());
             }
         }
+        let mut cstaticlibs = vec![];
+        for library in &pkg.cstaticlibs {
+            info!("    {}", sty.apply_to(format!("[cstaticlib] {}", library)));
+            // In the future might want to allow this to be granular for each library
+            if disabled_reason.is_none() {
+                cstaticlibs.push(library.to_owned());
+            }
+        }
 
         // If any binaries were accepted for this package, it's a Release!
-        if !binaries.is_empty() || !cdylibs.is_empty() {
+        if !binaries.is_empty() || !cdylibs.is_empty() || !cstaticlibs.is_empty() {
             let release = ReleaseArtifacts {
                 package_idx: pkg_id,
                 executables: binaries,
                 cdylibs,
+                cstaticlibs,
             };
             releases.push(release);
         }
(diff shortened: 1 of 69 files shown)


cargo-dist/src/announce.rs
```diff
@@ -173,20 +177,25 @@ fn check_dist_package(
...
-        pkg.binaries.is_empty() && pkg.cdylibs.is_empty()
-    } else {
-        pkg.binaries.is_empty()
-    };
+    let mut conditions = vec![pkg.binaries.is_empty()];
+    if package_libraries.contains(&LibraryStyle::CDynamic) {
+        conditions.push(pkg.cdylibs.is_empty());
+    }
+    if package_libraries.contains(&LibraryStyle::CStatic) {
+        conditions.push(pkg.cstaticlibs.is_empty());
```
Gankra, line 191: I was very enamoured with how clever this was but just realized you can replace every `push` with just `&=`, which is probably clearer/simpler?

cargo-dist/src/announce.rs
```diff
@@ -173,20 +177,25 @@ fn check_dist_package(
...
+    }
+
+    let package_empty = conditions.into_iter().all(identity);
 
     // Nothing to publish if there's no binaries/libraries!
     if package_empty {
-        return Some("no binaries or cdylibs".to_owned());
+        return Some("no binaries or cdylibs/cstaticlibs".to_owned());
```
Gankra, line 198: Hmm could we build up this string in the above conditions...

cargo-dist/src/build/mod.rs
```diff
@@ -249,35 +253,6 @@ impl BuildExpectations {
     }
 }
 
-/// Returns the base filename for `src_path`, with file extension
-/// stripped and, for .so/.dylib libraries, `lib` prefix too
-fn bin_basename(src_path: &Utf8PathBuf) -> Option<String> {
```
Gankra, line 254: you love to see it

cargo-dist/src/config.rs
```diff
@@ -886,6 +888,43 @@ impl std::str::FromStr for CiStyle {
...
+}
+
+impl std::str::FromStr for LibraryStyle {
+    type Err = DistError;
+    fn from_str(val: &str) -> DistResult<Self> {
+        let res = match val {
+            "cdylib" => Self::CDynamic,
+            "cstaticlib" => Self::CStatic,
+            s => {
+                return Err(DistError::UnrecognizedCiStyle {
```
Gankra, line 919: whoops forgot to make a new error variant

cargo-dist/src/tasks.rs
```diff
@@ -1286,11 +1293,11 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
 
-        let package_cdylibs = package_cdylibs.unwrap_or(false);
-        let install_cdylibs = if !package_cdylibs {
-            false
+        let package_libraries = package_libraries.clone().unwrap_or(vec![]);
+        let install_libraries = if package_libraries.is_empty() {
+            vec![]
         } else {
-            install_cdylibs.unwrap_or(false)
+            install_libraries.clone().unwrap_or(vec![])
```
Gankra, line 1300: technically this can be `unwrap_or_default`

Gankra (approved):
Some little nits but the gist looks great

Gankra:
(disclaimer i kinda glazed over in the installer templates)

cargo-dist/src/tasks.rs
mistydemeo: Surprised clippy didn't yell at me for this.

cargo-dist/src/announce.rs
mistydemeo: This message only shows up in the `--verbose=info` output, but might as well

PR #1215, opened 2024-07-17
chore: rework RuntimeConditions

Gankra:
This turns RuntimeConditions into a flat struct with Options, reflecting the fact that a list of enums is really hard to work with and can express confusing ideas.

This also makes PlatformSupport::new compute RuntimeConditions from the current system and linkage info, working bottom-up from binary conditions up to archive and platform conditions.

It also adds a new bridging API conflated_runtime_conditions to be paired with ragments users.

The implementation is currently unused and missing a few key parts at the bottom for glibc version, which largely exist in Misty's PR.

### DIFF
--- a/cargo-dist/src/platform.rs
+++ b/cargo-dist/src/platform.rs
@@ -2,7 +2,8 @@
 use axoproject::platforms::{
     TARGET_ARM64_MAC, TARGET_ARM64_WINDOWS, TARGET_X64_MAC, TARGET_X64_WINDOWS, TARGET_X86_WINDOWS,
 };
-use cargo_dist_schema::ArtifactId;
+use cargo_dist_schema::{ArtifactId, AssetId, DistManifest, Linkage, SystemInfo};
+use serde::Serialize;
 
 use crate::{
     backend::installer::{ExecutableZipFragment, UpdaterFragment},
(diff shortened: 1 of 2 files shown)


Gankra:
A caveat about this design (which I believe is shared with Misty's impl too):  when doing `cargo dist build` locally, which does local and global builds together, the info the installer is handed is "stale". Really all the installers should compute `PlatformInfo::new()` *right before* applying the template to get the freshest possible linkage info!

This is a known migration step for "fully" adopting the PlatformInfo abstraction (and not the reduced precision `fragments` adapter API).

cargo-dist/src/platform.rs
```diff
@@ -417,3 +436,108 @@ fn supports(
...
+        let RuntimeConditions {
+            min_glibc_version,
+            min_musl_version,
+            rosetta2,
+        } = other;
+
+        self.min_glibc_version =
+            max_of_min_libc_versions(&self.min_glibc_version, min_glibc_version);
+        self.min_musl_version = max_of_min_libc_versions(&self.min_musl_version, min_musl_version);
+        self.rosetta2 &= rosetta2;
```
Gankra, line 451:
```suggestion
        self.rosetta2 |= rosetta2;
```

cargo-dist/src/platform.rs
```diff
@@ -417,3 +436,108 @@ fn supports(
...
+        if lib.is_glibc {
+            // TODO: misty's branch has this
+            if let Some(system_glibc) = system.glibc_version {
+                // If there's a system libc, assume that's what it was built against
+                return Some(system_glibc);
+            } else {
+                // If the system has no known libc version use Ubuntu 20.04's glibc as a guess
+                return Some(LibcVersion {
+                    major: 2,
+                    series: 11,
```
mistydemeo, line 531: We should make this the same as our current hardcoded one from the shell template. It'd also be nice to make this a method on `LibcVersion` so we can keep track of where we've inlined it, eg `LibcVersion::default_glibc()` or something.

```suggestion
                    series: 31,
```

cargo-dist/src/platform.rs
```diff
@@ -417,3 +436,108 @@ fn supports(
...
+
+/// Combine two min_libc_versions to get a new min that satisfies both
+fn max_of_min_libc_versions(
+    lhs: &Option<LibcVersion>,
+    rhs: &Option<LibcVersion>,
+) -> Option<LibcVersion> {
+    match (*lhs, *rhs) {
+        (None, None) => None,
+        (Some(ver), None) | (None, Some(ver)) => Some(ver),
+        (Some(lhs), Some(rhs)) => Some(lhs.max(rhs)),
```
mistydemeo, line 463: Does the default `ord` derive handle this for us or do we have to implement our own so it knows how to compare both values?

mistydemeo (commented):
Nice, I think this is going to pair well with the other branch.

Gankra:
now in misty's pr

PR #1221, opened 2024-07-22
fix: add missing triples to linkage check

### LINKED ISSUE #1219: Can't build zip on Windows because unable to run linkage report
On Windows, when I run `cargo dist build --artifacts local --target x86_64-pc-windows-gnu`, I get this:
```console
building artifacts:
  myapp-x86_64-pc-windows-gnu.zip
  myapp-x86_64-pc-windows-gnu.zip.sha256

building cargo target (x86_64-pc-windows-gnu/dist --workspace)

    Finished dist [optimized] target(s) in 2.35s
  x unable to run linkage report for this type of binary
```
It fails with this error "unable to run linkage report", because the target x86_64-pc-windows-gnu is not supported for this I guess, and I don't get any zip being built :-( .
But how can I skip/ignore this linkage report step and still build the zip anyway ?

mistydemeo:
We had the -msvc triples here, but were missing the -gnu ones.

Fixes #1219.

### DIFF
--- a/cargo-dist/src/linkage.rs
+++ b/cargo-dist/src/linkage.rs
@@ -393,9 +393,12 @@ pub fn determine_linkage(path: &Utf8PathBuf, target: &str) -> DistResult<Linkage
             do_ldd(path)?
         }
         // Can be run on any OS
-        "i686-pc-windows-msvc" | "x86_64-pc-windows-msvc" | "aarch64-pc-windows-msvc" => {
-            do_pe(path)?
-        }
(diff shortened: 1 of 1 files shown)


Gankra:
We should increasingly be considering using https://docs.rs/target-lexicon/0.12.15/target_lexicon/struct.Triple.html

Gankra (approved):

mistydemeo:
Good call, I think. I'll merge this for now though.

PR #1223, opened 2024-07-22
feat: add Repo struct

mistydemeo:
This captures information about the git repository we're running builds in, including its absolute path on disk and its current git HEAD.

We already used the side effects of capturing this information in the past; this lets us capture and hold on to it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,6 +184,7 @@ name = "axoproject"
 version = "0.20.0-prerelease.3"
 dependencies = [
  "axoasset",
+ "axoprocess",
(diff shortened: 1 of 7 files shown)


mistydemeo:
Pushed a version that moves this into axoproject. I've also added a second `WorkspaceGraph` constructor that works from a git repo rather than a clamp_to_dir param, and deduped the repo detection by getting cargo-dist's config to use that.

Gankra (approved):
looks good but can you tweak DistGraph::repo_dir to use this?

PR #1243, opened 2024-07-27
Change `if_expr` to `if` in templates

FreeMasen:
Objects in templates will already be renamed so `if_expr` will always be `undefined` when outputting templates.


mistydemeo (approved):
Good catch, thank you!

PR #1252, opened 2024-07-30
feat: allow init to migrate config

mistydemeo:
Now that we support generic workspaces that include Cargo members, some users may prefer having their cargo-dist config in a cargo-dist-specific format instead of hitching a ride in Cargo.toml. This provides an option to migrate configuration.

At the start of init, check if we read config from Cargo.toml and offer to use dist-workspace.toml. If the user says yes, and we already had configuration in Cargo.toml, finish off the init process by deleting the old content from Cargo.toml.

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -43,6 +43,15 @@ struct MultiDistMetadata {
     packages: SortedMap<String, DistMetadata>,
 }
 
+fn theme() -> dialoguer::theme::ColorfulTheme {
+    dialoguer::theme::ColorfulTheme {
+        checked_item_prefix: console::style("  [x]".to_string()).for_stderr().green(),
+        unchecked_item_prefix: console::style("  [ ]".to_string()).for_stderr().dim(),
+        active_item_style: console::Style::new().for_stderr().cyan().bold(),
+        ..dialoguer::theme::ColorfulTheme::default()
+    }
+}
+
 /// Run 'cargo dist init'
 pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
     let workspaces = config::get_project()?;
@@ -68,7 +77,42 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
     }
 
     // Load in the root workspace toml to edit and write back
-    let mut workspace_toml = config::load_cargo_toml(&root_workspace.manifest_path)?;
+    let workspace_toml = config::load_cargo_toml(&root_workspace.manifest_path)?;
+    let mut original_workspace_toml = workspace_toml.clone();
+    let initted = has_metadata_table(root_workspace);
+
+    // If this is a Cargo.toml, offer to either write their config to
+    // a dist-workspace.toml, or migrate existing config there
+    let mut is_migrating = false;
+    // Users who haven't initted yet should be opted into the
+    // new config format by default.
+    let desired_workspace_kind = if root_workspace.kind == WorkspaceKind::Rust && !initted {
+        WorkspaceKind::Generic
+    // Already-initted users should be asked whether to migrate.
+    } else if root_workspace.kind == WorkspaceKind::Rust {
+        let prompt = r#"Would you like to keep your configuration in Cargo.toml?
+    Future versions of cargo-dist will feature major changes to the
+    configuration format, including a new cargo-dist-specific configuration file.
+    Answering "no" here will opt you into the new format early."#;
+        let res = if args.yes {
+            // We want to avoid --yes pulling it in at this point.
+            true
+        } else {
+            dialoguer::Confirm::with_theme(&theme())
+                .with_prompt(prompt)
+                .default(true)
+                .interact()?
+        };
+
+        if !res {
+            is_migrating = true;
+            WorkspaceKind::Generic
+        } else {
+            root_workspace.kind
+        }
+    } else {
+        root_workspace.kind
+    };
 
     let multi_meta = if let Some(json_path) = &args.with_json_config {
         // json update path, read from a file and apply all requested updates verbatim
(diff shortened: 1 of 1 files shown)


cargo-dist/src/init.rs
```diff
@@ -69,6 +79,32 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
     let mut workspace_toml = config::load_cargo_toml(&root_workspace.manifest_path)?;
+    let mut original_workspace_toml = workspace_toml.clone();
+    let initted = has_metadata_table(root_workspace);
+
+    // If this is a Cargo.toml, offer to either write their config to
+    // a dist-workspace.toml, or migrate existing config there
+    let mut is_migrating = false;
+    let desired_workspace_kind = if root_workspace.kind == WorkspaceKind::Rust {
+        let prompt = if initted {
+            r#"Would you like to migrate your project from Cargo.toml to dist-workspace.toml?"#
```
mistydemeo, line 90: We'll probably either want this to be really verbose, or have a docs page to link to.

cargo-dist/src/init.rs
```diff
@@ -84,15 +128,59 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
+        // Remove Cargo-specific workspace
+        workspace_toml.remove("workspace");
+
+        // Write generic workspace config
+        let mut table = toml_edit::table();
+        if let Some(t) = table.as_table_mut() {
+            let mut array = toml_edit::Array::new();
+            array.push("cargo:.");
+            t["members"] = toml_edit::value(array);
+            t.set_position(0);
```
Gankra, line 144: what on earth does this do

Gankra: OHHH

ok yeah maybe a comment to explain this is adding the Cargo.toml as a member of the workspace haha (also i wonder if `./` will be more legible?

cargo-dist/src/init.rs
```diff
@@ -84,15 +128,59 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
+        }
+        workspace_toml.insert("workspace", table);
+
+        // These fields will continue to be read from Cargo.toml
+        workspace_toml.remove("dependencies");
+        workspace_toml.remove("dev-dependencies");
+        workspace_toml.remove("package");
+        workspace_toml.remove("profile");
```
Gankra, line 152: We definitely can't play whackamole here. We know what we want is under `workspace.metadata.dist` and/or `package.metadata.dist`, why can't we just precisely grab that?

cargo-dist/src/init.rs
```diff
@@ -84,15 +128,59 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
-        apply_dist_to_workspace_toml(&mut workspace_toml, root_workspace.kind, meta);
+        apply_dist_to_workspace_toml(&mut workspace_toml, desired_workspace_kind, meta);
     }
 
     eprintln!();
 
+    let filename = match desired_workspace_kind {
+        WorkspaceKind::Rust => "Cargo.toml",
+        WorkspaceKind::Generic => "dist-workspace.toml",
+        WorkspaceKind::Javascript => todo!(),
```
Gankra, line 164: Also dist-workspace.toml, technically... Javascript is in a weird place in the code -- ends up being defacto generic at the top-level. Generally only shows up internally as a nested workspace (not your problem, just thinking out loud).

cargo-dist/src/init.rs
```diff
@@ -84,15 +128,59 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
+    };
+    let destination = root_workspace.workspace_dir.join(filename);
+
     // Save the workspace toml (potentially an effective no-op if we made no edits)
-    config::save_cargo_toml(&root_workspace.manifest_path, workspace_toml)?;
-    eprintln!("{check} added [workspace.metadata.dist] to your root Cargo.toml");
+    config::save_cargo_toml(&destination, workspace_toml)?;
+    eprintln!("{check} added [workspace.metadata.dist] to your root {filename}");
```
Gankra, line 170: This label is incorrect now (it's just `dist` in the dist-workspace.toml, right?)

Gankra (changes requested):
Also needs to handle recursive splitting of member tomls to dist.toml.

cargo-dist/src/init.rs
mistydemeo: Oh yes, good point. Not sure why I didn't think in terms of "grab what we want" instead of "delete what we don't want".

mistydemeo:
> Also needs to handle recursive splitting of member tomls to dist.toml.

Done.

cargo-dist/src/init.rs
```diff
@@ -84,15 +130,117 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
+    // If we're migrating, the configuration will be missing the
+    // generic workspace specification, and will have some
+    // extraneous cargo-specific stuff that we don't want.
+    let mut workspace_toml = if is_migrating {
+        // Fetch just the fields we want
+        let dist = workspace_toml
+            .get("workspace")
+            .and_then(|t| t.get("metadata"))
+            .and_then(|t| t.get("dist"))
+            .unwrap();
```
Gankra, line 142: I don't think this can/should be an unwrap. I think the migrate code can run on a non-initted workspace rn?

Gankra (approved):
this *looks* right but damn if it ain't a bit byzantine (as all the init code is).

probably a situation of "need to kick the tires on it" to work out any details i'm missing in review

cargo-dist/src/init.rs
mistydemeo: Oh yes, sorry, missed that from debugging. Will clean up.

cargo-dist/src/init.rs
mistydemeo: I was wondering about that, since we have the package.json-only config stuff. I'll add it to the Generic branch.

PR #1253, opened 2024-07-30
init: ensure terminal is restored when interrupted

mistydemeo:
If we receive ctrl-c during `init`, dialoguer leaves the cursor hidden and the user may not understand why that's happened. We need to make sure we restore the cursor before terminating the application.

While this may be handled by dialoguer/console at some point, this would make for a nicer user experience today.

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -67,7 +67,12 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
 
         // Immediately re-exit the process with the same
         // exit code the unhandled ctrl-c would have used
-        std::process::exit(130);
+        let exitstatus = if cfg!(windows) {
+            0xc000013a_u32 as i32
+        } else {
+            130
+        };
+        std::process::exit(exitstatus);
     });
 
     let workspaces = config::get_project()?;
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -54,6 +54,22 @@ fn theme() -> dialoguer::theme::ColorfulTheme {
 
 /// Run 'cargo dist init'
 pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
+    // on ctrl-c,  dialoguer/console will clean up the rest of its
+    // formatting, but the cursor will remain hidden unless we
+    // explicitly go in and show it again
+    // See: https://github.com/console-rs/dialoguer/issues/294
+    let ctrlc_handler = tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.unwrap();
+
+        let term = console::Term::stdout();
+        // Ignore the error here if there is any, this is best effort
+        let _ = term.show_cursor();
+
+        // Immediately re-exit the process with the same
+        // exit code the unhandled ctrl-c would have used
+        std::process::exit(130);
+    });
+
     let workspaces = config::get_project()?;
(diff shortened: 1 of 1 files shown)


cargo-dist/src/init.rs
```diff
@@ -45,6 +45,22 @@ struct MultiDistMetadata {
...
+    let ctrlc_handler = tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.unwrap();
+
+        let term = console::Term::stdout();
+        // Ignore the error here if there is any, this is best effort
+        let _ = term.show_cursor();
+
+        // Immediately re-exit the process with the same
+        // exit code the unhandled ctrl-c would have used
+        std::process::exit(130);
```
Gankra, line 61: Does this make sense on windows..?

Gankra: Also process::exit is always a maintenance hazard, since it prevents our own app from running its destructors. I suppose in this case that might be fine / this is the function to do anything like that. Just, noting it for future.

Gankra (approved):
tentatively approving but some questions

cargo-dist/src/init.rs
```diff
@@ -45,6 +45,22 @@ struct MultiDistMetadata {
 
 /// Run 'cargo dist init'
 pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
+    // on ctrl-c,  dialoguer/console will clean up the rest of its
+    // formatting, but the cursor will remain hidden unless we
+    // explicitly go in and show it again
+    // See: https://github.com/console-rs/dialoguer/issues/294
+    let ctrlc_handler = tokio::spawn(async move {
+        tokio::signal::ctrl_c().await.unwrap();
```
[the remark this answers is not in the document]
mistydemeo: My understanding is there isn't a builtin way to handle it, so I reached for the crate that's already in our dep tree. There's also a standalone `ctrlc` crate, but it's missing the ability to clear handlers after they're no longer needed.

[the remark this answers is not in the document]
Gankra: works for me

cargo-dist/src/init.rs
mistydemeo: Looks like no, Windows uses something extremely different. (I'm seeing `-1073741510` in both `cmd` and powershell, which I guess is `0xc000013a`.) Also looks like the hiding the cursor behaviour is different between `cmd` and powershell...?

cargo-dist/src/init.rs
```diff
@@ -58,7 +58,12 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
 
         // Immediately re-exit the process with the same
         // exit code the unhandled ctrl-c would have used
-        std::process::exit(130);
+        let exitstatus = if std::env::consts::OS == "windows" {
```
Gankra, line 61: 0.o i have never seen this before (more conventional would be `if cfg!(windows`) {`

mistydemeo: No idea why I reached for `std::env::consts::OS` here 😅

cargo-dist/src/init.rs
mistydemeo: > Also process::exit is always a maintenance hazard, since it prevents our own app from running its destructors. I suppose in this case that might be fine / this is the function to do anything like that. Just, noting it for future.

Yeah, the thing is I figured this is basically equivalent to what ctrl-c was doing here already.

Issue #830, opened 2024-03-06
Don't modify .bashrc when installation directoy is already in PATH


konstin:
My `.bashrc` contains the following:

```bash
export PATH="$HOME/.local/bin:$HOME/.cargo/bin/:$PATH"
```

After running `curl -LsSf https://astral.sh/uv/install.sh | sh`, it contains:

```bash
export PATH="$HOME/.local/bin:$HOME/.cargo/bin/:$PATH"

. "$HOME/.cargo/env"
```

Could the cargo-dist installer check if the target directory is already in PATH instead of changing the PATH precedence?

duckinator:
This should've been fixed in v0.21.0, via #1264.

PR #1264, opened 2024-08-01
Don't modify dotfiles when install dir is already in PATH

konstin:
This is a proof-of-concept for https://github.com/astral-sh/uv/issues/5576 and https://github.com/axodotdev/cargo-dist/issues/830: When the install dir is already in PATH, don't modify the user's dotfile.

For my use case, i set `export PATH="$HOME/.cargo/bin/:$PATH"` already and don't want to source arbitrary other files; I've had bad experiences with tools that would make intrusive changes to my shell by souring their slow script on every bash startup, and a single file configuration is much easier to debug.

Caveat: I have no idea how portable this code is.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -479,6 +479,12 @@ install() {
 
     say "{{ install_success_msg }}"
 
+    # Avoid modifying the users PATH if they are managing their PATH manually
+    case :$PATH:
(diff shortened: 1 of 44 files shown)


mistydemeo:
Thank you! I gave this a try in dash, and confirmed it works. Should be compatible enough for us. I'll add the test snapshots.

mistydemeo (approved):

PR #1266, opened 2024-08-01
docs(quickstart): add generic quickstart guide

mistydemeo:
This adds a new generic build quickstart guide, analogous to the existing Cargo-focused quickstart guide. I tried to keep it simple to follow, so I pared down some of the details from the original guide that I felt were a little too fiddly to present to first-time users. My goal was to get people from 0 to their first release as conveniently as possible.


mistydemeo:
Rebased on #1271. I've also adapted most of the language from the Rust version.

Gankra (approved):

PR #1267, opened 2024-08-01
feat: make cargo optional

Gankra:
This PR removes the runtime assumptions in the code that cargo/rust is installed and can be used to get info about the system and local cargo projects we might be tasked with building. When complete we will still require cargo to be installed if your project actually includes cargo packages, but e.g. javascript-only users won't need to randomly setup cargo.

* [x] allow cargo-dist to be invoked without extra 'dist' arg from cargo
* [x] add dist alias
* [x] add npm installer
  * ~~[ ] disable self-installer tests for a while to get over hump of changing to .tar.gz~~
  * [x] ...or fix the npm installer archive limitation
* [x] make Tools::cargo optional
  * [x] make code get host target differently
  * [x] make code error out only when you have a rust workspace
  * [x] make axoproject not consider it an error for cargo to be missing 
  * [x] give axoproject some logic to notice there is a Cargo.toml and complain you don't have cargo installed?


Gankra:
Note that this doesn't actually make us fully cargo-independent, as there are randomly places in the code that assume cargo unconditionally must exist. While there aren't a ton of these and it's not a huge uplift, I am not willing to push that extra complexity through right now.

Gankra:
ugh making the installers expect .tar.gz breaks the updater/installer self-tests which find .tar.xz on the current releases. and you need to go to .tar.gz for npm. 

handling this requires disabling these tests temporarily, until we have releases that use the new archive format.

mistydemeo:
At this point we have essentially everything except the very specific case of "warn explicitly if there's a Cargo.toml but you have no cargo". That *might* actually be covered by the existing checks here, but I want to test.

I was able to build a C-based project without having Cargo installed on my system.

mistydemeo:
From this point:

> make code get host target differently

I've given us *two* host targets: the one identifying the system we're on, and the one identifying what Cargo wants to build for. They can, hypothetically, be different values (though usually they won't be). When we're actually building a Cargo project and checking if we're cross-compiling, we check the Cargo value; otherwise, we check the one we determined for the system, and that doesn't require Cargo to be present.

cargo-dist/Cargo.toml
```diff
@@ -15,7 +15,8 @@ exclude = [
 ]
 
 [[bin]]
-name = "cargo-dist"
+name = "dist"
+path = "src/main.rs"
```
ashleygwilliams, line 19: this is necessary because the package name is still cargo-dist ya?

cargo-dist/tests/gallery/dist.rs
```diff
@@ -27,8 +27,8 @@ mod tools;
 #[cfg(any(target_family = "unix", target_family = "windows"))]
 const ENV_RUIN_ME: &str = "RUIN_MY_COMPUTER_WITH_INSTALLERS";
 /// Set this at runtime to override STATIC_CARGO_DIST_BIN
-const ENV_RUNTIME_CARGO_DIST_BIN: &str = "OVERRIDE_CARGO_BIN_EXE_cargo-dist";
-const STATIC_CARGO_DIST_BIN: &str = env!("CARGO_BIN_EXE_cargo-dist");
+const ENV_RUNTIME_CARGO_DIST_BIN: &str = "OVERRIDE_CARGO_BIN_EXE_dist";
+const STATIC_CARGO_DIST_BIN: &str = env!("CARGO_BIN_EXE_dist");
```
ashleygwilliams, line 31: should we update these env var names?

cargo-dist/Cargo.toml
mistydemeo: That's correct.

README.md
```diff
@@ -1,16 +1,16 @@
 <div class="oranda-hide">
 
-# `cargo-dist`
+# `dist`
```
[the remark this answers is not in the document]
mistydemeo: Oh, good call!

cargo-dist/tests/gallery/dist.rs
mistydemeo: Done!

ashleygwilliams (approved):

PR #1285, opened 2024-08-07
chore: bump min macos version to 13

Gankra:
12 is now deprecated and slated to be removed in winter


mistydemeo (approved):
We're drifting perilously close to the ARM64-only horizon, huh.

Issue #1297, opened 2024-08-08
linux appimage support?


Ayush1325:
Mostly relevant for GUI apps.

Current ways to package rust apps as appimage:
- https://crates.io/crates/cargo-appimage
- https://crates.io/crates/cargo-packager/0.10.1

ashleygwilliams:
hey @Ayush1325 ! sorry for the delay and thanks so much for kicking this off!! i am gonna talk about this with the team tomorrow and we'll get back to you by early next week at the latest (though hopefully sooner!) with next steps.

PR #1303, opened 2024-08-09
Add initial appimage support

Ayush1325:
- Uses appimagetool to create the appimage from AppDir folder
- Generte destkop entry file on the fly.
- Use AppRun from AppImageKit
- Use empty icon.

This is basically a MVP implementation. Some things that need discussion:
1. How to handle linux desktop entries since they will be used in most other linux formats.
2. How to handle icons. Appimage only seems to support PNG icons, while Windows uses ICO, so we need to decide on something cross-platform.
3. How to handle bundling platform deps. Basically needs an exclude list of some kind.

Fixes: https://github.com/axodotdev/cargo-dist/issues/1297

### DIFF
--- a/cargo-dist/src/backend/installer/appimage.rs
+++ b/cargo-dist/src/backend/installer/appimage.rs
@@ -1,0 +1,110 @@
+//! Appimage installer
+
+use std::{fs::Permissions, io::BufRead, os::unix::fs::PermissionsExt};
+
+use axoproject::Version;
+use camino::Utf8PathBuf;
+use itertools::Itertools;
+use serde::Serialize;
+
+use crate::{backend::templates::TEMPLATE_LINUX_DESKTOP, DistGraph, DistResult};
+
+#[derive(Debug, Clone)]
+/// Info needed to build an Appimage
+pub struct AppImageInfo {
+    /// An ideally unambiguous way to refer to a package for the purpose of cargo -p flags.
+    pub pkg_spec: String,
+    /// Final file path of the msi
+    pub file_path: Utf8PathBuf,
+    /// Dir stuff goes to
+    pub package_dir: Utf8PathBuf,
+    /// Binaries we'll be baking into the msi
+    pub target: String,
+    /// Path to the package Cargo.toml associated with this msi
+    pub version: Version,
+}
+
+#[derive(Serialize)]
+struct DesktopEntry {
+    pub pkg: String,
+    pub version: String,
+}
+
+impl AppImageInfo {
+    /// Build the appimage installer
+    pub fn build(&self, dist: &DistGraph) -> DistResult<()> {
+        let bin_path = self.package_dir.join("usr").join("bin");
+        std::fs::create_dir_all(&bin_path)?;
+        let app_bin_path = bin_path.join(&self.pkg_spec);
+
+        std::fs::rename(self.package_dir.join(&self.pkg_spec), &app_bin_path)?;
+
+        let output = std::process::Command::new("ldd")
+            .arg(app_bin_path)
+            .output()?;
+
+        assert!(output.status.success());
+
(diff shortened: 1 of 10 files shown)


mistydemeo:
Thank you very much for the contribution! We appreciate it. Before getting too deep into the code, I've got a few questions about your intentions and the implementation.

Do you have a specific project you're looking to distribute this way? We usually add features when we know they're being used by someone - that way, we can know for sure that it works, and have something we're comfortable testing against.

What's the way you'd want `appimagetool` to be installed for GitHub Actions jobs? It looks like this requires at least the appimage CLI tool, and that's not included by default on the Actions runners. (It looks like it might not be in the Ubuntu package manager either.)

Ayush1325:
> Thank you very much for the contribution! We appreciate it. Before getting too deep into the code, I've got a few general questions about your intentions and the implementation.
> 
> Do you have a specific project you're looking to distribute this way? We usually add features when we know they're being used by someone - that way, we can know for sure that it works, and have something we're comfortable testing against.

Yes, I am currently using it for building executables for [bb-imager-rs](https://openbeagle.org/ayush1325/bb-imager-rs). I am currently using cargo-appimage for building appimages but it would be great if I could switch to cargo-dist.


> What's the way you'd want `appimagetool` to be installed for GitHub Actions jobs? It looks like this requires at least the appimage CLI tool, and that's not included by default on the Actions runners. (It looks like it might not be in the Ubuntu package manager either.)

I am using Gitlab as seen from the link, so well, can't really comment on Github side of things. Ideally, I would prefer not to need an external tool, but well, not sure if there is a Rust crate or something we could use. Additionally, I would like to inform everyone that downloading `appimagetool` in CI randomly fails, especially when doing successive builds (one in PR and one after merge). So I would be open to alternatives.

I am also planning on working on improving portable windows exe and GitLab support for cargo-dist because well, currently I am using another crate to bundle icon and manifest with the windows executable and it would be great if cargo-dist could do that itself.

mistydemeo:
Thanks for the response! Just wanted to say that I'm out of the office today, so I won't have a chance to discuss more until Monday. Didn't want to leave you hanging.

mistydemeo:
> I am using Gitlab as seen from the link

Aha! That explains why I didn't see it earlier!

> Ideally, I would prefer not to need an external tool, but well, not sure if there is a Rust crate or something we could use. Additionally, I would like to inform everyone that downloading appimagetool in CI randomly fails, especially when doing successive builds (one in PR and one after merge). So I would be open to alternatives.

Yes, if there's a stable source of this - maybe a package manager - that would be very convenient. I also see you're hex editing the binary you've downloaded - did you find that that was necessary under Gitlab's runners, or is it there primarily because the Docker image you referenced also uses the same pattern?

Ayush1325:
> Yes, if there's a stable source of this - maybe a package manager - that would be very convenient. I also see you're hex editing the binary you've downloaded - did you find that that was necessary under Gitlab's runners, or is it there primarily because the Docker image you referenced also uses the same pattern?

The hex editing is taken from here: https://github.com/StratusFearMe21/cargo-appimage/blob/master/README.md#docker

Not sure if that is actually needed now though.

mistydemeo:
Got it.

I think, under the circumstances, it sounds like the appimage tooling here isn't distributed in a convenient and stable enough way for us to use in cargo-dist at this time. The tools we rely on to perform builds are generally ones that are either available by default, or readily accessible to us via package managers.

I'm going to close this PR at this time, but we're open to discussing other ways we can make our packaging/distribution better for Linux GUI apps. Please also let us know if you'd like to discuss the other Windows EXE issues you were having and if there's a way we can incorporate the fixes for that. And thank you again for the contribution and opening the discussion, even if we're not able to incorporate this PR at this time!

PR #1306, opened 2024-08-09
Fix generated comment describing `pr-run-mode`

zanieb:
e.g., the following is incorrect https://github.com/axodotdev/cargo-dist/blob/39abb63c095a8bf9559bf45ab339cf821d15148d/Cargo.toml#L88-L89


mistydemeo (approved):
Good catch, thank you!

PR #1309, opened 2024-08-09
fix: javascript workspace bugs

### LINKED ISSUE #1307: Make repository-url-sameness check more robust
`https://github.com/axodotdev/cargo-dist` and `https://github.com/axodotdev/cargo-dist.git` are treated as different URLs when we already have code that resolves them to the same "web url", which is the only value we actually care about!

This came up because in a package.json `"repository": "github:axodotdev/cargo-dist"` gets desugarred as a `.git` URL. I didn't notice this because i typo'd the package.json repo field in axolotlsay-hybrid, so it essentially had no repository url as far as we were concerned (and then the Cargo one covered for it).

### LINKED ISSUE #1308: dist.toml's [dist] section is never read for packages
I am honestly completely shocked I missed this in testing.

The logic for reading the [dist] section out of a dist.toml or dist-workspace.toml is special cased to only exist when that file is the *workspace* file. If it's a package we just completely fail to find it, because that path didn't need to exist in the original impl of generic projects (where you could only have the one root dist.toml with no children).

Note that the [package] section gets read fine (name, version, repository, etc), which is what was the most complex to implement and where all my testing went. We have several tests for that in axoproject, but axoproject (currently) *explicitly* doesn't understand the [dist] section, so testing for this needs to exist in cargo-dist itself.

I kinda just took it for granted that the [dist] section was being read on packages, and all of our emphasis went into never *needing* to create a dist.toml for packages (making JS more native!) so the place where I would have tested it in axolotlsay-hybrid got deleted. Oops!

For most common cases of usage this isn't a fatal issue, but it definitely needs to be fixed asap for anyone who needs to configure js packages in shared workspaces.

Gankra:
Use the [dist] section package dist.tomls, and add errors for cases where a metadata.dist exists but is being ignored.

Also handles npm package scopes better.

Also handles url equality checks better.

fixes #1308 
fixes #1307

### DIFF
--- a/axoproject/src/generic.rs
+++ b/axoproject/src/generic.rs
@@ -301,6 +301,7 @@ fn package_from(manifest_path: &Utf8Path) -> Result<PackageInfo> {
         cargo_metadata_table: None,
         #[cfg(feature = "cargo-projects")]
         cargo_package_id: None,
+        npm_scope: None,
     };
 
     // Load and apply auto-includes
--- a/axoproject/src/generic.rs
+++ b/axoproject/src/generic.rs
@@ -328,7 +328,7 @@ fn load_package_dist_toml(manifest_path: &Utf8Path) -> Result<PackageManifest> {
 fn merge_package_with_raw_generic(
     package: &mut PackageInfo,
     generic: Package,
-    generic_path: Utf8PathBuf,
+    generic_manifest_path: Utf8PathBuf,
 ) {
     let Package {
         name,
@@ -348,7 +348,7 @@ fn merge_package_with_raw_generic(
         version,
     } = generic;
 
-    package.dist_manifest_path = Some(generic_path);
+    package.dist_manifest_path = Some(generic_manifest_path);
 
     if let Some(val) = name {
         package.name = val;
--- a/axoproject/src/generic.rs
+++ b/axoproject/src/generic.rs
@@ -195,7 +195,7 @@ fn workspace_from(manifest_path: &Utf8Path) -> Result<WorkspaceStructure> {
             let paired_manifest = package.package_root.join(DIST_PACKAGE_TOML);
             if paired_manifest.exists() {
                 let generic = raw_package_from(&paired_manifest)?;
-                merge_package_with_raw_generic(package, generic);
+                merge_package_with_raw_generic(package, generic, paired_manifest);
             }
 
             crate::merge_auto_includes(package, &root_auto_includes);
@@ -210,6 +210,7 @@ fn workspace_from(manifest_path: &Utf8Path) -> Result<WorkspaceStructure> {
             target_dir: workspace_dir.join(DIST_TARGET_DIR),
             workspace_dir,
             manifest_path: manifest_path.to_owned(),
+            dist_manifest_path: Some(manifest_path.to_owned()),
             root_auto_includes,
             #[cfg(feature = "cargo-projects")]
             cargo_metadata_table: None,
@@ -229,6 +230,7 @@ fn single_package_workspace_from(manifest_path: &Utf8Path) -> Result<WorkspaceSt
             target_dir: package.package_root.join(DIST_TARGET_DIR),
             workspace_dir: package.package_root.clone(),
             manifest_path: package.manifest_path.clone(),
+            dist_manifest_path: Some(package.manifest_path.clone()),
             root_auto_includes,
 
             #[cfg(feature = "cargo-projects")]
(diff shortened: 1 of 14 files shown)


Gankra:
currently missing: a checked in test that stresses this (did some local edits to axolotlsay-hybrid to smoke test, but we need a real test before landing)

axoproject/src/lib.rs
```diff
@@ -443,8 +449,13 @@ pub struct PackageInfo {
     ///
     /// This may be needed when e.g. talking to cargo about the package.
     pub true_version: Option<Version>,
-    /// Path to the manifest for this package
+    /// Path to the primary manifest for this package
     pub manifest_path: Utf8PathBuf,
+    /// Path to a dist.toml overriding this package's contents
+    ///
+    /// NOTE: this WILL be the same path as the manifest_path for a pure dist.toml
+    /// package. Don't assume these are disjoint!
```
ashleygwilliams, line 457: is there a reason people might assume they are disjoint beyond that they are 2 diff vars? and if so what's the consequence? besides... being wrong

cargo-dist/src/errors.rs
```diff
@@ -465,6 +465,18 @@ pub enum DistError {
         /// Error messsage details
         message: String,
     },
+
+    /// Something has metadata.dist but shouldn't
+    #[error("The metadata.dist entry in this Cargo.toml isn't being used:\n{manifest_path}")]
```
ashleygwilliams, line 470: love this

cargo-dist/src/tasks.rs
```diff
@@ -846,6 +846,22 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
     ) -> DistResult<Self> {
         let root_workspace_idx = workspaces.root_workspace_idx();
         let root_workspace = workspaces.workspace(root_workspace_idx);
+
+        // Complain if someone still has [workspace.metadata.dist] in a dist-workspace.toml scenario
+        if let Some(dist_manifest_path) = root_workspace.dist_manifest_path.as_deref() {
```
ashleygwilliams, line 851: are we turning this on now?

ashleygwilliams (commented):
lgtm modulo a few questions- mostly want to avoid forcing a migration with this release

axoproject/src/lib.rs
Gankra: Code like init which loads up some state, edits it, and then writes it back could go sideways if it's too clever about batching edits / caching state, but forgets that these "two" files are the same one.

This is not a problem with the current impl.

cargo-dist/src/tasks.rs
Gankra: This code only runs if someone has migrated to dist-workspace.toml, so this is basically just erroring on "it looks like you started a migration but I see some gunk you missed".

axoproject/src/generic.rs
```diff
@@ -321,7 +325,11 @@ fn load_package_dist_toml(manifest_path: &Utf8Path) -> Result<PackageManifest> {
     Ok(manifest)
 }
 
-fn merge_package_with_raw_generic(package: &mut PackageInfo, generic: Package) {
+fn merge_package_with_raw_generic(
+    package: &mut PackageInfo,
+    generic: Package,
+    generic_path: Utf8PathBuf,
```
mistydemeo, line 331: Clarifying what `generic_path` here is in a comment might be nice

axoproject/src/lib.rs
```diff
@@ -355,10 +355,16 @@ pub struct WorkspaceInfo {
...
+    /// This can be either a Cargo.toml, package.json, dist.toml, or dist-workspace.toml.
+    ///
+    /// In most cases this manifest may or may not represent a "real" package. Many systems
+    /// have some notion of "virtual" manifest which exists only to list the actual packages
+    /// in the workspace.
     pub manifest_path: Utf8PathBuf,
+    /// Path to the dist.toml or dist-workspace.toml for the workspace
+    ///
+    /// If this is ever set, then this will be the same as manifest_path.
+    pub dist_manifest_path: Option<Utf8PathBuf>,
```
mistydemeo, line 367: Is it true that this is the same as `manifest_path` for a `package.json`/`Cargo.toml` workspace?

Gankra: The "if this is ever set" clause is doing heavy lifting here. Basically this option is morally a boolean saying manifest_path might be a dist(-workspace).toml, but I decided to make it an option to make code that deals with the pacakge version of dist_manifest_path uniform.

Gankra: Or i guess to put it another way -- if this is ever set, it's shared/generic workspace.

PR #1312, opened 2024-08-12
feat: Mac .pkg installer

mistydemeo:
This adds support for building Mac .pkg installers. This current implementation is intentionally a bit simplistic: it produces one .pkg per package in the user's workspace, without any special branding applied. We currently only allow the app identifier and install location to be configured, but we can consider adding further customization in the future.

Although I added support for configuring this in `init`, I ran into a bit of trouble serializing that value since the new config is set up as a struct rather than two separate fields.

Here's sample config from one app:

```toml
[workspace.metadata.dist.mac-pkg-config]
identifier = "dev.axo.cargo-dist"
install-location = "/usr/local"
```

And here's the No Name / Sans nom installer it produces:

image

### DIFF
--- a/cargo-dist/src/backend/installer/macpkg.rs
+++ b/cargo-dist/src/backend/installer/macpkg.rs
@@ -1,0 +1,99 @@
+//! Code for generating installer.pkg
+
+use std::{collections::BTreeMap, fs};
+
+use axoasset::LocalAsset;
+use axoprocess::Cmd;
+use camino::Utf8PathBuf;
+use serde::Serialize;
+use temp_dir::TempDir;
+use tracing::info;
+
+use crate::{create_tmp, DistResult};
+
+use super::ExecutableZipFragment;
+
+/// Info about a package installer
+#[derive(Debug, Clone, Serialize)]
+pub struct PkgInstallerInfo {
+    /// ExecutableZipFragment for this variant
+    pub artifact: ExecutableZipFragment,
+    /// Identifier for the final installer
+    pub identifier: String,
+    /// Default install location
+    pub install_location: String,
+    /// Final file path of the pkg
+    pub file_path: Utf8PathBuf,
+    /// Dir stuff goes to
+    pub package_dir: Utf8PathBuf,
+    /// The app version
+    pub version: String,
+    /// Executable aliases
+    pub bin_aliases: BTreeMap<String, Vec<String>>,
+}
+
+impl PkgInstallerInfo {
+    /// Build the pkg installer
+    pub fn build(&self) -> DistResult<()> {
+        info!("building a pkg: {}", self.identifier);
+
+        // We can't build directly from dist_dir because the
+        // package installer wants the directory we feed it
+        // to have the final package layout, which in this case
+        // is going to be an FHS-ish path installed into a public
+        // location. So instead we create a new tree with our stuff
+        // like we want it, and feed that to pkgbuild.
+        let (_build_dir, build_dir) = create_tmp()?;
+        let bindir = build_dir.join("bin");
+        LocalAsset::create_dir_all(&bindir)?;
+        let libdir = build_dir.join("lib");
+        LocalAsset::create_dir_all(&libdir)?;
+
+        info!("Copying executables");
+        for exe in &self.artifact.executables {
+            info!("{} => {:?}", &self.package_dir.join(exe), bindir.join(exe));
+            LocalAsset::copy_file_to_file(&self.package_dir.join(exe), bindir.join(exe))?;
+        }
+        #[cfg(unix)]
+        for (bin, targets) in &self.bin_aliases {
+            for target in targets {
+                std::os::unix::fs::symlink(&bindir.join(bin), &bindir.join(target))?;
+            }
+        }
+        for lib in self
+            .artifact
+            .cdylibs
+            .iter()
+            .chain(self.artifact.cstaticlibs.iter())
+        {
+            LocalAsset::copy_file_to_file(&self.package_dir.join(lib), libdir.join(lib))?;
+        }
+
+        // The path the two pkg files get placed in while building
+        let pkg_output = TempDir::new()?;
+        let pkg_output_path = pkg_output.path();
+        let pkg_path = pkg_output_path.join("package.pkg");
+        let product_path = pkg_output_path.join("product.pkg");
+
+        let mut pkgcmd = Cmd::new("/usr/bin/pkgbuild", "create individual pkg");
+        pkgcmd.arg("--root").arg(build_dir);
+        pkgcmd.arg("--identifier").arg(&self.identifier);
+        pkgcmd.arg("--install-location").arg(&self.install_location);
+        pkgcmd.arg("--version").arg(&self.version);
+        pkgcmd.arg(&pkg_path);
+        // Ensures stdout from the build process doesn't taint the dist-manifest
+        pkgcmd.stdout_to_stderr();
+        pkgcmd.run()?;
+
+        // OK, we've made a package. Now wrap it in a product pkg.
+        let mut productcmd = Cmd::new("/usr/bin/productbuild", "create final product .pkg");
+        productcmd.arg("--package").arg(&pkg_path);
+        productcmd.arg(&product_path);
+        productcmd.stdout_to_stderr();
+        productcmd.run()?;
+
+        fs::copy(&product_path, &self.file_path)?;
+
+        Ok(())
+    }
+}
--- a/cargo-dist/src/config.rs
+++ b/cargo-dist/src/config.rs
@@ -439,6 +439,11 @@ pub struct DistMetadata {
     /// Any additional steps that need to be performed before building local artifacts
     #[serde(default)]
     pub github_build_setup: Option<String>,
+
+    /// Configuration specific to Mac .pkg installers
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub mac_pkg_config: Option<MacPkgConfig>,
 }
 
 /// values of the form `permission-name: read`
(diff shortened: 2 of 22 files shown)


cargo-dist/src/backend/installer/macpkg.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        // OK, we've made a package. Now wrap it in a product pkg.
+        let mut productcmd = Cmd::new("/usr/bin/productbuild", "create final product .pkg");
+        productcmd.arg("--package").arg(&pkg_path);
+        productcmd.arg(&product_path);
+        productcmd.run()?;
+
+        fs::copy(&product_path, &self.file_path)?;
+
+        assert!(self.file_path.exists());
```
Gankra, line 88: what's up with this? debug code?

cargo-dist/src/config.rs
```diff
@@ -435,6 +435,11 @@ pub struct DistMetadata {
     /// Any additional steps that need to be performed before building local artifacts
     #[serde(default)]
     pub github_build_setup: Option<String>,
+
+    /// Configuration specific to Mac .pkg installers
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub mac_pkg_config: Option<MacPkgConfig>,
```
Gankra, line 442: glossing over this as something we can freely rejig in config 1.0...

cargo-dist/src/init.rs
```diff
@@ -799,6 +802,66 @@ fn get_new_dist_metadata(
         }
     }
 
+    // Special handling of the Homebrew installer
```
Gankra, line 805: copy-paste error

cargo-dist/src/init.rs
```diff
@@ -799,6 +802,66 @@ fn get_new_dist_metadata(
...
+                    .interact_text()?;
+                eprintln!();
+                res
+            };
+            let identifier = identifier.trim();
+            if identifier.is_empty() {
+                return Err(DistError::MacPkgBundleIdentifierMissing {});
+            }
+
+            let prompt = r#"Please enter the installation prefix this .pkg should use."#;
```
Gankra, line 840: should probably specify the default, or we shouldn't even ask? Not sure how typical it is to overload this.

cargo-dist/src/init.rs
```diff
@@ -997,6 +1060,7 @@ fn apply_dist_to_metadata(metadata: &mut toml_edit::Item, meta: &DistMetadata) {
         bin_aliases: _,
         system_dependencies: _,
         github_build_setup: _,
+        mac_pkg_config: _,
```
Gankra, line 1063: Wait doesn't this make init pointless.

cargo-dist/src/tasks.rs
```diff
@@ -2484,6 +2488,115 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+        let bin_aliases = release.bin_aliases.clone();
+        let checksum = release.checksum;
+
+        // Make a pkg for every darwin platform
+        for variant_idx in variants {
+            let variant = self.variant(variant_idx);
+            let binaries = variant.binaries.clone();
+            let target = &variant.target;
+            if !target.contains("darwin") {
+                continue;
```
Gankra, line 2508: Normally we warn here, I think?

cargo-dist/src/tasks.rs
```diff
@@ -2484,6 +2488,115 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+            let dir_path = self.inner.dist_dir.join(&dir_name);
+
+            // Compute which package we're actually building, based on the binaries
+            let mut package_info: Option<(String, PackageIdx)> = None;
+            for &binary_idx in &binaries {
+                let binary = self.binary(binary_idx);
+                if let Some((existing_spec, _)) = &package_info {
+                    // we haven't set ourselves up to bundle multiple packages yet
+                    if existing_spec != &binary.pkg_spec {
+                        return Err(DistError::MultiPackage {
```
Gankra, line 2524: ...wild why was this logic exclusive to msi

Gankra: Oh right I think it's because I needed to enforce this for the sake of calling into cargo-wix, but I don't think the pkg stuff has that limitation?

Gankra (changes requested):

cargo-dist/src/backend/installer/macpkg.rs
mistydemeo: Sure is, I'll get rid of it

cargo-dist/src/init.rs
mistydemeo: Overloading it is pretty common. Many people are going to want the default `/usr/local`, which is in the `PATH` but shared with other software. Other people are going to prefer, say, `/opt/mypkg`, which is out of the default `PATH` but a prefix fully owned by them, etc.

I was going for an "accept NULL, unwrap to `/usr/local` later" approach. But I guess that means the default could shift under the user, which maybe isn't ideal - I guess "inline a default if not specified" is probably better.

cargo-dist/src/init.rs
mistydemeo: I left a comment above about having difficulty with the `init` stuff because it's a struct rather than two bare fields - that's why this is currently omitted. Was hoping to talk with you about it and/or make this depend on config 1.0, where I assume we'll have better handling.

cargo-dist/src/tasks.rs
mistydemeo: I was following the pattern from the MSI, which doesn't warn here.

cargo-dist/src/config.rs
mistydemeo: Yep, like I mentioned elsewhere this is something I wanted to check in with you about wrt handling before/after config 1.0.

cargo-dist/src/tasks.rs
mistydemeo: Oh, so the goal wasn't to create a single MSI containing every package but rather just that cargo-wix can't understand what to do when given multiple packages? I think I misunderstood the purpose of this guard then.

cargo-dist/src/backend/installer/macpkg.rs
```diff
@@ -0,0 +1,92 @@
...
+        // to have the final package layout, which in this case
+        // is going to be an FHS-ish path installed into a public
+        // location. So instead we create a new tree with our stuff
+        // like we want it, and feed that to pkgbuild.
+        let build_dir_target = TempDir::new()?;
+        let build_dir = build_dir_target.path();
+        let bindir = build_dir.join("bin");
+        fs::create_dir_all(&bindir)?;
+        let libdir = build_dir.join("lib");
+        fs::create_dir_all(&libdir)?;
```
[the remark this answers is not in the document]
mistydemeo: Promoted the tempdir thing to a public method; making me wonder if we just want this to be in axoasset.

Converted the rest into axoasset calls except the symlinks, which axoasset doesn't know about yet. I can add that though, feels in scope.

cargo-dist/src/backend/installer/macpkg.rs
```diff
@@ -0,0 +1,92 @@
...
+        }
+        #[cfg(unix)]
+        for (bin, targets) in &self.bin_aliases {
+            for target in targets {
+                std::os::unix::fs::symlink(&bindir.join(bin), &bindir.join(target))?;
+            }
+        }
+        for lib in &self.artifact.cdylibs {
+            fs::copy(&self.package_dir.join(lib), libdir.join(lib))?;
+        }
```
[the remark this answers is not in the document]
mistydemeo: Oversight (or a sign I wrote this between cdylibs and cstaticlibs landing?), fixed

cargo-dist/src/tasks.rs
mistydemeo: Right, I see now why we've got a separate conflict here. We're fetching some package-level metadata to pass to the package build - specifically, the identifier field is package-level metadata. We don't have the package directly accessible to us - just on the binaries - and if we *did* end up with a conflict of which package they resolve to, I don't know if I could get a better answer than the MSI path does.

cargo-dist/src/init.rs
```diff
@@ -1430,3 +1502,26 @@ where
...
+    if let Some(mac_pkg_config) = val {
+        let new_item = &mut table[key];
+        let mut new_table = toml_edit::table();
+        if let Some(new_table) = new_table.as_table_mut() {
+            new_table.insert("identifier", toml_edit::value(&mac_pkg_config.identifier));
+            if let Some(location) = &mac_pkg_config.install_location {
+                new_table.insert("install-location", toml_edit::value(location));
+            }
+            new_table.decor_mut().set_prefix(desc);
+        }
```
Gankra, line 1522: Is there a reason these can't call into apply_optional and friends?

cargo-dist/src/init.rs
```diff
@@ -1430,3 +1502,26 @@ where
...
+    desc: &str,
+    val: Option<&MacPkgConfig>,
+) {
+    if let Some(mac_pkg_config) = val {
+        let new_item = &mut table[key];
+        let mut new_table = toml_edit::table();
+        if let Some(new_table) = new_table.as_table_mut() {
+            new_table.insert("identifier", toml_edit::value(&mac_pkg_config.identifier));
+            if let Some(location) = &mac_pkg_config.install_location {
+                new_table.insert("install-location", toml_edit::value(location));
```
Gankra, line 1519: please destructure the value rather than individually matching the fields -- like this code should basically look exactly like the caller function

Gankra (changes requested):
(also we need to disable the init code temporarily to land this)

cargo-dist/src/init.rs
mistydemeo: Huh, right, I guess I didn't really consider that this is *also* a table we could just `apply_optional_*` into.

mistydemeo: Would you like me to break this init commit out into a separate PR? Then we can land the rest with the init stuff missing.

Gankra (approved):

PR #1317, opened 2024-08-13
fix(book): redirect waytooquickstart to new rust quickstart

ashleygwilliams:


book/book.toml
```diff
@@ -13,5 +13,8 @@ search.use-boolean-and = true
 [output.linkcheck]
 warning-policy = "error"
 
+[output.html.redirect]
+"/way-too-quickstart.html" = "/quickstart/rust.html"
```
mistydemeo, line 17: This needs to be relative to handle the subpath the book is nested under. With just `/`, it tries to go to the root. eg this makes http://127.0.0.1:7979/book/way-too-quickstart.html redirect to http://127.0.0.1:7979/rust/quickstart.html instead of http://127.0.0.1:7979/book/rust/quickstart.html

```suggestion
"/way-too-quickstart.html" = "./quickstart/rust.html"
```

mistydemeo (changes requested):

mistydemeo (approved):
Tested locally with `oranda dev`; http://127.0.0.1:7979/book/way-too-quickstart.html goes to the right place.

Issue #1275, opened 2024-08-02
Unchecking homebrew should leave no trace of homebrew


alexanderjophus:
After running `cargo dist init`, I checked homebrew, then later realised I didn't need it yet.

I then ran `cargo dist init` again, unchecked homebrew, however the following was still in my Cargo.toml

```toml
# Config for 'cargo dist'
[workspace.metadata.dist]
# Publish jobs to run in CI
publish-jobs = ["homebrew"]
```

Similarly in CI the following step was still there `publish-homebrew-formula` (and depended on by `announce`

Open to input on whether it should fully take out the CI step - I think it should, but is there the use case of someone editing it and needing the edits later (could they just look at their git history?)

Also, if this is needed, I'd love to try contributing back!

Gankra:
Oh this is a great catch!

This is where we currently apply the *reverse* logic of asking whether they want to enable homebrew publish. In this case I think we can automatically force it off instead of prompting, but the same basic logic that's here applies, just a bit reversed.

Instead of checking if homebrew is newly *added* we can check if homebrew is newly *removed* and then unconditionally strip the entry from publish if it's there.

https://github.com/axodotdev/cargo-dist/blob/7eeb4c3cd290d97f9c4b5e6252a9db33e8a13132/cargo-dist/src/init.rs#L738-L751

https://github.com/axodotdev/cargo-dist/blob/7eeb4c3cd290d97f9c4b5e6252a9db33e8a13132/cargo-dist/src/init.rs#L767-L773

alexanderjophus:
@Gankra I'm planning on attempting this this weekend (unless anyone else has given it a go). My rust isn't great, so please be patient 🤞

Gankra:
merged!

PR #1329, opened 2024-08-16
reverse homebrew logic if user unchecks it

alexanderjophus:
Addresses https://github.com/axodotdev/cargo-dist/issues/1275

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -787,6 +787,16 @@ fn get_new_dist_metadata(
                 );
             }
         }
+    } else {
+        let homebrew_toggled_off = orig_meta
+            .installers
+            .as_deref()
+            .unwrap_or_default()
+            .contains(&InstallerStyle::Homebrew);
+        if homebrew_toggled_off {
+            meta.tap = None;
+            publish_jobs.retain(|job| job != &PublishStyle::Homebrew);
+        }
     }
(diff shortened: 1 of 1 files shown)


cargo-dist/src/init.rs
```diff
@@ -787,6 +787,10 @@ fn get_new_dist_metadata(
                 );
             }
         }
+    } else {
+        // homebrew doesn't exist in the new config
+        meta.tap = None;
+        publish_jobs.retain(|job| job != &PublishStyle::Homebrew);
```
Gankra, line 793: Good start, but this is a bit too aggressive.

The homebrew installer can be enabled per-package, but tap/publish settings can/should be global. So I think it's only sound to unset these settings if they specifically *unchecked* the homebrew installer in *this* run, because they're clearly trying to disable the homebrew installer globally. You can see a check like this in `homebrew_is_new` above (you just gotta invert it to make `homebrew_was_set`).

Gankra (changes requested):

cargo-dist/src/init.rs
alexanderjophus: I think I understood what to do, but I don't understand _why_. Thanks for the review!

Gankra (approved):

Gankra:
Thanks!

Issue #810, opened 2024-02-22
Homebrew publish fails tests


ucodery:
Hi, I am new to cargo-dist but already a super fan! I am trying to figure out how to publish my app as a brew formula and have got as far as I can with the docs. This is my first time using cargo-dist and my first time managing my own brew tap, so it is entirely possible that I did something wrong.

I have successfully got my Rust project committed to my homebrew project but this triggers brew tests in CI that fail on the new Formula file. The `brew test-bot` command ultimately fails ([here](https://github.com/ucodery/homebrew-draft/actions/runs/7995352292/job/21835482798)) with
```
1 file inspected, 9 offenses detected, 7 offenses autocorrectable
```
I'm sure I could fix these mistakes but I am worried they would reappear on the next dist publish.

To get here I ran
```
brew tap-new
```
and uploaded that one commit. nothing else in the homebrew repo.
On the app repo I added `tap` and `publish-jobs` and bumped the version in Cargo.toml, then ran
```
cargo dist generate ci --check
```
and uploaded with a new tag (and added a repo secret).

That led to a new formula file in my homebrew repo, but the above-mentioned failure.

ashleygwilliams:
interesting! let me take a look at this

ashleygwilliams:
this is very neat and i think we can both fix these issues, but also add this tooling to our own tests to make sure we're generating formulae that pass.

before i dive in- are you at all interested in contributing some/all of these changes? would be happy to point you in the right direction if so- if not no worries!!

ucodery:
@ashleygwilliams I am interested in making the changes. brew's suggestions seem very straightforward and reasonable. I just wanted to confirm it was a real issue with the code first (and not maybe that I read docs wrong).

ashleygwilliams:
yup! i agree the changes seem very straightforward and reasonable. our inhouse brew expert @mistydemeo is on vacation, and so i'd want to have her review before we merged anything, but we are all cargo fmt fans here, and this seems like that for brew- hard for me to quibble with it! i do think the corrected code will likely end up behaving the same way as the current code but i like following standards, especially tool-enforced ones.

ucodery:
[...]

PR #1340, opened 2024-08-18
format formulas with brew style --fix

### LINKED ISSUE #1331: Homebrew publishing: autogenerated formula fails autogenerated tests
The formulas autogenerated by cargo-dist when publishing to a Homebrew tap have formatting that fails the default Github Actions tests that are created when you run `brew tap-new`. Specifically, they fail the style checks run with `brew style`. 

Apologies if this is out of scope, but given that users will run into this CI failure if they follow the [cargo-dist instructions](https://opensource.axo.dev/cargo-dist/book/installers/homebrew.html) for setting up homebrew releases, I figured I'd open an issue. Similar to [this issue](https://github.com/axodotdev/cargo-dist/issues/810).

Here are the logged failures I personally ran into (first one is fixable user-side, the others are from cargo-dist's template):

```
==> Running TapSyntax#run!
==> brew style cxreiff/tap
==> FAILED
Full style cxreiff/tap output
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:16:5: C: [Correctable] Style/SoleNestedConditional: Consider merging nested conditions into outer if conditions.
      if Hardware::CPU.intel?
      ^^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:23:20: C: [Correctable] Layout/SpaceInsideHashLiteralBraces: Space inside { missing.
    BINARY_ALIASES = {"aarch64-apple-darwin": {}, "x86_64-apple-darwin": {}, "x86_64-pc-windows-gnu": {}, "x86_64-unknown-linux-gnu": {}}
                     ^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:23:20: C: [Correctable] Style/MutableConstant: Freeze mutable objects assigned to constants.
    BINARY_ALIASES = {"aarch64-apple-darwin": {}, "x86_64-apple-darwin": {}, "x86_64-pc-windows-gnu": {}, "x86_64-unknown-linux-gnu": {}}
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:23:119: C: [Correctable] Layout/LineLength: Line is too long. [135/118]
    BINARY_ALIASES = {"aarch64-apple-darwin": {}, "x86_64-apple-darwin": {}, "x86_64-pc-windows-gnu": {}, "x86_64-unknown-linux-gnu": {}}
                                                                                                                        ^^^^^^^^^^^^^^^^^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:23:135: C: [Correctable] Layout/SpaceInsideHashLiteralBraces: Space inside } missing.
    BINARY_ALIASES = {"aarch64-apple-darwin": {}, "x86_64-apple-darwin": {}, "x86_64-pc-windows-gnu": {}, "x86_64-unknown-linux-gnu": {}}
                                                                                                                                        ^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:41:5: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
      if OS.mac? && Hardware::CPU.arm?
      ^^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:44:5: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
      if OS.mac? && Hardware::CPU.intel?
      ^^
  Taps/cxreiff/homebrew-tap/Formula/ttysvr.rb:47:5: C: [Correctable] Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
      if OS.linux? && Hardware::CPU.intel?
      ^^
  
  1 file inspected, 9 offenses detected, 9 offenses autocorrectable
```

cxreiff:
Fixes #1331.

## Problem
Creating a personal tap with `brew tap-new` creates a default [Github Actions workflow](https://github.com/cxreiff/homebrew-tap/blob/main/.github/workflows/tests.yml) that runs a homebrew-specific style check (`brew style <TAPNAME>`) on the repo. Cargo-dist's generated homebrew formulas fail this style check. This does not block the core functionality of cargo-dist's homebrew publishing functionality (formula still functions), ~~but does mean that users following [cargo-dist's instructions](https://opensource.axo.dev/cargo-dist/book/installers/homebrew.html) exactly will run into workflow failures~~ (EDIT: This was not true at all- docs mention a plain repo, not one created with `brew tap-new`).

## Previously
It looks like this has come up before (#810) and it was solved by individually fixing each formatting problem in the templates ([PR 818](https://github.com/axodotdev/cargo-dist/pull/818)).

## Solution
Rather than separately chase formatting issues as they come up, the homebrew publish workflow could run `brew style --fix` on each formula before committing it.

### Pros
- Changes to the homebrew style rules can be adjusted to automatically in the future without changes to cargo-dist's templates.
- Changes can be made to cargo-dist's templates more freely with less fear of breaking style rules.
- Keeps templates simple: some Homebrew style rules interact with cargo-dist's templating strategy in ways that take complicated template conditions to solve (there are rules that apply to single statements inside conditionals, while in the formula templates the number of statements in a conditional might vary based on supported platforms.

### Cons
- Homebrew publish workflow will take longer as it will need to install Homebrew every run.
- Depends on reliability of `brew style --fix` (could break something without cargo-dist code changes if brew pushes a bad change).
- Not all style changes are auto-fixable.

## Testing
I've tested this by manually making these changes in the generated workflow itself and tagging a release. This means that the generation itself hasn't explicitly been tested.

Example run with new action steps: https://github.com/cxreiff/ttysvr/actions/runs/10436301681


cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2
```diff
@@ -24,6 +24,16 @@
           pattern: artifacts-*
           path: Formula/
           merge-multiple: true
+      # Set up and cache homebrew for formatting formulas
+      - name: Set up Homebrew
```
mistydemeo, line 28: We shouldn't need to do this. [Homebrew is preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md#package-management) on the Linux runners we use as the global runners; we don't need to reinstall it. You should just be able to perform `brew update` before `brew style --fix` to make sure it's the latest version.

mistydemeo (changes requested):
Thank you! This won't fix all issues, but it does seem likely to reduce the number of issues like this one going forward.

This PR looks good, but one suggestion.

> does mean that users following cargo-dist's instructions exactly will run into workflow failures.

We actually *don't* recommend that users create a tap with `brew tap-new` - it's explicitly not mentioned in our documentation! We recommend that users simply create an empty repository to publish packages to. There are some advantages to `brew test-bot`, but fewer in the cargo-dist case: while one of brew-test-bot's strengths is its ability to produce binary bottle packages, cargo-dist is already distributing binaries without using that infrastructure.

Since `brew style` iterates through styles very quickly, I do feel it's not always helpful for user taps to make use of it for generated formulae like this one - `brew style --fix` will help keep us up to date, but we still may end up being somewhat behind Homebrew's style schedule.

cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2
cxreiff: Had no idea, thank you! Fixed.

cxreiff:
Oh wow, I must have read somewhere how to set up a tap and then totally misremembered where I saw it, apologies! Added a correction in the description.

mistydemeo:
All good! I thought it might be something like that - and was a helpful reminder to me to check our docs to make sure they're actually suggesting what I intended them to!

I'll give this a test tomorrow.

cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2
```diff
@@ -36,6 +36,9 @@
             name=$(echo "$filename" | sed "s/\.rb$//")
             version=$(echo "$release" | jq .app_version --raw-output)
 
+            brew update
+            brew style --fix "Formula/${filename}"
```
mistydemeo, line 40: It looks like `brew style` exits non-zero if it fails to fix every possible issue with a formula; we should guard this with `|| true` to make sure that doesn't fail the build.

```suggestion
            brew style --fix "Formula/${filename}" || true
```

mistydemeo (commented):
Gave it a test, and the changes it made are looking good: https://github.com/mistydemeo/homebrew-cargodisttest/commit/1c5e0ac422f77e6f6ea88b418cf7e5659d0ad041

Just a couple more suggestions before this gets merged!

mistydemeo (approved):
Thank you! Really appreciate it.

PR #1341, opened 2024-08-18
fix(redirect): quickstart -> quickstart index

ashleygwilliams:
when i originally made this redirect i was focused on doing a best-effort match to the previous content. however, i didn't realize the quickstart was showing up for generic "cargo-dist" searches (see below, makes a lot of sense). i think putting the redirect at the index page makes more sense here. we *may* want to soup up that page so it's a little nicer than it is currently, but i think the best redirect goes to the index, souped up or not :)

Screenshot 2024-08-18 at 5 54 41 PM


mistydemeo (approved):
Good call.

Gankra (approved):

PR #1345, opened 2024-08-19
Attempt conversion of SPDX license expressions to Homebrew DSL

### LINKED ISSUE #820: Homebrew: improve license field handling
While Homebrew uses SPDX licenses, it doesn't parse the full SPDX grammar. Rather, it [uses a Ruby DSL](https://docs.brew.sh/License-Guidelines#complex-spdx-license-expressions) to express the relationships between licenses. Cargo, meanwhile, uses SPDX strings directly, so that's what we pass through to the Homebrew formula. This is mostly harmless: it means Homebrew won't be able to provide detailed information about the license, but there's very little harm to it. If we felt like it though, we could use the [spdx crate](https://crates.io/crates/spdx) to parse the field from Cargo.toml and output the specific format that Homebrew expects. It's a nice-to-have, but not essential.

What we do now:

```ruby
license "MIT OR Apache-2.0"
```

What Homebrew expects:

```ruby
license any_of: ["MIT", "Apache-2.0"]
```

cxreiff:
Resolves #820.

## Problem
Currently, when publishing to a Homebrew tap, whatever string is in the crate's Cargo.toml license field (usually a SPDX license expression) gets passed into `license` in the Homebrew formula. However, rather than using SPDX syntax, Homebrew has its own ruby syntax for specifying licenses. So, when passed an SPDX expression, Homebrew:
- Cannot provide as detailed information about the package's licensing.
- Throws errors when linting the formula with `brew audit`/`brew test-bot`.

## Solution
- Use the `spdx` crate to parse the SPDX license expression.
- Iterate through the resulting token stream to construct a string matching Homebrew's license domain-specific language.
- Pass the result into Homebrew's `license` command, falling back to the original license string if there are issues parsing.

## Scope
- This PR just handles OR/AND operators, as the `spdx` crate has not yet implemented v3 of the spec. This means that the WITH operator is not yet covered, and neither are lower-case versions of the operators.
  - `spdx` issue for implementing spec v3 is [here](https://github.com/EmbarkStudios/spdx/issues/68).
- This PR formats everything on one line- in combination with [PR 1340](https://github.com/axodotdev/cargo-dist/pull/1340) this shouldn't be an issue, as it will be auto-formatted.

## Testing
I added tests for the conversion utility function, tested the output in a Homebrew formula, and I inspected and updated the test snapshots. This PR could certainly use some additional testing from someone with deeper familiarity with the project.

### DIFF
--- a/cargo-dist/src/backend/installer/homebrew.rs
+++ b/cargo-dist/src/backend/installer/homebrew.rs
@@ -3,6 +3,10 @@
 use axoasset::LocalAsset;
 use cargo_dist_schema::DistManifest;
 use serde::Serialize;
+use spdx::{
+    expression::{ExprNode, Operator},
+    Expression, ParseError,
+};
 
 use super::InstallerInfo;
 use crate::{
@@ -178,9 +182,64 @@ pub fn to_class_case(app_name: &str) -> String {
     chars.iter().collect()
 }
 
+/// Converts SPDX license string into Homebrew Ruby DSL
+// Homebrew DSL reference: https://docs.brew.sh/License-Guidelines
+pub fn to_homebrew_license_format(app_license: &str) -> Result<String, ParseError> {
+    let spdx = Expression::parse(app_license)?;
+    let mut spdx = spdx.iter().peekable();
+    let mut buffer: Vec<String> = vec![];
+
+    while let Some(token) = spdx.next() {
+        match token {
+            ExprNode::Req(req) => {
+                // If token is a license, push to the buffer as-is for next operator or end.
+                let requirement = format!("\"{}\"", req.req);
+                buffer.push(requirement);
+            }
+            ExprNode::Op(op) => {
+                // If token is an operation, group operands in buffer into all_of/any_of clause.
+                // Operations are postfix, so we pop off the previous two elements and combine.
+                let second_operand = buffer.pop().expect("Operator missing first operand.");
+                let first_operand = buffer.pop().expect("Operator missing second operand.");
+                let mut combined = format!("{}, {}", first_operand, second_operand);
+
+                // If the operations that immediately follow are the same as the current operation,
+                // squash their operands into the same all_of/any_of clause.
+                while let Some(ExprNode::Op(next_op)) = spdx.peek() {
+                    if next_op != op {
+                        break;
+                    }
+                    let _ = spdx.next();
+                    let operand = buffer.pop().expect("Operator missing first operand.");
+                    combined = format!("{}, {}", operand, combined);
+                }
+
+                // Use corresponding homebrew DSL keyword and square bracket the list of licenses.
+                let operation = match op {
+                    Operator::And => "all_of",
+                    Operator::Or => "any_of",
+                };
+                let mut enclosed = format!("{operation}: [{combined}]");
+
+                // Only wrap all_of/any_of clause in brackets if it is nested within an outer clause.
+                if spdx.peek().is_some() {
+                    enclosed = format!("{{ {enclosed} }}");
+                }
+
+                // Push clause back onto the buffer, as it might be an operand in another clause.
+                buffer.push(enclosed);
+            }
+        }
+    }
+
+    // After all tokens have been iterated through, if the SPDX expression is well-formed, there
+    // should only be a single element left in the buffer: a single license or outermost clause.
+    Ok(buffer[0].clone())
+}
+
 #[cfg(test)]
 mod tests {
-    use super::to_class_case;
+    use super::{to_class_case, to_homebrew_license_format};
 
     fn run_comparison(in_str: &str, expected: &str) {
         let out_str = to_class_case(in_str);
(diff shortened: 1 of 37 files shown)


cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -178,9 +182,64 @@ pub fn to_class_case(app_name: &str) -> String {
...
+        match token {
+            ExprNode::Req(req) => {
+                // If token is a license, push to the buffer as-is for next operator or end.
+                let requirement = format!("\"{}\"", req.req);
+                buffer.push(requirement);
+            }
+            ExprNode::Op(op) => {
+                // If token is an operation, group operands in buffer into all_of/any_of clause.
+                // Operations are postfix, so we pop off the previous two elements and combine.
+                let second_operand = buffer.pop().expect("Operator missing first operand.");
```
mistydemeo, line 202: Are these cases unreachable, or could they happen with a malformed SPDX string?

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -265,4 +324,40 @@ mod tests {
...
         run_comparison("openssl@blah", "Openssl@blah");
     }
+
+    fn run_spdx_comparison(spdx_string: &str, homebrew_dsl: &str) {
+        let result = to_homebrew_license_format(spdx_string).unwrap();
+        assert_eq!(result, homebrew_dsl);
+    }
+
+    #[test]
+    fn spdx_single() {
```
mistydemeo, line 334: Thank you for including tests! Might be worth also importing Homebrew's tests wholesale so we know it works with the stuff Homebrew tests against too: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/utils/spdx_spec.rb#L214-L239

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -265,4 +324,40 @@ mod tests {
     fn ampersand_but_no_digit() {
         run_comparison("openssl@blah", "Openssl@blah");
     }
+
+    fn run_spdx_comparison(spdx_string: &str, homebrew_dsl: &str) {
```
mistydemeo, line 328: We can probably tag this with `#[test]` too since it doesn't need to be compiled into the production binary.

mistydemeo (commented):
Thank you so much! This is quite thorough - I appreciate it. Just a couple of comments.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: They _should_ be unreachable, as `Expression::parse()` will return an error and early-out if the SPDX string is malformed. I added error messages here anyways because the invariant (operators are always preceded by two operands) is not statically enforced. I considered returning a ParseError, but it gets a bit icky because instantiating `spdx`'s ParseError requires a "span" of which characters in the string were unable to be parsed, and only the license tokens keep track of their span, not the operators.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: I was following the example of the `run_comparison()` function in the same test module, my assumption was the whole module gets excluded because of the `#[cfg(test)]`. Should I tag `run_comparison()` as well?

mistydemeo: No, you had it right - I missed this was inside the test module.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: Basically SPDX takes a string like `"A OR B"` and then creates a stream like `[Req(A), Req(B), Op(Or)]`, so there is no valid SPDX string that would result in an operator without two operands. If this panic occurs, it means there is a serious logic issue in either this code or in the `spdx` crate. I have tested a couple weird strings that abstractly correspond to this case, e.g. "Apache-2.0 OR" or "MIT AND OR", but both properly return ParseErrors when parsed by spdx (I'll add these "invalid" test cases along with homebrew's cases mentioned below).

mistydemeo: Perfect, thank you! I appreciate you testing this out to confirm it.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: Added more tests to cover homebrew's cases–not all of them as this change does not support "WITH", but now parentheses and plus signs are covered, as well as some invalid cases.

mistydemeo: Perfect, thanks so much! Not handling `WITH` at this point makes sense given the upstream crate's limitations.

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -265,4 +326,99 @@ mod tests {
...
+            "MIT OR Apache-2.0 OR CC-BY-4.0",
+            r#"any_of: ["MIT", "Apache-2.0", "CC-BY-4.0"]"#,
+        );
+    }
+
+    #[test]
+    fn spdx_three_licenses_or_and() {
+        run_spdx_comparison(
+            "MIT OR Apache-2.0 AND CC-BY-4.0",
+            r#"any_of: ["MIT", { all_of: ["Apache-2.0", "CC-BY-4.0"] }]"#,
```
mistydemeo, line 372: Looks like Homebrew parses this one to 

```ruby
{:all_of=>[{:any_of=>["MIT", "Apache-2.0"]}, "CC-BY-4.0"]}
```

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -265,4 +326,99 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn spdx_three_licenses_and_or() {
+        run_spdx_comparison(
+            "MIT AND Apache-2.0 OR CC-BY-4.0",
+            r#"any_of: [{ all_of: ["MIT", "Apache-2.0"] }, "CC-BY-4.0"]"#,
```
mistydemeo, line 380: Likewise, Homebrew parses this to

```ruby
{:all_of=>["MIT", {:any_of=>["Apache-2.0", "CC-BY-4.0"]}]}
```

mistydemeo (commented):
Looking good! Just noticed that two conditions here are parsed differently from how Homebrew does it - wanted to check in and see how you're feeling about that at this point.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: Yeah... I did notice that, but I believe Homebrew is incorrect here 😅. In the [spdx spec](https://spdx.github.io/spdx-spec/v3.0/annexes/SPDX-license-expressions/#d45-order-of-precedence-and-parentheses) they specify that the operator order of precedence is WITH -> AND -> OR, and actually use an example that is basically the same as this test case.

Homebrew is just going left-to-right I believe- it shouldn't be a problem though, for a one-way conversion.

cargo-dist/src/backend/installer/homebrew.rs
cxreiff: Actually scratch what I said about left-to-right, this case would indicate that they _are_ looking at operator precedence, but just have OR/AND swapped.

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Hah! OK, that makes sense. Happy for us to have the more correct version here.

mistydemeo (approved):
Thank you!

PR #1346, opened 2024-08-19
feat: make glibc version reqs per-platform

Gankra:


cargo-dist/tests/snapshots/akaikatana_basic.snap
```diff
@@ -185,10 +186,11 @@ download_binary_and_run_installer() {
             _updater_bin=""
             ;;
         *)
-            err "there isn't a package for $_arch"
+            err "internal installer error: $_artifact_name doesn't exist!?"
```
mistydemeo, line 189: Right, I guess that's because if we'd failed to pick a real archive in `select_archive_for_arch` we would have errored out before we got here?

mistydemeo (approved):

cargo-dist/tests/snapshots/akaikatana_basic.snap
Gankra: Yep

PR #1348, opened 2024-08-20
fix: parsing multiline brew output

mistydemeo:
The `brew bundle exec -- /usr/bin/env` output now contains multiline strings, which I should always have been guarding against anyway. This updates us to use the `-0` option to `env`, which separates lines in the output by nul bytes instead of newlines - that way we can more safely parse the output of environment variables containing newlines.

### DIFF
--- a/cargo-dist/src/env.rs
+++ b/cargo-dist/src/env.rs
@@ -13,7 +13,7 @@ use camino::Utf8Path;
 pub fn fetch_brew_env(
     dist_graph: &DistGraph,
     working_dir: &Utf8Path,
-) -> DistResult<Option<String>> {
+) -> DistResult<Option<Vec<String>>> {
     if let Some(brew) = &dist_graph.tools.brew {
         if Utf8Path::new("Brewfile").exists() {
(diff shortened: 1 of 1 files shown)


cargo-dist/src/env.rs
```diff
@@ -24,24 +24,37 @@ pub fn fetch_brew_env(
...
+                .arg("-0")
                 .current_dir(working_dir)
                 .output()?;
 
-            return Ok(Some(String::from_utf8_lossy(&result.stdout).to_string()));
+            let s = String::from_utf8_lossy(&result.stdout).to_string();
+            // Split into lines based on the \0 separator
+            let output = s
+                // If there's a trailing NUL, we want that gone too
+                .trim_end_matches(&['\0', '\n'])
```
Gankra, line 38: whoops meant to be commenting on this

mistydemeo: I guess that's legal, and also extremely chaotic. Hm. I guess I could `trim_end_matches()` for the newline first, then `trim_end_matches` for `\0` as a second call. That'd avoid chomping a newline that comes before the `\0`.

PR #1352, opened 2024-08-21
docs: add notes to two SPDX tests

mistydemeo:
These two tests produce different results in Homebrew's SPDX parser, but that appears to be a bug in Homebrew's parser rather than this one. Added notes to both tests so that we expect the results to be different from what Homebrew returns.

refs https://github.com/axodotdev/cargo-dist/pull/1345#discussion_r1723932405.

cc @cxreiff

### DIFF
--- a/cargo-dist/src/backend/installer/homebrew.rs
+++ b/cargo-dist/src/backend/installer/homebrew.rs
@@ -369,6 +369,9 @@ mod tests {
     fn spdx_three_licenses_or_and() {
         run_spdx_comparison(
             "MIT OR Apache-2.0 AND CC-BY-4.0",
+            // NOTE: Homebrew parses this as {:all_of=>[{:any_of=>["MIT", "Apache-2.0"]}, "CC-BY-4.0"]}
+            // According to the spec, this seems to be wrong -
+            // the result produced here is correct.
             r#"any_of: ["MIT", { all_of: ["Apache-2.0", "CC-BY-4.0"] }]"#,
         );
     }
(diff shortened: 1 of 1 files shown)


cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -369,6 +369,9 @@ mod tests {
     fn spdx_three_licenses_or_and() {
         run_spdx_comparison(
             "MIT OR Apache-2.0 AND CC-BY-4.0",
+            // NOTE: Homebrew parses this as {:all_of=>[{:any_of=>["MIT", "Apache-2.0"]}, "CC-BY-4.0"]}
+            // According to the spec, this seems to be wrong -
+            // the result produced here is correct.
```
cxreiff, line 374: Not sure what your preferences are around comment verbosity, but I've suggested an explanation and the spec URL in case you'd like to use any of it.

```suggestion
            // NOTE: Homebrew parses this as {:all_of=>[{:any_of=>["MIT", "Apache-2.0"]}, "CC-BY-4.0"]}
            // According to the SPDX v3 spec, this seems to be wrong, as operator precedence is specified
            // as WITH > AND > OR (while Homebrew evaluates OR operators, then AND operators).
            // The result produced in this test is correct.
            //
            // https://spdx.github.io/spdx-spec/v3.0/annexes/SPDX-license-expressions/#d45-order-of-precedence-and-parentheses
```

cxreiff (approved):
Looks good to me- suggested some additional text around the specific difference in behavior in case you find it useful, but not sure how much text you want to cram in the comment.

cargo-dist/src/backend/installer/homebrew.rs
mistydemeo: Perfect, thanks!

Gankra (approved):

PR #1354, opened 2024-08-21
tests: enable `brew style` in linters

mistydemeo:
This adds a `brew style` linter to make sure that our generated Homebrew formulas will pass Homebrew's `brew style` for users who have enabled Homebrew's default CI setup for their taps. ~~I've also ensured our snapshots contain the auto-restyled copies after `brew style --fix` is run.~~

This required one stylistic change: the hashmap for binary aliases would violate the line length lint, and unfortunately Rubocop's line length autoformatter was consistently getting the formatting wrong. This instead (ab)uses the fact that JSON hashmaps have a compatible syntax to Ruby 1.9+ hashmaps so long as symbols are the keys - and we were already using symbolized keys. We can therefore use jinja's `tojson` with an indent to enable pretty-printing.

### DIFF
--- a/cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2
+++ b/cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2
@@ -38,7 +38,8 @@
 
             export PATH="/home/linuxbrew/.linuxbrew/bin:$PATH"
             brew update
-            brew style --fix "Formula/${filename}" || true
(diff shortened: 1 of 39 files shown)


mistydemeo:
Removed the restyled copies from the snapshots. It broke Linux and Windows, which don't have the ability to run `brew style --fix`.

Gankra (approved):

PR #1357, opened 2024-08-22
feat: improve staticlib handling, errors

### LINKED ISSUE #1356: Packaging static library causes "unable to run linkage report for this type of binary" error on Windows
I have the following `cargo dist` config:

```
# Config for 'cargo dist'
[workspace.metadata.dist]
# The preferred cargo-dist version to use in CI (Cargo.toml SemVer syntax)
cargo-dist-version = "0.21.0"
# CI backends to support
ci = "github"
# Which kinds of built libraries to include in the final archives
package-libraries = "cstaticlib"
# Which kinds of packaged libraries to install
install-libraries = "cstaticlib"
# The installers to generate for each app
installers = ["shell", "powershell"]
# Target platforms to build apps for (Rust target-triple syntax)
targets = ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-pc-windows-msvc"]
# Which actions to run on pull requests
pr-run-mode = "plan"
# Path that installers should place binaries in
install-path = "~/.limbo"
# Whether to install an updater program
install-updater = true
# Whether to consider the binaries in a package for distribution (defaults true)
dist = false
# Whether to enable GitHub Attestations
github-attestations = true
```

But it's failing on Windows:

```
warning: `limbo-wasm` (lib) generated 1 warning
   Compiling limbo_sim v0.0.4 (D:\a\limbo\limbo\simulator)
warning: `py-limbo` (lib) generated 4 warnings
    Finished `dist` profile [optimized] target(s) in 3m 50s
  × unable to run linkage report for this type of binary

cargo dist ran successfully
Error: Process completed with exit code 1.
```

Full log here:

https://github.com/penberg/limbo/actions/runs/10510564356/job/29119512466

mistydemeo:
* Windows staticlibs are `ar` archives, not PE executables. Make sure we handle those in the Windows object parsing case.
* Make unsupported binary checks nonfatal. In this case we can simply return a default instead of failing the build.
* Provide the filename in the LinkageCheckUnsupportedBinary error, making it more informative in the cases it does end up getting printed.

Fixes #1356.

### DIFF
--- a/cargo-dist/src/linkage.rs
+++ b/cargo-dist/src/linkage.rs
@@ -14,6 +14,7 @@ use cargo_dist_schema::{
 use comfy_table::{presets::UTF8_FULL, Table};
 use goblin::Object;
 use mach_object::{LoadCommand, OFile};
+use tracing::warn;
 
 use crate::{config::Config, errors::*, gather_work, Artifact, DistGraph};
 
@@ -81,7 +82,27 @@ fn compute_linkage_assuming_local_build(
                 if !bin_path.exists() {
                     eprintln!("Binary {bin_path} missing; skipping check");
                 } else {
-                    let linkage = determine_linkage(&bin_path, target)?;
+                    let linkage = match determine_linkage(&bin_path, target) {
+                        Ok(linkage) => linkage,
+                        // We don't want to treat the primary linkage
+                        // failure check as a failure - instead, print
+                        // a warning and provide a default.
+                        Err(DistError::LinkageCheckUnsupportedBinary { file_name }) => {
+                            if let Some(file_name) = file_name {
+                                warn!("Unable to fetch linkage from {file_name}");
+                            } else {
+                                warn!("Unable to fetch linkage");
+                            }
+                            Linkage::default()
+                        }
+                        Err(e) => {
+                            warn!(
+                                "Encountered an issue trying to determine linkage: {:?}",
+                                miette::Report::new(e)
+                            );
+                            Linkage::default()
+                        }
+                    };
 
                     manifest.assets.insert(
                         bin.id.clone(),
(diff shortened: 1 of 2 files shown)


cargo-dist/src/linkage.rs
```diff
@@ -81,7 +82,21 @@ fn compute_linkage_assuming_local_build(
...
+                        // a warning and provide a default.
+                        Err(DistError::LinkageCheckUnsupportedBinary { file_name }) => {
+                            if let Some(file_name) = file_name {
+                                warn!("Unable to fetch linkage from {file_name}");
+                            } else {
+                                warn!("Unable to fetch linkage");
+                            }
+                            Linkage::default()
+                        }
+                        Err(e) => return Err(e),
```
Gankra, line 98: Why should we allow any linkage errors to kill anything?

Gankra (approved):

cargo-dist/src/linkage.rs
mistydemeo: Good point, there are a handful of other errors here that we should probably just warn about instead of making nonfatal.

mistydemeo: Turned this into a "warn and move on" case too, just with a more generic warning than `LinkageCheckUnsupportedBinary`.

cargo-dist/src/linkage.rs
```diff
@@ -81,7 +82,24 @@ fn compute_linkage_assuming_local_build(
...
+                        Err(DistError::LinkageCheckUnsupportedBinary { file_name }) => {
+                            if let Some(file_name) = file_name {
+                                warn!("Unable to fetch linkage from {file_name}");
+                            } else {
+                                warn!("Unable to fetch linkage");
+                            }
+                            Linkage::default()
+                        }
+                        Err(e) => {
+                            warn!("Encountered an issue trying to determine linkage: {e}");
```
Gankra, line 99:
```suggestion
                            warn!("Encountered an issue trying to determine linkage: {:?}", miette::Report::new(e));
```

I think?

Gankra (approved):

Issue #469, opened 2023-10-01
installer: macOS pkg


davidkna:
I think it would be nice to support creating `pkg` installers for macOS. Compared to `dmg`/`app` this could be more suitable for CLI applications (#24). For example, [starship](https://github.com/starship/starship/tree/master/install/macos_packages) publishes `pkg` installers.

indygreg:
FYI I have a pure Rust `pkg` file format implementation at https://crates.io/crates/apple-flat-package. Unfortunately, it doesn't currently support writing because my `Bom` file implementation (https://github.com/indygreg/apple-platform-rs/tree/main/apple-bom) doesn't yet support writing. It's an unknown amount of work to finish implementing writing support.

I just wanted to let you know that generating `pkg` files without using Apple tooling is just outside the realm of possibility.

(The linked repository also has a pure Rust implementation of Apple code signing and notarization so you can sign + notarize from Linux in case that's interesting to you.)

mistydemeo:
I'm looking into this again at the moment. I'm currently looking to leverage the native Apple tools for this. My plan is to use both `pkgbuild` and `productbuild` so that we can produce a package with similar rich presentation data as in the Windows package building support.

StudentWeis:
I need it!! Plz!

PR #1361, opened 2024-08-22
feat: macOS codesigning

mistydemeo:
This adds support for codesigning on macOS, similar to the ssl.com feature on Windows. In the current revision, we only perform signing with a certificate - notarization will come in a subsequent feature.

As with ssl.com, the only feature in our config is the feature controlling whether or not to sign - the rest is handled via the environment/Actions secrets. The user needs to provide three fields:

* A pkcs12 certificate, base64-encoded.
* The passphrase for that certificate.
* The name of the codesigning identity belonging to that certificate.

If all three are present, we'll codesign any binaries and libraries during the local build process.

This uses the builtin macOS tooling, which means it can only be performed on a macOS host. (That's not too different from how we can only do the Windows codesigning on a Windows host.) Because interacting with the keychain is slightly wonky on something like a hosted runner, we've got a slightly complex process that takes a few steps. (This process is inspired by official documentation from GitHub: https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development)

* We create a fresh keychain database in a temporary directory, with a randomized password under our control.
* We unlock that keychain and configure it to remain unlocked for the duration of the signing process.
* We set that keychain as the default.
* We import our certificate into that keychain, and configure it so that it can only be used by `codesign`, `security` and `productsign`.
* We configure the key to be used for signing.
* We actually sign the binary.

I structured the core keychain stuff into a struct that can be reused, with a mind to using it in the `pkgbuild`/`productbuild` stages from #469.

Because the keychain is placed in a temporary directory under the control of `temp_dir`, it's deleted automatically after the signing process completes. That should mean that we don't even need to rely on the ephemerality of the runner to ensure that the certificate is deleted prior to the next job running.

### DIFF
--- a/cargo-dist/src/sign/macos.rs
+++ b/cargo-dist/src/sign/macos.rs
@@ -18,29 +18,28 @@
 //! 5) Let the keychain be deleted when the temporary directory is dropped.
 //!
 //! In the future, this module will also support notarization.
-use std::path::PathBuf;
-
 use axoasset::LocalAsset;
 use axoprocess::Cmd;
 use base64::Engine;
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use temp_dir::TempDir;
 use tracing::warn;
 
-use crate::{DistError, DistResult, TargetTriple};
+use crate::{create_tmp, DistError, DistResult, TargetTriple};
 
 struct Keychain {
-    root: TempDir,
+    _root: TempDir,
+    root_path: Utf8PathBuf,
     password: String,
-    pub path: PathBuf,
+    pub path: Utf8PathBuf,
 }
 
 impl Keychain {
     /// Creates a keychain in a temporary directory, secured
     /// with the provided password.
     pub fn create(password: String) -> DistResult<Self> {
-        let root = TempDir::new()?;
-        let path = root.path().join("signing.keychain-db");
+        let (root, root_path) = create_tmp()?;
+        let path = root_path.join("signing.keychain-db");
 
         let mut cmd = Cmd::new("/usr/bin/security", "create keychain");
         cmd.arg("create-keychain");
@@ -77,7 +76,8 @@ impl Keychain {
         cmd.status()?;
 
         Ok(Self {
-            root,
+            _root: root,
+            root_path,
             password,
             path,
         })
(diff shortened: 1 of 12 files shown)


cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -203,6 +203,9 @@ jobs:
     runs-on: ${{ matrix.runner }}
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      CODESIGN_CERTIFICATE: ${{ secrets.CODESIGN_CERTIFICATE }}
+      CODESIGN_CERTIFICATE_PASSWORD: ${{ secrets.CODESIGN_CERTIFICATE_PASSWORD }}
+      CODESIGN_IDENTITY: ${{ secrets.CODESIGN_IDENTITY }}
```
Gankra, line 208: I feel like these should probably be wrapped in an if..?

cargo-dist/src/sign/mod.rs
```diff
@@ -34,6 +45,18 @@ impl Signing {
                 signer.sign(file)?;
             }
         }
+        if let Some(signer) = &self.macos {
+            // TODO: restructure, this is just to keep Windows
+            // from flagging dead code
+            #[cfg(unix)]
+            let is_executable = file.metadata()?.permissions().mode() & 0o111 != 0;
```
Gankra, line 52: Are there non-executable things macos can/should sign?

cargo-dist/src/sign/macos.rs
```diff
@@ -0,0 +1,178 @@
+use std::path::PathBuf;
```
Gankra, line 1: It'd be nice to have a top-level module comment here of like the general flow/idea, or any Exciting Issues this code deals with.

Gankra (approved):
Rough gist seems right, left some random thoughts

cargo-dist/src/sign/mod.rs
mistydemeo: I'll leave a comment. At the moment, no. In the future if we end up distributing `.app` bundles or other bundles, we'll want to sign those.

The `.pkg` signing is going to be handled by `pkgbuild`/`productbuild`, so we don't need to handle that here.

cargo-dist/src/sign/macos.rs
```diff
@@ -0,0 +1,178 @@
...
+    root: TempDir,
+    password: String,
+    pub path: PathBuf,
+}
+
+impl Keychain {
+    /// Creates a keychain in a temporary directory, secured
+    /// with the provided password.
+    pub fn create(password: String) -> DistResult<Self> {
+        let root = TempDir::new()?;
```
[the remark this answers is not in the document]
mistydemeo: Swapped over to the same shared `create_tmp()` as in the other PR; made it a separate commit for ease of merge conflict stuff, depending on which makes it in first

Gankra (approved):
seems legit?

PR #1373, opened 2024-08-27
fix: improve powershell installer portability

Gankra:
* pass -ExecutionPolicy ByPass to run on default-security windows installs
* clarify restart system instead of shell
* suggest an expr for fixing the current session


Gankra:
currently draft because the expr is a WIP placeholder -- need to settle on a final one.

mistydemeo (approved):
Do we need to ensure that axoupdater provides the same argument to `powershell`?

Gankra:
It should, yes.

Issue #1375, opened 2024-08-27
rename CARGO_DIST_FORCE_INSTALL_DIR


Gankra:
Specifically support both for a while (forever)?

* minimal version: `FORCE_INSTALL_DIR`
* maybe ideal version: app_name.to_upper_case() + "_FORCE_INSTALL_DIR"

axoupdater interop is an issue here, but if we keep the old env-var working we don't need to deal with it immediately.

Gankra:
Needs to be done in shell and powershell installer

Gankra:
A similar rename should probably happen for the special target env-vars cargo-dist uses to talk to bun scripts (the script pretends a renamed var exists but falls back to the one that actually exists)

mistydemeo:
Option 2 seems ideal to me, and very easy for us to produce something all our environments can agree on. In the short term, I think ideally we should

a) Have the installer respect both variables, and
b) Have axoupdater set both variables.

Gankra:
shipped in 0.22.0

PR #1377, opened 2024-08-28
feat: support app-specific force install

mistydemeo:
This supports a new app-specific force install environment variable. We use the app's name, uppercased and with `-` replaced by `_`. For example, if the app name is `akaikatana-repack`, the new force install environment variable becomes `AKAIKATANA_REPACK_FORCE_INSTALL_DIR`.

This supports the new variable alongside the old one. In the longterm, we could consider removing the old one after a transition period where both are supported.

An axolotlsay PR to use the new environment variable is here: https://github.com/axodotdev/axoupdater/pull/169

refs #1375.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -291,6 +291,10 @@ pub struct Release {
     /// The version of the app
     // FIXME: should be a Version but JsonSchema doesn't support (yet?)
     pub app_version: String,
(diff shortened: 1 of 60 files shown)


Gankra (approved):
hmm increasingly feel like we should be computing this env-var up in rust code and passing it down... jinja and rust disagreeing is a spooky prospect -- also this way we can also shove it in a dist-manifest.json or something.

PR #1379, opened 2024-08-28
feat: powershell respects NoModifyPath from env

mistydemeo:
This is supported from the environment in the shell installer, but was previously only supported on the CLI in the powershell installer.

This will be useful to ensure axoupdater tests don't pollute the environment: https://github.com/axodotdev/axoupdater/pull/168

refs #1374.


Gankra (approved):
is `-not` the right operator here still..?

mistydemeo:
I think so? Empty string is also falsy. Does the CLI parsing here convert to true/false?

```
PS /Users/mistydemeo> -not $null
True
PS /Users/mistydemeo> -not ""
True
PS /Users/mistydemeo> -not "0"
False
PS /Users/mistydemeo> -not "1"
False
```

PR #1381, opened 2024-08-28
feat: pkg followups

mistydemeo:
This contains two commits broken out from #469 so that we can ship them later: the documentation, and the user-facing `init` code. This can be merged later so that, if we decide to, we can ship unadvertised pkg support early and then make it officially supported later.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -18,6 +18,7 @@
   - [npm](./installers/npm.md)
   - [homebrew](./installers/homebrew.md)
   - [msi](./installers/msi.md)
+  - [pkg](./installers/pkg.md)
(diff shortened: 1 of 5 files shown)


ashleygwilliams:
is this something we can merge? i know the code signing part is still not possible, but is this?

mistydemeo:
I think so, yeah. Because of the init code, rebasing this is going to make the most sense after config 1.0 is merged.

mistydemeo:
This conflicts, very badly, with #1653 and so this shouldn't be merged until that PR is done. The init code here will have to be 100% rewritten in order to be usable. It follows the patterns of the surrounding code, so it should ideally be obvious how to rewrite it to work with the new config structure once the surrounding code has been rewritten and is merged.

PR #1382, opened 2024-08-28
fix: give instructions for cmd too

Gankra:


cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -381,7 +381,8 @@ function Invoke-Installer($artifacts, $platforms) {
         Write-Information ""
         Write-Information "To add $dest_dir to your PATH, either restart your system or run:"
         Write-Information ""
-        Write-Information "    `$env:Path = `"$dest_dir;`$env:Path`""
+        Write-Information "    `set Path=$dest_dir;%Path%   (cmd)"
```
mistydemeo, line 384: Is the backtick the escape character? If so, do we actually need to escape the start of `set`?

mistydemeo (approved):

cargo-dist/templates/installer/installer.ps1.j2
Gankra: wow i can't believe you caught that by eye (you are correct and fixed)

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -381,7 +381,8 @@ function Invoke-Installer($artifacts, $platforms) {
         Write-Information ""
         Write-Information "To add $dest_dir to your PATH, either restart your system or run:"
         Write-Information ""
-        Write-Information "    `$env:Path = `"$dest_dir;`$env:Path`""
+        Write-Information "    set Path=$dest_dir;%Path%   (cmd)"
```
mistydemeo, line 384: We don't have to escape `%Path%`, right? I guess that's not powershell syntax

mistydemeo (approved):

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Turns out I'm hyper-sensitive to mismatched backticks from bash land 😅

cargo-dist/templates/installer/installer.ps1.j2
Gankra: yeah powershell is very %percentsyntaxwho%

Gankra:
I am skittish about this PR potentially not handling something like spaces in paths, I don't have time to think through and test it.

PR #1390, opened 2024-09-03
fix: make linkage truly infallible

Gankra:
also some driveby cleanups

### DIFF
--- a/cargo-dist/src/build/mod.rs
+++ b/cargo-dist/src/build/mod.rs
@@ -173,7 +173,7 @@ impl BuildExpectations {
                 let bin = dist.binary(result_bin.idx);
 
                 // compute linkage for the binary
-                self.compute_linkage(dist, manifest, result_bin, &bin.target)?;
+                self.compute_linkage_and_sign(dist, manifest, result_bin, &bin.target)?;
 
                 // copy files to their final homes
                 self.copy_assets(result_bin, bin)?;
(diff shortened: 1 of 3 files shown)


cargo-dist/src/build/mod.rs
```diff
@@ -215,7 +215,7 @@ impl BuildExpectations {
             });
             linkage
         } else {
-            determine_linkage(src_path, target)?
+            determine_linkage(src_path, target)
```
ashleygwilliams, line 218: this is the diff i was looking for haha

cargo-dist/src/errors.rs
```diff
@@ -303,11 +303,8 @@ pub enum DistError {
     },
 
     /// Linkage report can't be run for this target
-    #[error("unable to run linkage report for this type of binary; file was: {file_name:?}")]
-    LinkageCheckUnsupportedBinary {
-        /// The object's file name
-        file_name: Option<String>,
-    },
+    #[error("unable to run linkage report for this type of binary")]
+    LinkageCheckUnsupportedBinary,
```
ashleygwilliams, line 307: i do think maintaining the name of the file would be useful

ashleygwilliams (commented):
in general looks good, brought up in discord but does seem strange to group linkage and signing to me, but this might be because i am missing context?

cargo-dist/src/errors.rs
Gankra: The warning all of these flow into handles reporting the filename so it ends up being redundant noise

PR #1395, opened 2024-09-04
feat: make dist-workspace.toml able to be a package

Gankra:


Gankra:
This is currently a draft because I need to modernize some of the test-suite.

PR #1401, opened 2024-09-09
Use singular quotes, alternatively we should escape the !

ThomasTNO:
Minor correction in instructions


Gankra (approved):
ah, ty so much!

PR #1403, opened 2024-09-09
Use a more recent version of Ubuntu as the Linux runner

esteve:
This PR updates the defaults so that it uses the latest LTS version of Ubuntu 24.04, instead of 20.04 which will be EOL'd next year.

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -497,7 +497,7 @@ fn distribute_targets_to_runners_split<'a>(
 /// A string representing a Github Runner
 type GithubRunner = String;
 /// The Github Runner to use for Linux
-const GITHUB_LINUX_RUNNER: &str = "ubuntu-20.04";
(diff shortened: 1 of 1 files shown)


Gankra:
We definitely can't bump up to 24.04 -- we need to be using the oldest possible to ensure people's linux builds run as many machines as possible, per this recent blog post:

https://blog.axo.dev/2024/08/glibc

I am monitoring the 20.04 situation. It will only *start* its deprecation process when 24.04 becomes the default ("latest") ubuntu runner, but [it's still in beta](https://github.com/actions/runner-images/?tab=readme-ov-file#available-images) (aiui due to some compat issues they're still hammering out).

When 20.04 is marked as deprecated we'll bump the default runner to 22.04.

If, for whatever reason, you need to Not be on 20.04 Right Now, you can [use the custom runner feature](https://opensource.axo.dev/cargo-dist/book/ci/customizing.html#custom-runners).

esteve:
@Gankra thanks for the explanation, makes total sense to keep with the oldest available Ubuntu distro.

PR #1410, opened 2024-09-11
feat: unmanaged installs

### LINKED ISSUE #1400: create "ci" mode for ephemeral environments, with bare minimum env assumptions
The installer does several things that are useful/important for the full breadth of its supported usecases, but which are unnecessary or detrimental for folks who want to take control of the installation process. As such, we need a way to reliably tell the installer "don't do anything clever, and don't error out for things that don't really matter here".

The original proposal was to just make it non-fatal for certain operations like the install-receipt to be non-fatal to setup. The only consequence of a failed receipt is the updater gets bricked. In really minimal CI-like systems the updater doesn't matter, so we should not fall over if the receipt fails.

However as we discussed this more, we realized there's more kinds of usecases and something more comprehensive/intentional was desirable.

mistydemeo:
The new "unmanaged" installation mode allows for something close to "untar and move on" installations. It disables modifying the `PATH` and creation of environment files, and ensures all files are installed into a single flat directory no matter what other configuration is in place. It also disables installing a standalone updater and the creation of an install receipt.

The new "don't install an updater/receipt" behaviour is also controllable behind a new public environment variable, `$APPNAME_DISABLE_UPDATE`.

Fixes #1400.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -283,6 +283,19 @@ pub struct SystemInfo {
     pub build_environment: BuildEnvironment,
 }
 
+/// Release-specific environment variables
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct EnvironmentVariables {
+    /// Environment variable to force an install location
+    pub install_dir_env_var: String,
+    /// Environment variable to force an unmanaged install location
+    pub unmanaged_dir_env_var: String,
+    /// Environment variable to disable updater features
+    pub disable_update_env_var: String,
+    /// Environment variable to disable modifying the path
+    pub no_modify_path_env_var: String,
+}
+
 /// A Release of an Application
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct Release {
@@ -291,10 +304,10 @@ pub struct Release {
     /// The version of the app
     // FIXME: should be a Version but JsonSchema doesn't support (yet?)
     pub app_version: String,
-    /// Environment variable to force an install location
+    /// Environment variables which control this release's installer's behaviour
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub install_dir_env_var: Option<String>,
+    pub env: Option<EnvironmentVariables>,
     /// Alternative display name that can be prettier
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -610,12 +623,23 @@ impl DistManifest {
         if let Some(position) = self.releases.iter().position(|r| r.app_name == name) {
             &mut self.releases[position]
         } else {
-            let install_dir_env_var =
-                Some(name.to_ascii_uppercase().replace('-', "_") + "_INSTALL_DIR");
+            let env_app_name = name.to_ascii_uppercase().replace('-', "_");
+            let install_dir_env_var = format!("{env_app_name}_INSTALL_DIR");
+            let unmanaged_dir_env_var = format!("{env_app_name}_UNMANAGED_INSTALL");
+            let disable_update_env_var = format!("{env_app_name}_DISABLE_UPDATE");
+            let no_modify_path_env_var = format!("{env_app_name}_NO_MODIFY_PATH");
+
+            let environment_variables = EnvironmentVariables {
+                install_dir_env_var,
+                unmanaged_dir_env_var,
+                disable_update_env_var,
+                no_modify_path_env_var,
+            };
+
             self.releases.push(Release {
                 app_name: name,
                 app_version: version,
-                install_dir_env_var,
+                env: Some(environment_variables),
                 artifacts: vec![],
                 hosting: Hosting::default(),
                 display: None,
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -61,7 +61,13 @@ if ($env:{{ disable_update_env_var }}) {
   $install_updater = $true
 }
 
+$NoModifyPath = $false
+if ($env:{{ no_modify_path_env_var }}) {
+    $NoModifyPath = $true
+}
+
 $unmanaged_install = $env:{{ unmanaged_dir_env_var }}
+
 if ($unmanaged_install) {
   $NoModifyPath = $true
   $install_updater = $false
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -55,6 +55,18 @@ $receipt = @"
 "@
 $receipt_home = "${env:LOCALAPPDATA}\{{ app_name }}"
 
+if ($env:{{ disable_update_env_var }}) {
+  $install_updater = $false
+} else {
+  $install_updater = $true
+}
+
+$unmanaged_install = $env:{{ unmanaged_dir_env_var }}
+if ($unmanaged_install) {
+  $NoModifyPath = $true
+  $install_updater = $false
+}
+
 function Install-Binary($install_args) {
   if ($Help) {
     Get-Help $PSCommandPath -Detailed
@@ -219,7 +231,7 @@ function Download($download_url, $platforms) {
     $staticlib_paths += "$tmp\$lib_name"
   }
 
-  if ($null -ne $info["updater"]) {
+  if (($null -ne $info["updater"]) -and $install_updater) {
     $updater_id = $info["updater"]["artifact_name"]
     $updater_url = "$download_url/$updater_id"
     $out_name = "$tmp\{{ app_name }}-update.exe"
@@ -248,12 +260,18 @@ function Invoke-Installer($artifacts, $platforms) {
 
   # Forces the install to occur at this path, not the default
   $force_install_dir = $null
+  $flat_install_layout = $false
   # Check the newer app-specific variable before falling back
   # to the older generic one
   if (($env:{{ install_dir_env_var }})) {
     $force_install_dir = $env:{{ install_dir_env_var }}
+    $flat_install_layout = {% if install_paths | selectattr("kind", "equalto", "CargoHome") -%} $false {%- else -%} $true {%- endif %}
   } elseif (($env:CARGO_DIST_FORCE_INSTALL_DIR)) {
     $force_install_dir = $env:CARGO_DIST_FORCE_INSTALL_DIR
+    $flat_install_layout = {% if install_paths | selectattr("kind", "equalto", "CargoHome") -%} $false {%- else -%} $true {%- endif %}
+  } elseif ($unmanaged_install) {
+    $force_install_dir = $unmanaged_install
+    $flat_install_layout = $true
   }
 
   # The actual path we're going to install to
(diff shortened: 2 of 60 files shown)


ashleygwilliams:
q: does this still attempt receipt creation? i think it'd be best to just not try at all

mistydemeo:
The issue sort of identified a conflict between the two user groups this attempts to serve - one would care about receipts, the other is neutral on them. I tried to split the difference by *attempting* receipt creation, but treating receipt creation failure as harmless.

ashleygwilliams:
hrm, i think splitting the difference here is maybe the less good option and we should consider 2 modes

ashleygwilliams:
part of the reason i say this is that an unmanaged install is desirable for ephemeral environments which are usually automated- receipt creation is an i/o operation and while not particularly expensive it's not free- this feature in my mind is specifically for the "conductor" persona and they are the type that care about milliseconds

mistydemeo:
Based on our previous discussion, I've made a few changes:

* We no longer split the difference between two groups of users; this is now completely focused on automated/CI usecases, and the updater/receipt is unconditionally disabled.
* The updater disabling functionality can be enabled separately with a new environment variable.
* The environment variables have been moved into a new grouping to make reasoning about them as a group easier:

```diff
-      "install_dir_env_var": "CARGO_DIST_SCHEMA_INSTALL_DIR",
+      "environment_variables": {
+        "install_dir_env_var": "CARGO_DIST_SCHEMA_INSTALL_DIR",
+        "unmanaged_dir_env_var": "CARGO_DIST_SCHEMA_UNMANAGED_INSTALL",
+        "disable_update_env_var": "CARGO_DIST_SCHEMA_DISABLE_UPDATE"
+      },
```

ashleygwilliams:
we could probably name that `env` for brevity's sake

mistydemeo:
Sounds good, done.

```diff
-      "environment_variables": {
+      "env": {
```

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -55,6 +55,18 @@ $receipt = @"
...
 
+if ($env:{{ disable_update_env_var }}) {
+  $install_updater = $false
+} else {
+  $install_updater = $true
+}
+
+$unmanaged_install = $env:{{ unmanaged_dir_env_var }}
+if ($unmanaged_install) {
+  $NoModifyPath = $true
```
Gankra, line 66: We should be deleting the -NoModifyPath `param`, people keep finding these weird inconsistent APIs we just shouldn't support them

cargo-dist-schema/src/lib.rs
```diff
@@ -610,12 +621,21 @@ impl DistManifest {
         if let Some(position) = self.releases.iter().position(|r| r.app_name == name) {
             &mut self.releases[position]
         } else {
-            let install_dir_env_var =
-                Some(name.to_ascii_uppercase().replace('-', "_") + "_INSTALL_DIR");
+            let env_app_name = name.to_ascii_uppercase().replace('-', "_");
+            let install_dir_env_var = format!("{env_app_name}_INSTALL_DIR");
+            let unmanaged_dir_env_var = format!("{env_app_name}_UNMANAGED_INSTALL");
+            let disable_update_env_var = format!("{env_app_name}_DISABLE_UPDATE");
```
Gankra, line 627: not DISABLE_UPDATER?

Gankra, line 627: Should have *_NO_MODIFY_PATH in here too

cargo-dist-schema/src/lib.rs
```diff
@@ -610,12 +621,21 @@ impl DistManifest {
...
+
```
Gankra, line 628: We should maybe also expose FLAT_INSTALL_LAYOUT as a raw option too? less certain of that

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -55,6 +55,18 @@ $receipt = @"
 "@
 $receipt_home = "${env:LOCALAPPDATA}\{{ app_name }}"
 
+if ($env:{{ disable_update_env_var }}) {
```
Gankra, line 58: We need to be checking INSTALLER_INSTALL_UPDATE(R) as well here.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -22,6 +22,19 @@ ARTIFACT_DOWNLOAD_URL="${INSTALLER_DOWNLOAD_URL:-{{ base_url }}}"
 PRINT_VERBOSE=${INSTALLER_PRINT_VERBOSE:-0}
 PRINT_QUIET=${INSTALLER_PRINT_QUIET:-0}
 NO_MODIFY_PATH=${INSTALLER_NO_MODIFY_PATH:-0}
+ALLOW_RECEIPTS_TO_FAIL=0
```
Gankra, line 25: I don't think we need/want to add this -- certainly the powershell installer doesn't so this is just making it more inconsistent?

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -22,6 +22,19 @@ ARTIFACT_DOWNLOAD_URL="${INSTALLER_DOWNLOAD_URL:-{{ base_url }}}"
...
+if [ "{{ '${' }}{{ disable_update_env_var }}:-0}" = "1" ]; then
```
Gankra, line 26: need to be checking an INSTALLER_ variant as well

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -22,6 +22,19 @@ ARTIFACT_DOWNLOAD_URL="${INSTALLER_DOWNLOAD_URL:-{{ base_url }}}"
...
+    INSTALL_UPDATER=0
+else
+    INSTALL_UPDATER=1
+fi
+UNMANAGED_INSTALL="{{ '${' }}{{ unmanaged_dir_env_var }}:-}"
```
Gankra, line 31: need to be checking an INSTALLER_ variant as well

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -241,12 +254,23 @@ download_binary_and_run_installer() {
...
-    # shellcheck disable=SC2320
-    local _retval=$?
+    if [ "$INSTALL_UPDATER" = "1" ]; then
+        if ! mkdir -p "$RECEIPT_HOME"; then
+            if [ "$ALLOW_RECEIPTS_TO_FAIL" = "1" ]; then
+                say "failed to create an install receipt in $RECEIPT_HOME"
+                say "this installation will not be able to self-update"
+                local _retval=0
+            else
+                err "unable to create receipt directory at $RECEIPT_HOME"
```
Gankra, line 264: "install receipt"

Gankra (changes requested):

cargo-dist-schema/src/lib.rs
ashleygwilliams: my thought here was that the user cares about the "update" behavior. there is not always an "updater" so "update" focused on the behavior the user understands, not the impl

cargo-dist/templates/installer/installer.ps1.j2
Gankra: disregard this, we are explicitly *not* adding generic INSTALLER_ vars in this PR

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -267,13 +285,13 @@ function Invoke-Installer($artifacts, $platforms) {
...
   # if we're overriding it.
   if (($force_install_dir)) {
-{% if install_paths| selectattr("kind", "equalto", "CargoHome") %}
-    $dest_dir = Join-Path $force_install_dir "bin"
-    $dest_dir_lib = $dest_dir
-{%- else -%}
-    $dest_dir = $force_install_dir
-    $dest_dir_lib = $dest_dir
-{%- endif %}
+    if (!$flat_install_layout) {
```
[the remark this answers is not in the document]
mistydemeo: Wow yeah, good point. Well, I'll swap to `-not` for consistency

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Yeah, this isn't properly used anymore so I'll take it out.

cargo-dist-schema/src/lib.rs
mistydemeo: How about we circle back around to publicizing this around the time we also give developers full control over the install layout?

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -55,6 +50,24 @@ $receipt = @"
 "@
 $receipt_home = "${env:LOCALAPPDATA}\{{ app_name }}"
 
+if ($env:{{ disable_update_env_var }}) {
+  $install_updater = $false
+} else {
+  $install_updater = $true
+}
+
+$NoModifyPath = $false
```
Gankra, line 59: format of this variable name is weird now but not a big deal?

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -55,6 +50,24 @@ $receipt = @"
...
+if ($env:{{ no_modify_path_env_var }}) {
+    $NoModifyPath = $true
+}
+
+$unmanaged_install = $env:{{ unmanaged_dir_env_var }}
+
+if ($unmanaged_install) {
+  $NoModifyPath = $true
+  $install_updater = $false
+}
```
Gankra, line 69: I think the "mode" variable should get applied first, and then the other overrides should kick in..?

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -21,7 +21,22 @@ APP_VERSION="{{ app_version }}"
...
+if [ "{{ '${' }}{{ disable_update_env_var }}:-0}" = "1" ]; then
+    INSTALL_UPDATER=0
+else
+    INSTALL_UPDATER=1
+fi
+UNMANAGED_INSTALL="{{ '${' }}{{ unmanaged_dir_env_var }}:-}"
+if [ -n "${UNMANAGED_INSTALL}" ]; then
+    NO_MODIFY_PATH=1
+    INSTALL_UPDATER=0
+fi
```
Gankra, line 38: ditto on mode should go first..?

Gankra (approved):
think we're basically there, a few more nits

cargo-dist/templates/installer/installer.sh.j2
mistydemeo: Honestly I think "unmanaged mode" is opinionated enough that it probably should override the individual options instead of the other way around. I see the individual options as being useful in normal mode, not for this.

mistydemeo:
I just realized that the shell installer *does* provide a CLI param for `--no-modify-path`, so I've removed the change where I removed that feature from powershell. Instead, I've just added the env override while retaining CLI support.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Marking as resolved since I've decided to keep the param for consistency with shell.

PR #1419, opened 2024-09-18
feat: track install layout in receipt

mistydemeo:
In preparation for allowing this to be configured by devs and overridden by users, this starts tracking the installation layout in install receipts. Doing this makes sure we can tell the updater what layout to use in the future, even if the desired layout isn't the default for the software.

Right now we have two kinds:
* "flat" - all files in a single directory
* "hierarchical" - files separated into subdirectories
* "cargo-home" - like hierarchical, but with only a "bin" subdirectory


Gankra (approved):
This is interesting and nothing seems to jump out to me...

PR #1420, opened 2024-09-18
fix(ci): always serialize upload_files

### LINKED ISSUE #1387: cargo-dist autoselects non-GitHub remote and fails
The problem I'm experiencing is this:
```
  × GitHub CI support requires a GitHub repository
  ╰─▶   × Your repository URL https://<not github> couldn't be parsed.
        help: Only GitHub URLs are supported at the moment.
```

The repo in question has two remotes: one `origin` that points at something that's not GitHub and one `github` that *does* point at GitHub.

Is there a way for me to tell cargo-dist to just use the GH remote?

mistydemeo:
While it's usually an error if we received a totally empty set of artifacts, it's possible in the global build job under a very specific set of conditions:

* There are no installers, *and*
* The source tarball was disabled

In this case we need to make sure we still provide an upload_files array in the manifest so that we can always index into it.

Fixes #1387.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -106,7 +106,8 @@ pub struct DistManifest {
     pub linkage: Vec<Linkage>,
     /// Files to upload
     #[serde(default)]
-    #[serde(skip_serializing_if = "Vec::is_empty")]
+    // We need to make sure we always serialize this when it's empty,
+    // because we index into this array unconditionally during upload.
     pub upload_files: Vec<String>,
     /// Whether Artifact Attestations should be found in the GitHub Release
     ///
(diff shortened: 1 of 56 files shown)


cargo-dist-schema/src/lib.rs
```diff
@@ -106,7 +106,6 @@ pub struct DistManifest {
     pub linkage: Vec<Linkage>,
     /// Files to upload
```
ashleygwilliams, line 107: do you think we may want to leave a comment here about why we want to serialize the empty array?

mistydemeo: Good callout, done.

Gankra (approved):

PR #1421, opened 2024-09-18
feat: track NO_MODIFY_PATH for updates

mistydemeo:
An updater-side change can keep track of this from the receipt and use it to decide whether to pass the same flag to the updater.

Fixes https://github.com/astral-sh/uv/issues/7319.
Fixes https://github.com/astral-sh/uv/issues/5576.
Fixes https://github.com/axodotdev/axoupdater/issues/174.


Gankra (approved):

mistydemeo:
Tested and working on Mac and Windows.

PR #1422, opened 2024-09-18
doc: address most config TODOs

Gankra:


cargo-dist/src/config/v1/installers/pkg.rs
```diff
@@ -35,8 +35,12 @@ impl PkgInstallerConfig {
         _pkg_idx: PackageIdx,
         common: &CommonInstallerConfig,
     ) -> Self {
+
         Self {
             common: common.clone(),
+            // TODO: you *need* to provide this to make a pkg
+            // installer, so it *should* be non-optional, but the
+            // whole "defaults first" thing makes this messed up...
             identifier: "TODO".to_owned(),
```
Gankra, line 44: This one still persists, as it was a detail in the pkg rebase I needed to breeze past. With correct config this doesn't matter as it always gets overloaded, but with incorrect config we'll get confused and have bad diagnostics.

mistydemeo (approved):
Makes sense to me - just need to handle the `cargo fmt` and `clippy` stuff.

PR #1423, opened 2024-09-18
feat: support any format in npm

### LINKED ISSUE #226: [npm-installer] support more than tar.gz
cargo-dist currently lets you select your binary archive format:

* .zip (default windows)
* .tar.gz
* .tar.xz (default unix)
* .tar.zstd

But the npm installer can only handle .tar.gz! This is a limitation inherited from [the binary-install npm package](https://github.com/EverlastingBugstopper/binary-install) (yes the "fork" is the Real One), which in turn inherits the limitation [from the npm tar package](https://github.com/isaacs/node-tar) which in turn inherits the limitation from "node-js bundles a gzip encoder/decoder and that's all we use".

As a workaround I will be making the new interactive UX default to .tar.gz compression if you enable npm installers, but we should really deal with by fixing/forking/ditching binary-install.

mistydemeo:
This removes the node tar library, and instead commits to unzipping/untarring files the same way that we do in the shell and powershell installers.

Fixes #226.

This isn't tested yet; opening a PR to make sure I'm not completely off base in the JS-side formatting/logic.

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -7,8 +7,8 @@ use serde::Deserialize;
 
 use crate::{
     config::{
-        self, CiStyle, CompressionImpl, Config, DistMetadata, HostingStyle, InstallPathStrategy,
-        InstallerStyle, MacPkgConfig, PublishStyle, ZipStyle,
+        self, CiStyle, Config, DistMetadata, HostingStyle, InstallPathStrategy, InstallerStyle,
+        MacPkgConfig, PublishStyle,
     },
     do_generate,
     errors::{DistError, DistResult},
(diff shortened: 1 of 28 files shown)


cargo-dist/src/init.rs
```diff
@@ -786,31 +786,6 @@ fn get_new_dist_metadata(
             }
             eprintln!();
         }
-
-        // FIXME (#226): If they have an npm installer, force on tar.gz compression
```
Gankra, line 790: 😻

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -90,11 +91,43 @@ class Package {
...
+            let tempFile = join(directory, filename);
+            const sink = res.data.pipe(createWriteStream(tempFile));
+            sink.on("error", (err) => reject(err));
+            sink.on("finish", () => {
+              if (/\.tar\.*/.test(filename)) {
+                const result = spawnSync("tar", [
+                  "xf",
+                  tempFile,
+                  "--strip-components",
+                  "1",
```
Gankra, line 105: this is often good to mention why we do this (and not for zip)

cargo-dist/src/init.rs
Gankra: Is there no similar code in backend/installers/npm.rs or tasks.rs?

mistydemeo:
Current status: this works fine on macOS. On Windows, I'm getting a race condition where the ZIP doesn't seem to be (fully?) written at the time we try to unzip it, but delaying the process even slightly by, say, doing a `fs.statSync` call on the ZIP file before unzipping causes it to succeed.

```
Downloading release from https://mistydemeo.artifacts.axodotdev.host/axolotlsay/ax_MbxL0uIRST_tFQjWW717x/axolotlsay-x86_64-pc-windows-msvc.zip
Error fetching release: An error occurred unzipping the artifact: stdout: ; stderr: [C:/Users/misty/AppData/Local/Temp/4mq3nj/axolotlsay-x86_64-pc-windows-msvc.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in C:/Users/misty/AppData/Local/Temp/4mq3nj/axolotlsay-x86_64-pc-windows-msvc.zip,
        and cannot find C:/Users/misty/AppData/Local/Temp/4mq3nj/axolotlsay-x86_64-pc-windows-msvc.zip.zip, period.
```

Sample npm package is here: https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.243

mistydemeo:
Got it working. Looks like I needed to wait on the `closed` event instead of the `finished` event for the download; with that change, it now works perfectly on Windows.

mistydemeo:
Tested with multiple configurations, and it's working fine.

Default settings (ZIP on Windows, .tar.gz on Unix): https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.245

.tar.gz: https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.246

.tar.xz: https://github.com/mistydemeo/cargodisttest/releases/tag/v0.2.247

PR #1426, opened 2024-09-19
feat: adjust new config opt-in

mistydemeo:
Based on a conversation with @ashleygwilliams, a few changes to the new config format stuff in `cargo dist init`.

* Removes the `--opt-in-migrate` flag
* New projects always get the new config
* Existing projects are offered the chance to migrate, with the default being no
* Existing projects using `cargo dist init -y` don't get migrated

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -26,8 +26,6 @@ pub struct InitArgs {
     pub with_json_config: Option<Utf8PathBuf>,
     /// Hosts to enable
     pub host: Vec<HostingStyle>,
-    /// Whether to migrate to the new config format
-    pub migrate_to_new_config: bool,
 }
 
 /// Input for --with-json-config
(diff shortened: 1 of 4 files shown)


cargo-dist/src/init.rs
```diff
@@ -109,34 +109,30 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
-            newly_initted_generic = true;
-            WorkspaceKind::Generic
-        // Already-initted users should be asked whether to migrate.
-        } else if root_workspace.kind == WorkspaceKind::Rust && args.migrate_to_new_config {
-            let prompt = r#"Would you like to opt in to the new configuration format?
+    let desired_workspace_kind = if root_workspace.kind == WorkspaceKind::Rust && !initted {
+        newly_initted_generic = true;
+        WorkspaceKind::Generic
+    // Already-initted users should be asked whether to migrate.
+    } else if root_workspace.kind == WorkspaceKind::Rust && std::io::stdin().is_terminal() {
```
Gankra, line 116: I don't recall anywhere we do this kind of check, why is this not a place for the standard --yes checks?

cargo-dist/src/init.rs
```diff
@@ -109,34 +109,30 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
-            let res = if args.yes {
-                // We want to avoid --yes pulling it in at this point.
-                false
-            } else {
-                dialoguer::Confirm::with_theme(&theme())
-                    .with_prompt(prompt)
-                    .default(args.migrate_to_new_config)
-                    .interact()?
-            };
+        let res = {
```
Gankra, line 120: this extra block is weird now

cargo-dist/src/init.rs
mistydemeo: Yeah, good point. Changed.

At some point it might be good for us to handle stdinless runs if `--yes` isn't passed, but that's not today.

Gankra (approved):

PR #1430, opened 2024-09-25
fix: handle final config todo

mistydemeo:


Gankra (approved):
good, but it would be nice if we could error early enough that plan fails

mistydemeo:
Shipped a change that gets us to error out during plan.

```
  × No bundle identifier was specified
  help: Please either enter a bundle identifier, or disable the Mac .pkg
```

This message will get updated with a link to the config once we get to the point of shipping the pkg docs.

PR #1432, opened 2024-09-27
feat: add minimum supported axoupdater check (library)

### LINKED ISSUE #1392: Add "minimum supported axoupdater version" checks
While many installer changes can be done in a forward/backward-compatible way with axoupdater, some changes, like #1355, strictly require the installed version of axoupdater to be new enough to be aware of the new behaviour. As such they are dangerous to land without some way to tell cargo-dist users "hey you must update axoupdater".

Given we already use guppy/cargo-metadata, we have dependency info at our fingertips and can in fact complain that you're using an older version of axoupdater.

So this feature amounts to:

* add some constant in cargo-dist that records the minimum supported axoupdater version
* for each distable package, if we see it depends on axoupdater *as a library*, complain if the axoupdater version is < the constant

We don't currently process dependency information so someone will need to write that logic. I'm reasonably familiar with the relevant APIs.

mistydemeo:
This adds a check when building Rust software which uses axoupdater as a library. If the user is using a version older than what this version of cargo-dist was for, it emits a warning at buildtime. I tested it on cargo-dist itself by bumping the const to something that doesn't exist, and confirmed it works:

```
 WARN Your project uses axoupdater as a library. We've detected that you're using 0.7.2, but this version of cargo-dist expects a version of at least 0.10.0 (or greater). Please consider updating to ensure your users don't experience inconsistent update behaviour.
```

To support this, I added support in axoproject for detecting if axoupdater is present and, if it is, its version is recorded. I briefly considered tracking the full dep tree in the axoproject info but that seemed like overkill.

Fixes #1392.

### DIFF
--- a/axoproject/src/rust.rs
+++ b/axoproject/src/rust.rs
@@ -79,7 +79,7 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
     let root_auto_includes = crate::find_auto_includes(workspace_root)?;
     let mut all_package_info = vec![];
     for package in members.packages(DependencyDirection::Forward) {
-        let mut info = package_info(workspace_root, &package)?;
+        let mut info = package_info(workspace_root, &package, pkg_graph)?;
         crate::merge_auto_includes(&mut info, &root_auto_includes);
         all_package_info.push(info);
     }
@@ -104,7 +104,11 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
     })
 }
 
-fn package_info(_workspace_root: &Utf8Path, package: &PackageMetadata) -> Result<PackageInfo> {
+fn package_info(
+    _workspace_root: &Utf8Path,
+    package: &PackageMetadata,
+    pkg_graph: &PackageGraph,
+) -> Result<PackageInfo> {
     let manifest_path = package.manifest_path().to_owned();
     let package_root = manifest_path
         .parent()
@@ -198,18 +202,20 @@ fn package_info(_workspace_root: &Utf8Path, package: &PackageMetadata) -> Result
             )
         };
 
-    let axoupdater_version = if package.name() == "axoupdater" {
-        Some(Version::Cargo(package.version().to_owned()))
-    } else {
-        let mut version = None;
-        for subpackage in package.direct_links() {
-            if subpackage.dep_name() == "axoupdater" {
-                version = Some(Version::Cargo(subpackage.to().version().to_owned()));
+    let query = pkg_graph.query_forward(std::iter::once(package.id()))?;
+    let package_set = query.resolve();
+    let mut axoupdater_versions = vec![];
+    for package in package_set.packages(DependencyDirection::Reverse) {
+        if package.name() == "axoupdater" {
+            axoupdater_versions.push(Version::Cargo(package.version().to_owned()))
+        } else {
+            for subpackage in package.direct_links() {
+                if subpackage.dep_name() == "axoupdater" {
+                    axoupdater_versions.push(Version::Cargo(subpackage.to().version().to_owned()))
+                }
             }
         }
-
-        version
-    };
+    }
 
     let version = Some(Version::Cargo(package.version().clone()));
     let mut info = PackageInfo {
@@ -242,7 +248,7 @@ fn package_info(_workspace_root: &Utf8Path, package: &PackageMetadata) -> Result
         cargo_package_id,
         npm_scope: None,
         build_command: None,
-        axoupdater_version,
+        axoupdater_versions,
     };
 
     // Find files we might want to auto-include
--- a/axoproject/src/rust.rs
+++ b/axoproject/src/rust.rs
@@ -52,8 +52,6 @@ pub fn get_workspace(start_dir: &Utf8Path, clamp_to_dir: Option<&Utf8Path>) -> W
 fn package_graph(start_dir: &Utf8Path) -> Result<PackageGraph> {
     let mut metadata_cmd = MetadataCommand::new();
 
-    // We don't care about dependency information, and disabling it makes things much faster!
-    metadata_cmd.no_deps();
     metadata_cmd.current_dir(start_dir);
 
     let pkg_graph = metadata_cmd.build_graph()?;
@@ -89,6 +87,23 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
     let target_dir = workspace.target_directory().to_owned();
     let workspace_dir = workspace.root().to_owned();
 
+    let mut axoupdater_version = None;
+    for package in members.packages(DependencyDirection::Reverse) {
+        let mut version = None;
+        if package.name() == "axoupdater" {
+            version = Some(package.version());
+        } else {
+            for package in package.direct_links() {
+                if package.dep_name() == "axoupdater" {
+                    version = Some(package.to().version());
+                }
+            }
+        }
+        if let Some(version) = version {
+            axoupdater_version = Some(Version::Cargo(version.to_owned()));
+        }
+    }
+
     Ok(WorkspaceStructure {
         sub_workspaces: vec![],
         packages: all_package_info,
@@ -102,6 +117,7 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
             root_auto_includes,
             cargo_metadata_table,
             cargo_profiles,
+            axoupdater_version,
         },
     })
 }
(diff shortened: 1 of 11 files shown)


axoproject/src/rust.rs
```diff
@@ -89,6 +87,23 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
     let target_dir = workspace.target_directory().to_owned();
     let workspace_dir = workspace.root().to_owned();
 
+    let mut axoupdater_version = None;
```
Gankra, line 90: Ok so this is *probably* fine but strictly speaking it's not precise enough. In an ideal world we compute this value *per-package* (so construct a Map of Package -> Opt<Version>). That way if a package we don't care about "randomly" depends on axoupdater we don't freak out.

axoproject/src/rust.rs
```diff
@@ -89,6 +87,23 @@ fn workspace_info(pkg_graph: &PackageGraph) -> Result<WorkspaceStructure> {
...
+    for package in members.packages(DependencyDirection::Reverse) {
+        let mut version = None;
+        if package.name() == "axoupdater" {
+            version = Some(package.version());
+        } else {
+            for package in package.direct_links() {
+                if package.dep_name() == "axoupdater" {
+                    version = Some(package.to().version());
```
Gankra, line 98: This needs to be merging in any existing versions we've found. I was gonna say "min" but actually accumulating a Vec would probably produce better diagnostics.

cargo-dist/src/build/cargo.rs
```diff
@@ -20,11 +23,18 @@ impl<'a> DistGraphBuilder<'a> {
...
-            .workspace_dir
-            .clone();
+
+        let workspace = self.workspaces.workspace(workspace_idx);
+        if let Some(axoproject::Version::Cargo(version)) = &workspace.axoupdater_version {
+            let axoupdater_min_version = semver::Version::parse(AXOUPDATER_MINIMUM_VERSION)
+                .expect("invalid axoupdater const?!");
+
+            if *version < axoupdater_min_version {
+                warn!("Your project uses axoupdater as a library. We've detected that you're using {}, but this version of cargo-dist expects a version of at least {} (or greater). Please consider updating to ensure your users don't experience inconsistent update behaviour.", version, axoupdater_min_version);
```
Gankra, line 33: Needs to be a *hard error*, we *cannot* let them proceed or they're in hell.

Gankra (approved):
Approved but some changes to make

mistydemeo:
Pushed up a version with some changes.

I realized that since we were already iterating over direct packages in the workspace and not transitive dependencies (and since Gankra's suggestion was to ignore unexpected packages), we could do this a little easier by putting that dependency info on each package rather than on the workspace. Makes it easier to decide if we need to care about that axoupdater version per-package. We don't - currently - know of any projects that actually transitively depend on axoupdater, I suppose.

axoproject/src/rust.rs
```diff
@@ -200,6 +198,19 @@ fn package_info(_workspace_root: &Utf8Path, package: &PackageMetadata) -> Result
             )
         };
 
+    let axoupdater_version = if package.name() == "axoupdater" {
+        Some(Version::Cargo(package.version().to_owned()))
+    } else {
+        let mut version = None;
+        for subpackage in package.direct_links() {
```
ashleygwilliams, line 205: what would it mean if someone transitively depended on axoupdater (trying to sort if holding the transitive user of axoupdater to the dist version restrictions makes sense)

axoproject/src/tests.rs
```diff
@@ -5,6 +5,15 @@ use crate::{
     WorkspaceKind,
 };
 
+fn get_package(packages: &[(PackageIdx, &crate::PackageInfo)], name: &str) -> crate::PackageInfo {
+    packages
+        .iter()
+        .find(|(_, p)| p.name == name)
+        .unwrap()
```
ashleygwilliams, line 12: are we sure this unwrap is safe? we may want to change it to an expect at the very least

axoproject/src/tests.rs
```diff
@@ -15,7 +24,7 @@ fn test_self_detect() {
     assert_eq!(project.kind, WorkspaceKind::Rust);
     assert_eq!(packages.len(), 3);
 
-    let package = packages[0].1;
+    let package = packages[1].1;
```
ashleygwilliams, line 27: is there a way to more expressively write this (or comment why it's choosing things this way?)

axoproject/src/tests.rs
```diff
@@ -130,38 +139,38 @@ fn test_cargo_nonvirtual() {
...
-        let package = packages[0].1;
-        assert_eq!(package.name, "some-cdylib");
+        let package = get_package(&packages, "some-cdylib");
+        assert_eq!(package.name, "some-cdylib", "packages: {:?}", packages);
         assert!(package.binaries.is_empty());
     }
 
     {
-        let package = packages[1].1;
+        let package = get_package(&packages, "some-lib");
```
ashleygwilliams, line 148: love this, so much more expressive :)

axoproject/src/rust.rs
ashleygwilliams: at the very least, i think we should preserve the information that it's depended on transitively as that will be very useful for the end user receiving the error message

cargo-dist/src/build/cargo.rs
```diff
@@ -25,7 +31,23 @@ impl<'a> DistGraphBuilder<'a> {
...
+
         for (binary_idx, binary) in self.inner.binaries.iter().enumerate() {
+            let package = self.workspaces.package(binary.pkg_idx);
+
+            if let Some(axoproject::Version::Cargo(version)) = &package.axoupdater_version {
+                let axoupdater_min_version = semver::Version::parse(AXOUPDATER_MINIMUM_VERSION)
+                    .expect("invalid axoupdater const?!");
+
+                if *version < axoupdater_min_version {
+                    return Err(DistError::AxoupdaterTooOld {
```
ashleygwilliams, line 43: if we error on transitive axoupdater deps, we should make sure we tell them how it is in their dependency tree

cargo-dist/src/errors.rs
```diff
@@ -555,6 +555,20 @@ pub enum DistError {
     #[error("No bundle identifier was specified")]
     #[diagnostic(help("Please either enter a bundle identifier, or disable the Mac .pkg"))]
     MacPkgBundleIdentifierMissing {},
+
+    /// Project depends on a too-old axoupdater
+    #[error("Your project ({package_name}) uses axoupdater as a library, but the version specified ({your_version}) is older than the minimum supported version ({minimum}),")]
```
ashleygwilliams, line 560: it'd be great if we linked folks to a page in the docs that explained this a bit more

ashleygwilliams (commented):
some comments, mostly focused on: if we error on a transitive dep, we need to give the user more info. also makes a suggestion that we link to a docs page on this with the error.

axoproject/src/tests.rs
mistydemeo: I'll swap over to an `expect`. This is in the tests so we want it to crash if it's not there, but failing with more informative info will be more helpful.

cargo-dist/src/errors.rs
mistydemeo: Good call, I'll write a docs page for this and link to it.

axoproject/src/tests.rs
mistydemeo: When I spoke to Gankra, she mentioned her original goal with these tests was to test that package order was somewhat stable - hence accessing these with indices instead of by name. Since we've updated how we're fetching packages, it has shuffled the package order returned by axoproject so the original assumptions weren't relevant anymore.

cargo-dist/src/build/cargo.rs
mistydemeo: Updated:

```
  × Your project (cargo-dist) uses axoupdater as a library, but the version specified (0.7.2) is older than the
  │ minimum supported version (0.10.0). (The dependency comes via cargo-dist in the dependency tree.)
  help: https://opensource.axo.dev/cargo-dist/book/installers/updater.html#minimum-supported-updater-version-
        checking
```

mistydemeo:
Fixed a bug that caused us to report the wrong transitive dep version. Works fine for both direct and transitive deps now.

Direct:

```
  × Your project (cargo-dist) uses axoupdater as a library, but the version specified (0.7.2) is older than the
  │ minimum supported version (0.10.0). (The dependency comes via cargo-dist in the dependency tree.)
  help: https://opensource.axo.dev/cargo-dist/book/installers/updater.html#minimum-supported-updater-version-
        checking
```

Transitive:

```
  × Your project (deptest) uses axoupdater as a library, but the version specified (0.7.2) is older than the minimum
  │ supported version (0.10.0). (The dependency comes via cargo-dist in the dependency tree.)
  help: https://opensource.axo.dev/cargo-dist/book/installers/updater.html#minimum-supported-updater-version-
        checking
```

PR #1441, opened 2024-10-07
feat: disable single-library mode for dist=false

### LINKED ISSUE #1287: make dist=false make a package inelligible for single-package-mode (also don't error for "unknown" tags?)
I am using cargo-dist in a workspace and it would be good if I could push tags for some library crates without having cargo-dist create a release using the [singular library hack](https://opensource.axo.dev/cargo-dist/book/workspaces/workspace-guide.html?highlight=hack#singular-library-hack). Ideally, packages marked with `dist=false` should be really ignored, even when they are the only thing tagged.

Currently I have to manually remove these releases because otherwise [linking to the latest release](https://opensource.axo.dev/cargo-dist/book/reference/artifact-url.html?highlight=latest#linking-github-latest) breaks. For example, my workspace contains multiple libraries with independent versions and a single package with multiple binaries (this is the only one I want to actually have releases for).

mistydemeo:
In the past, we'd ignore `dist=false` when trying to enable single-package/library mode. This is, frankly, quite confusing and not desirable. Even if it's a slightly backwards incompatible change, we should make this do what it actually says.

Fixes #1287.

### DIFF
--- a/cargo-dist/src/announce.rs
+++ b/cargo-dist/src/announce.rs
@@ -545,12 +545,16 @@ fn select_packages(
     // add that package as a release still, on the assumption it's a Library
     if releases.is_empty() {
         if let ReleaseType::Package { idx, version: _ } = announcing.release {
-            releases.push(ReleaseArtifacts {
-                package_idx: PackageIdx(idx),
-                executables: vec![],
-                cdylibs: vec![],
-                cstaticlibs: vec![],
-            });
+            let config = graph.package_config(PackageIdx(idx));
+
+            if config.dist != Some(false) {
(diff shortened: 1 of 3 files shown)


ashleygwilliams (commented):
this generally looks good to me- could we add a test or two?

mistydemeo:
Good thought, I'll check and see how testable this seems.

ashleygwilliams (approved):

mistydemeo:
Test needs a little more work. This doesn't test the precise conditions that this PR addresses; that test passes on `main`.

mistydemeo:
Found a proper repro case; I'm readying a test case that would work here.

Issue #1453, opened 2024-10-15
Error: Couldn't generate main.wxs for <package>'s msi installer for workspace in a sub-folder


pnehrer:
We have a GitHub repository that has no `Cargo.toml` in the root, because the Rust workspace is in a sub-folder.

We do, however, have a `dist-workspace.toml` in the root, which includes the Rust workspace sub-folder in its `members` list.

When running `cargo dist generate --mode msi -v debug --check`, we get an error like:

```
  × Couldn't generate main.wxs for <package>'s msi installer
  ├─▶ The '/Users/pnehrer/git/repository/Cargo.toml' path does not exist
  ╰─▶ /Users/pnehrer/git/repository/Cargo.toml
```

It appears that the tool is looking for Cargo.toml in the repository root, and not in the Rust workspace sub-folder.

FWIW I have a small change that I can propose in a PR, which address this issue.

ashleygwilliams:
With the PR you submitted merged- I *think* this issue should be considered solved! If not, please reopen with more details :) Thanks so much again for contributing!! Out of curiosity we'd love to know what you are releasing with cargo-dist- feel free to join our discord https://discord.gg/RUC4EwG3

pnehrer:
Much appreciated @ashleygwilliams thank you!

pnehrer:
Unfortunately we didn't discover a related issue until now: https://github.com/axodotdev/cargo-dist/pull/1566

@ashleygwilliams should I create another issue?

PR #1454, opened 2024-10-15
Specify path to Cargo.toml when generating wix for package

pnehrer:
To avoid an error when generating a wix file for a package whose Cargo.toml is in a sub-folder, specify the full path to the package's Cargo.toml.

Fixes Issue #1453.


mistydemeo:
Good catch, thank you! We're prepping a release right now, but I'll review and test this afterwards.

mistydemeo (approved):
Gave it a test to confirm this doesn't negatively effect wix projects in the standard root path. Looking good - thanks for this!

pnehrer:
Outstanding, thank you! 🙂

PR #1466, opened 2024-10-18
chore: Cargo update

setarcos:
nix 0.26 can't compile on my platform (loongarch64), bump it to 0.29 will do the trick.


mistydemeo:
Thanks for reporting this - we haven't built on loongarch64, so I hadn't realized there was an issue. I've gone for a more targeted approach in #1471, which updates only the crates needed to bump nix. The next release should include this dependency update.

PR #1479, opened 2024-10-23
fix: tweak unified checksum format

mistydemeo:
shasum expects exactly two spaces between the hash and the filename. While sha256sum is tolerant of a single space, shasum will reject files that aren't in exactly its format.

Before:

```
$ shasum -a256 -c sha256.sum
shasum: sha256.sum: no properly formatted SHA checksum lines found
$ sha256sum --check sha256.sum
cargo-dist-npm-package.tar.gz: OK
source.tar.gz: OK
```

After:

```
$ shasum -a256 -c sha256.sum
cargo-dist-npm-package.tar.gz: OK
source.tar.gz: OK
$ sha256sum --check sha256.sum
cargo-dist-npm-package.tar.gz: OK
source.tar.gz: OK
```

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -433,7 +433,11 @@ fn generate_unified_checksum(
                 continue;
             }
 
-            writeln!(&mut output, "{checksum} {artifact_name}").unwrap();
+            // We write with exactly two spaces to match the output of
+            // shasum. While sha256sum is tolerant of varying numbers of
+            // spaces, shasum itself will only match precisely two -
+            // we need to be precise here for compatibility.
+            writeln!(&mut output, "{checksum}  {artifact_name}").unwrap();
         }
     }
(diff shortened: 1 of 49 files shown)


cargo-dist/src/lib.rs
```diff
@@ -433,7 +433,7 @@ fn generate_unified_checksum(
                 continue;
             }
 
-            writeln!(&mut output, "{checksum} {artifact_name}").unwrap();
+            writeln!(&mut output, "{checksum}  {artifact_name}").unwrap();
```
ashleygwilliams, line 436: it would be useful to put a comment here explaining that there are purposefully 2 spaces here and why- since it's hard to know from just this line :)

mistydemeo: Good call!

fasterthanlime (approved):
Nice catch!

PR #1480, opened 2024-10-23
tests: include real filenames for unified checksums

mistydemeo:
I was looking at the snapshots and realizing we should probably be able to include the real filenames while keeping the snapshots stable. How do you feel, @fasterthanlime?


fasterthanlime (approved):
Looks good to me! I didn’t realize capture groups were supported!

(Also I figured it’d be better to make those stable first)

PR #1485, opened 2024-10-24
docs: replace cargo-dist with dist

mistydemeo:
This handles the documentation side of the rename.

~~Note that we need to review all links to `cli.md` after merging #1267 - that file is auto-generated, so any links to sections of that document will change after that PR is merged. Those changes will need to get added to this PR at that point.~~ Done.

Refs #1267.


book/src/install.md
```diff
@@ -34,7 +36,7 @@ brew install axodotdev/tap/cargo-dist
 
 #### Pacman (Arch Linux)
 
-Arch Linux users can install `cargo-dist` from the [extra repository](https://archlinux.org/packages/extra/x86_64/cargo-dist/) using [pacman](https://wiki.archlinux.org/title/Pacman):
```
ashleygwilliams, line 37: we may need to leave this one as the arch package is not renamed

book/src/installers/other.md
```diff
@@ -1,13 +1,13 @@
 # Other Installation Methods
 
-cargo-dist projects can also theoretically be installed with the following, through no active effort of our own:
+dist projects can also theoretically be installed with the following, through no active effort of our own:
 
 * [cargo-install][] (just [cargo publish][] like normal)
```
ashleygwilliams, line 5: this is probably a more general note but talking about cargo install here doesn't make much sense anymore

book/src/installers/other.md
```diff
@@ -1,13 +1,13 @@
...
 * [cargo-binstall][] (the URL schema we use for Github Releases is auto-detected)
 
 In the future we might [support displaying these kinds of install methods][issue-info-install].
 
-Note that cargo-install is just building from the uploaded source with the --release profile, and so if you're relying on cargo-dist or unpublished files for some key behaviours, this may cause problems. [It also disrespects your lockfile unless you pass --locked][install-locked]. You can more closely approximate cargo-dist's build with:
+Note that cargo-install is just building from the uploaded source with the --release profile, and so if you're relying on dist or unpublished files for some key behaviours, this may cause problems. [It also disrespects your lockfile unless you pass --locked][install-locked]. You can more closely approximate dist's build with:
```
ashleygwilliams, line 10: same here

book/src/installers/updater.md
```diff
@@ -6,7 +6,7 @@ NOTE: This feature is currently experimental.
...
 
-If you add `install-updater = true` to your `Cargo.toml`, cargo-dist's shell and PowerShell installers will include an updater program alongside your program itself. This standalone program will be installed as the name `yourpackage-update`, and users can simply run it to poll for new releases and have them installed. The source code for this program is open source in the [axoupdater] repository.
+If you add `install-updater = true` to your `Cargo.toml`, dist's shell and PowerShell installers will include an updater program alongside your program itself. This standalone program will be installed as the name `yourpackage-update`, and users can simply run it to poll for new releases and have them installed. The source code for this program is open source in the [axoupdater] repository.
```
ashleygwilliams, line 9: this needs a config change

book/src/quickstart/everyone-else.md
```diff
@@ -26,48 +26,48 @@ license = "GPL-3.0-only"
...
-cargo dist init --yes
+# setup dist in your project (--yes to accept defaults)
+dist init --yes
 git add .
-git commit -am 'chore: wow shiny new cargo-dist CI!'
+git commit -am 'chore: wow shiny new dist CI!'
 ```
 
-**It's very common for `cargo dist init` to return an error about the "repository" URLs set in your Cargo.toml. If this happens, no work will be lost.** You can just follow the instructions in the error and rerun `cargo dist init` again and it will pick up where you left off.**
+**It's very common for `dist init` to return an error about the "repository" URLs set in your Cargo.toml. If this happens, no work will be lost.** You can just follow the instructions in the error and rerun `dist init` again and it will pick up where you left off.**
```
ashleygwilliams, line 48: def shouldn't mention cargo.toml here

ashleygwilliams (approved):
realized almost all my comments are not about this particular change so we should ship this but then we do really need a docs pass (yay that i'm already doing one)

book/src/install.md
mistydemeo: Oh, good call. I intended this to be "you can install `dist` (the software) using the command bellow (with a different name)" but I realize there's a lot of ambiguity in how this is written.

book/src/installers/other.md
mistydemeo: We can probably get rid of this whole section? It doesn't seem especially relevant even to our Rust users.

PR #1488, opened 2024-10-24
Improve target understanding

fasterthanlime:
This adds a dependency on https://docs.rs/target-lexicon and explores which places we could start using it (for cross-compilation, definitely)


fasterthanlime:
Closing in favor of #1470

PR #1497, opened 2024-10-28
Verify checksums in `install.sh`

fasterthanlime:
This is a WIP I'm pushing before stepping away from the computer — it includes making `ArtifactId` strongly-typed, which _could_ be extracted into its own PR (but I'd rather spend my time elsewhere).

Only the shell installer is doing checks for now, and I haven't tested all the codepaths (Or even a single one, as I'm pushing this) — this is just the result of research.

I have to say, most of the time was spent tracking down why/where PlatformSupport was supposed to magically gain checksums (spoiler: it wasn't! anywhere!).

I think we can tackle the powershell installer in a separate PR, probably. I have a minimal testing plan for this, but not a comprehensive one.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -160,19 +160,25 @@ download_binary_and_run_installer() {
     assert_nz "$_true_arch" "arch"
     local _cur_arch="$_true_arch"
 
-    # Lookup what archives support this platform
+
+    # look up what archives support this platform
     local _artifact_name
     _artifact_name="$(select_archive_for_arch "$_true_arch")" || return 1
     local _bins
     local _zip_ext
     local _arch
+    local _checksum_style
+    local _checksum_value
 
-    # try each archive, checking runtime conditions like libc versions
-    # accepting the first one that matches, as it's the best match
+    # destructure selected archive info into locals
     case "$_artifact_name" in {% for archive in platform_support.archives %}
         "{{ archive.id }}")
             _arch="{{ archive.target_triple }}"
             _zip_ext="{{ archive.zip_style }}"
+            {%- if archive.checksum %}
+            _checksum_style="{{ archive.checksum.style }}"
+            _checksum_value="{{ archive.checksum.value }}"
+            {%- endif %}
             _bins="{% for bin in archive.executables %}{{ bin }}{{ " " if not loop.last else "" }}{% endfor %}"
             _bins_js_array='{% for bin in archive.executables %}"{{ bin }}"{{ "," if not loop.last else ""}}{% endfor %}'
             {%- if "cdylib" in install_libraries %}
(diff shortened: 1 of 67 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -146,19 +146,25 @@ download_binary_and_run_installer() {
...
+    local _checksum_value
 
-    # try each archive, checking runtime conditions like libc versions
-    # accepting the first one that matches, as it's the best match
+    # destructure selected archive info into locals
     case "$_artifact_name" in {% for archive in platform_support.archives %}
         "{{ archive.id }}")
             _arch="{{ archive.target_triple }}"
             _zip_ext="{{ archive.zip_style }}"
+            {%- if archive.checksum %}
```
mistydemeo, line 164: Huge gotcha here: `archive.checksum` is an index to an artifact struct, not the struct itself. Because jinja2 was originally designed for Python, `0` is falsy - and so is `Some(0)`. This check has the chance to return false for certain checksums. See `{%- if archive.updater != None %}` below in this same case statement for another place we had to work around this.

We also need to make sure `_checksum_style` and `_checksum_value` get initialized to somethign since we do `set -u` at the top of the file - `local _checksum_style` isn't enough to define it, so referencing it later will error out. (You'll see that happen in CI on this PR.)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -214,6 +220,8 @@ download_binary_and_run_installer() {
       exit 1
     fi
 
+    verify_checksum "$_file" "$_checksum_style" "$_checksum_value"
```
mistydemeo, line 223: Alternately, you can provide a default here, eg `"${_checksum_style:-}"`, so referencing it when undefined isn't fatal.

cargo-dist/tests/gallery/dist/snapshot.rs
```diff
@@ -169,7 +169,7 @@ pub fn snapshot_settings_with_gallery_filter() -> insta::Settings {
         r#""build_environment": "indeterminate""#,
     );
     settings.add_filter(
-        r"[0-9a-f]{64}  (?<filename>([a-zA-Z0-9-_]+)(\.tar\.gz|\.pkg))",
+        r"[0-9a-f]{64} .(?<filename>([a-zA-Z0-9-_]+)(\.tar\.gz|\.pkg))",
```
fasterthanlime, line 172: needed because we changed to actually emitting ` *` (space star)  instead of `  ` (space space)

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -1167,4 +1182,102 @@ downloader() {
...
+    esac
+
+    if [ "$_calculated_checksum" != "$_checksum_value" ]; then
+        err "checksum mismatch
+            want: $_checksum_value
+            got:  $_calculated_checksum"
+    fi
+}
+
+verify_sha256sum() {
```
mistydemeo, line 1262: Looks like this version may be currently unused?

PR #1499, opened 2024-10-29
Explore 'strict undefined behavior' (for minjinja2)

fasterthanlime:
Most likely we need to contribute back to minijinja2 to get proper tracking information for "undefined errors" because right now all we see is this:

CleanShot 2024-10-29 at 14 02 49@2x


fasterthanlime:
Current state of this PR: the corresponding minjinja fix has landed in their `main` branch, but not in a point release yet, which means our dep now looks like this:

CleanShot 2024-11-01 at 16 04 50@2x

So I don't love that for us and this PR can probably wait for a new minjinja to land (releases seem regular)

---

BUT! it looks awesome and is very useful: I pulled in color-backtrace for coloring and filtering and this helps me at least (I'm curious to see what @mistydemeo and @duckinator think about it specifically) narrow down what's wrong almost immediately:

CleanShot 2024-11-01 at 16 08 54@2x

color-backtrace is _awesome_ (and the big dependency here is `backtrace`, which we already have) and has a `RUST_LIB_BACKTRACE=full` mode that even shows you the Rust source code:

CleanShot 2024-11-01 at 16 09 48@2x

Even though I'd like to wait for a minjinja point release to drop, I'm gonna stack my other PRs on top of this one because, damn, that's good tooling.

fasterthanlime:
Note 1: I've _forced_ backtrace capture, even if you haven't otherwise exported `RUST_LIB_BACKTRACE=1` because I consider this equivalent to a panic, it might happen in CI, we're telling our users to report it and when they do, we'll want as much info as possible.

---

Note 2: this PR also fixes the few instances of us violating "strict undefined behavior", minjinja is going to get a little legalistic on us — the only thing y'all need to know is that if it complains something like:

```
      undefined value (in installer/npm/run.js:4)
   ╭─[4:1]
 3 │ const { run } = require("./binary");
 4 │ {%- if bin %}
   · ─────────────
 5 │ run({{ bin }});
   ╰────
```

Then you need `if bin is defined`, and the reason this happens is usually because the field is defined something like this:

```rust
struct Blah {
    pub bin: Option<String>,
}
```

But in the _actual_ case of `bin`? it's because it's rendered without any "context/globals" object once, as evidenced later in the code:

https://github.com/axodotdev/cargo-dist/blob/33316c58184a936ab293cc5567efc1fadba10453/cargo-dist/src/backend/installer/npm.rs#L89-L93

ie. while "rendering the directory template", it gets rendered once with zero context, and then it's removed and re-rendered with the correct context. That's a hack and I hate it, but with just `if bin is defined` instead of `if bin`, we can go ahead anyway:

```jinja2
{%- if bin is defined %}
run({{ bin }});
{%- endif %}
```

fasterthanlime:
After discussing in Discord, we'll publish `minjinja` as `axominijinja` and switch to that for the next release, then switch back when it's done.

fasterthanlime:
Taking this out of draft since it's ready for being merged imho!

fasterthanlime:
CleanShot 2024-11-01 at 17 16 27@2x

I caught another template issue (a benign one) while working on https://github.com/axodotdev/cargo-dist/pull/1509 — which mostly tells me the homebrew stuff isn't _all_ covered by unit tests

cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -1,70 +1,32 @@
...
+
+  {% macro fragment_url_and_hash(cond, frag) -%}
+  {%- if frag %}
+  if {{ cond }}
+    url "{{ inner.base_url }}/{{ frag.id }}"
+    {%- if frag.sha256 is defined %}
+    sha256 "{{ frag.sha256 }}"
     {%- endif %}
-  end
+  endif
```
mistydemeo, line 15: Typo? This is in the Ruby code, not jinja syntax - Ruby uses `if/end`, not `if/endif`.
```suggestion
  end
```

cargo-dist/templates/installer/homebrew.rb.j2
```diff
@@ -1,70 +1,32 @@
...
-    {%- endif %}
-    {%- if x86_64_linux.id %}
-    if Hardware::CPU.intel?
-      url "{{ inner.base_url }}/{{ x86_64_linux.id }}"
-      {%- if x86_64_linux_sha256 %}
-      sha256 "{{ x86_64_linux_sha256 }}"
-      {%- endif %}
-    end
+
+  {% macro fragment_url_and_hash(cond, frag) -%}
```
mistydemeo, line 8: A comment to note this is a jinja macro could be helpful - we don't use a lot of jinja macros (and the `macro` keyword is a little overloaded given we're writing Rust). It'll help make it clearer to the next person looking at this.

cargo-dist/templates/installer/npm/run.js.j2
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
 
 const { run } = require("./binary");
+{%- if bin is defined %}
 run({{ bin }});
```
mistydemeo, line 5: Did we run into a scenario where `bin` is undefined? I think that would actually be a for-real error here, not something for us to work around.

cargo-dist/tests/snapshots/akaikatana_basic.snap
```diff
@@ -1220,20 +1220,26 @@ class AkaikatanaRepack < Formula
...
-    if Hardware::CPU.intel?
-      url "https://github.com/mistydemeo/akaikatana-repack/releases/download/v0.2.0/akaikatana-repack-x86_64-apple-darwin.tar.xz"
-    end
-  end
-  if OS.linux?
-    if Hardware::CPU.intel?
-      url "https://github.com/mistydemeo/akaikatana-repack/releases/download/v0.2.0/akaikatana-repack-x86_64-unknown-linux-gnu.tar.xz"
-    end
-  end
+
```
mistydemeo, line 1223: We need to be a little more careful about adding too much extra whitespace here - let's try to trim that back.

cargo-dist/tests/snapshots/akaikatana_basic.snap
```diff
@@ -1220,20 +1220,26 @@ class AkaikatanaRepack < Formula
...
+  endif
+  
+  
+  if OS.linux? && Hardware::CPU.intel?
+    url "https://github.com/mistydemeo/akaikatana-repack/releases/download/v0.2.0/akaikatana-repack-x86_64-unknown-linux-gnu.tar.xz"
+    sha256 "CENSORED"
+  endif
+  
   license "GPL-2.0-or-later"
+
```
mistydemeo, line 1242: Ditto.

cargo-dist/tests/snapshots/akaikatana_basic.snap
```diff
@@ -1258,16 +1264,17 @@ class AkaikatanaRepack < Formula
   end
 
   def install
+
```
mistydemeo, line 1267: Here too and elsewhere in the `install` method.

mistydemeo (changes requested):
Looking good, but a few comments - one error, and a few other suggestions and questions.

cargo-dist/templates/installer/npm/run.js.j2
fasterthanlime: The code is doing something dirty here — it's rendering the whole `npm/` folder passing `NpmInstallerInfo` (which does _not_ have a `bin` property) — and then removing the generated `run.js` (which presumably has a null or whatever), and then re-generating from `run.js.j2` for every binary, see `fn mangle_run_js()` in `src/backend/installer/npm.rs`

fasterthanlime: (What this PR does is not the proper fix, but I was, once upon a time, trying to keep this PR small)

cargo-dist/tests/snapshots/axolotlsay_basic.snap
```diff
@@ -1,7 +1,6 @@
 ---
 source: cargo-dist/tests/gallery/dist/snapshot.rs
 expression: self.payload
-snapshot_kind: text
```
fasterthanlime, line 4: probably due to the minijinja update? idk

mistydemeo: What version of cargo-insta do you have? This was added when Ellen updated a test, which makes me think they might have a newer version than you or I do.

mistydemeo: OK yeah, this field landed in cargo-insta 1.41 https://github.com/mitsuhiko/insta/commit/0b81773e97535f1960c1443f85b491adceb36dd6

Prior to that insta only had a single snapshot format so there was no need to specify.

fasterthanlime:
@mistydemeo I addressed whitespace issues — in fact, I made sure the template produces _exactly_ the same input as before, just using those nice macros and the new "template input" structs.

I've also added a comment about the jinjaness of the macro (and a link to the surprising-to-find docs), and fixed the endif/end mishap.

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -56,64 +40,87 @@ pub struct HomebrewInstallerInfo {
...
-/// Grab the sha256 checksum for this artifact from the manifest
-fn use_sha256_checksum(
-    manifest: &DistManifest,
-    fragment: &Option<ExecutableZipFragment>,
-    checksum: &mut Option<String>,
-) {
     let checksum_key = ChecksumStyle::Sha256.ext();
-    if let Some(frag) = &fragment {
-        *checksum = manifest
+    let map_fragment = |fragment: ExecutableZipFragment| -> HomebrewFragment {
```
mistydemeo, line 65: Can we move this definition right above the only place we use it?

cargo-dist/src/backend/installer/homebrew.rs
```diff
@@ -56,64 +40,87 @@ pub struct HomebrewInstallerInfo {
...
+    #[serde(flatten)]
+    fragments: HomebrewFragments<HomebrewFragment>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+struct HomebrewFragment {
+    #[serde(flatten)]
+    fragment: ExecutableZipFragment,
+
+    /// SHA256 sum of the fragment — if we have it, which we might not, in tests at least?
```
mistydemeo, line 119: The local builds usually happen on a different host from the global build, so having these depends on the sha256s getting piped through all the way from the local build manifests to the info we have at the global step. We do that in CI, but it's not guaranteed that someone doing their own kinda builds outside this environment would have them.

(Same deal if you're just generating the installers via `dist build -aglobal` without having actually built any local artifacts at all.)

cargo-dist/templates/installer/npm/run.js.j2
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
 
 const { run } = require("./binary");
+{%- if bin is defined %}
```
mistydemeo, line 4: I don't mind us deferring this to later but we should add a comment so we understand why we did this.

mistydemeo (approved):
Looking good, just a few comments now!

cargo-dist/tests/snapshots/axolotlsay_basic.snap
fasterthanlime: I’ll upgrade locally!

cargo-dist/templates/installer/npm/run.js.j2
fasterthanlime: https://github.com/axodotdev/cargo-dist/issues/1524

cargo-dist/src/backend/installer/homebrew.rs
fasterthanlime: In my mind, it already is — putting it inside the `macro_rules!` block simply makes the macro bigger and harder to review.

fasterthanlime:
The mac tests are failing because it's trying to install a linuxbrew-only package..

PR #1501, opened 2024-10-29
Distinguish "cargo metadata failed" from "cargo is not installed"

### LINKED ISSUE #1500: "cargo metadata" returned non-zero exit code != "cargo" is not installed
As of e7ed7b5e34e6ea9b7754832ed017f4deb1a84275 at least, that error conflates both:

```
Run dist host --steps=check --output-format=json > plan-dist-manifest.json
  × We encountered an issue trying to read your workspace
  ╰─▶   × Your app has a Cargo.toml, but you don't appear to have cargo installed.
        help: Is cargo in your PATH? You can install cargo via: https://rustup.rs/
```

When really what happened was something like this:

```
hi on  main [!] via 🦀 v1.82.0
❯ cargo metadata
error: expected `.`, `=`
 --> Cargo.toml:1:8
  |
1 | package]
  |        ^
  |

hi on  main [!] via 🦀 v1.82.0
❯ echo $?
101
```

On a call with @ashleygwilliams, we agreed I'd contribute a quick fix, the plan is to:

  * Run the command as usual
  * If it failed, check if we think `cargo` is in path (we can use [which](https://lib.rs/crates/which))
    * If it's not, go for the "cargo needs to be installed" error
    * If it is installed, show cargo metadata stdout/stderr to explain why it failed

The unintuitive ordering here is just in case `which` ends up being stricter than the system and it doesn't "see" cargo in PATH when actually just _trying to execute cargo_ would work.

fasterthanlime:
Closes #1500

### DIFF
--- a/axoproject/src/rust.rs
+++ b/axoproject/src/rust.rs
@@ -43,8 +43,17 @@ pub fn get_workspace(start_dir: &Utf8Path, clamp_to_dir: Option<&Utf8Path>) -> W
             let error = match e {
                 // Indicates we failed to run `cargo metadata`; this is the
                 // one we want to intercept and replace with a friendlier error.
-                crate::AxoprojectError::CargoMetadata(guppy::Error::CommandError(_)) => {
-                    crate::AxoprojectError::CargoMissing {}
+                crate::AxoprojectError::CargoMetadata(e) => {
+                    if cargo_version_works() {
+                        // we have cargo, `cargo metadata` just failed though — relay
+                        // its stderr, which should clue in the users on TOML parse errors,
+                        // invalid dependencies, etc.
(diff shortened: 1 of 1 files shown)


fasterthanlime:
in action:

CleanShot 2024-10-29 at 16 31 45@2x

fasterthanlime:
(note: this was a regression introduced recently, which was trying (and succeeding, tbh!) in improving this error message:

CleanShot 2024-10-29 at 16 32 52@2x

)

duckinator (approved):
Works locally and the implementation is about as simple as it gets. Looks good to me. :+1:

mistydemeo (approved):
Looks good, seems like a good heuristic!

PR #1503, opened 2024-10-29
feat: allow installer domain to be overridden

### LINKED ISSUE #1495: Provide single environment variable for configuring GitHub installation and update URLs
We currently provide one environment variable to configure the installation URL for the shell and PowerShell installers: `INSTALLER_DOWNLOAD_URL`. This variable is release-specific, and contains not just the domain but also the path where the installers for *a specific version* should be found. This makes it less flexible than it could be; it's not usable as a "set and forget" URL for users who need to use mirrors for fetching artifacts. We'd like to be able to configure a base URL instead, that overrides just that `https://github.com` portion of the GitHub URL while having the installer provide the rest of the path.

We also need an environment variable suitable for use with the updater. The updater needs to access either the GitHub or axo Releases API, and some users may need to specify a proxy or mirror for those servers as well.

We had some initial discussions about reusing `INSTALLER_DOWNLOAD_URL` for this: we would take just the base URL and use that as the GitHub API root. This left us with an interesting ambiguity, however: the relationship between the web and API URLs isn't fully stable. For GitHub dotcom, web is at `github.com` and the API is at `api.github.com`. For GitHub Enterprise, however, the API is at `domain.com/api` instead of a subdomain. We expect that some users using this feature to mirror will be mirroring releases from GitHub dotcom to Enterprise, and so we need to make sure that whatever option we pick here doesn't lock one set of users out from using this feature.

While we could define separate URLs for the installer URL and the GitHub API, that's unnecessarily clunky and some users understandably won't want to do that. Instead, one option would be to define two environment variables for the two different styles of *hosting*: one for GitHub dotcom style URLs, one for GitHub Enterprise style URLs.

* `INSTALLER_GITHUB_BASE_URL` (or similar) - replaces the `github.com` part of the installer URL, looks for the API at the `api.` subdomain
* `INSTALLER_GITHUB_ENTERPRISE_BASE_URL` (or similar) - replaces the `github.com` part of the installer URL, looks for the API at the `/api` path at that domain

Since axo Releases has its own URL structures, we should focus on just overriding GitHub at this time and provide different URLs for overriding axo.

mistydemeo:
Previously, we allowed overriding the full URL to the installer path - that is, everything prior to the actual installer filename. This isn't flexible enough for people who are mirroring or proxying github.com, and who want to be able to set a single environment variable that can work for any version of a program.

This adds two environment variables, `INSTALLER_GITHUB_BASE_URL` and `INSTALLER_GITHUB_ENTERPRISE_BASE_URL`, which allow overriding just the domain. They do the same thing for these installers, but will have slightly different behaviour for axoupdater.

Implementation wise, this updates our hosting types a little bit. axoproject's `GithubRepo`[^1] hardcodes `github.com` slightly less - instead of being inlined in a bunch of strings, it provides a `domain()` method that itself returns a hardcoded string, which will give us some room to eventually stop hardcoding it without changing a bunch of occurrences of `github.com`. In addition to our methods that return fully-qualified URLs, I've added a method that allows retrieving just the path component of a URL with the domain omitted so we can do the domain substitution in the installers. I've also had us pipe the hosting data right through to the installers so that the full hosting data, and not just the computed URL, can be accessed.

Fixes #1495.

[^1]: Trademark-noncompliant struct name

### DIFF
--- a/axoproject/src/repo.rs
+++ b/axoproject/src/repo.rs
@@ -20,9 +20,20 @@ pub struct GithubRepo {
 }
 
 impl GithubRepo {
+    /// Returns the domain. At the moment this is hardcoded to github.com, but
+    /// it may support more options in the future.
+    pub fn domain(&self) -> String {
+        "https://github.com".to_owned()
+    }
+
+    /// Similar to `web_url`, but returns just the path component without the domain.
+    pub fn web_path(&self) -> String {
+        format!("/{}/{}", self.owner, self.name)
+    }
+
     /// Returns a URL suitable for web access to the repository.
     pub fn web_url(&self) -> String {
-        format!("https://github.com/{}/{}", self.owner, self.name)
+        format!("{}{}", self.domain(), self.web_path())
     }
 
     /// Constructs a new Github repository from a "owner/name" string. Notably, this does not check
--- a/axoproject/src/repo.rs
+++ b/axoproject/src/repo.rs
@@ -26,7 +26,7 @@ impl GithubRepo {
         "https://github.com".to_owned()
     }
 
-    /// Similar to `web_url`, but returns just the path component without the domain.
+    /// Path component. Used with `domain` to construct `web_url`.
     pub fn web_path(&self) -> String {
         format!("/{}/{}", self.owner, self.name)
     }
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -354,7 +354,7 @@ pub struct SystemInfo {
 }
 
 /// Release-specific environment variables
-#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct EnvironmentVariables {
     /// Environment variable to force an install location
     pub install_dir_env_var: String,
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -772,7 +772,14 @@ pub struct Hosting {
 /// Github Hosting
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct GithubHosting {
-    /// The URL of the Github Release's artifact downloads
+    /// The URL of the host for GitHub, usually `"https://github.com"`
+    /// (This can vary for GitHub Enterprise)
+    pub artifact_base_url: String,
+    /// The path of the release without the base URL
+    ///
+    /// e.g. `/myowner/myrepo/releases/download/v1.0.0/`
+    pub artifact_download_path: String,
+    /// The URL of the GitHub Release's artifact downloads
     ///
     /// e.g. `https://github.com/myowner/myrepo/releases/download/v1.0.0/`
     pub artifact_download_url: String,
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -758,7 +758,7 @@ impl DistManifest {
 
 impl Release {
     /// Get the base URL that artifacts should be downloaded from (append the artifact name to the URL)
-    pub fn artifact_download_url(&self) -> Option<&str> {
+    pub fn artifact_download_url(&self) -> Option<String> {
         self.hosting.artifact_download_url()
     }
 }
@@ -787,10 +787,6 @@ pub struct GithubHosting {
     ///
     /// e.g. `/myowner/myrepo/releases/download/v1.0.0/`
     pub artifact_download_path: String,
-    /// The URL of the GitHub Release's artifact downloads
-    ///
-    /// e.g. `https://github.com/myowner/myrepo/releases/download/v1.0.0/`
-    pub artifact_download_url: String,
     /// The owner of the repo
     pub owner: String,
     /// The name of the repo
@@ -799,14 +795,17 @@ pub struct GithubHosting {
 
 impl Hosting {
     /// Get the base URL that artifacts should be downloaded from (append the artifact name to the URL)
-    pub fn artifact_download_url(&self) -> Option<&str> {
+    pub fn artifact_download_url(&self) -> Option<String> {
         let Hosting { axodotdev, github } = &self;
         // Prefer axodotdev is present, otherwise github
         if let Some(host) = &axodotdev {
-            return host.set_download_url.as_deref();
+            return host.set_download_url.clone();
         }
         if let Some(host) = &github {
-            return Some(&host.artifact_download_url);
+            return Some(format!(
+                "{}{}",
+                host.artifact_base_url, host.artifact_download_path
+            ));
         }
         None
     }
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -367,7 +367,7 @@ pub struct EnvironmentVariables {
     /// Environment variable to set the GitHub base URL
     pub github_base_url_env_var: String,
     /// Environment variable to set the GitHub Enterprise base URL
-    pub github_enterprise_base_url_env_var: String,
+    pub ghe_base_url_env_var: String,
 }
 
 /// A Release of an Application
@@ -706,8 +706,7 @@ impl DistManifest {
             let disable_update_env_var = format!("{env_app_name}_DISABLE_UPDATE");
             let no_modify_path_env_var = format!("{env_app_name}_NO_MODIFY_PATH");
             let github_base_url_env_var = format!("{env_app_name}_INSTALLER_GITHUB_BASE_URL");
-            let github_enterprise_base_url_env_var =
-                format!("{env_app_name}_INSTALLER_GITHUB_ENTERPRISE_BASE_URL");
+            let ghe_base_url_env_var = format!("{env_app_name}_INSTALLER_GHE_BASE_URL");
 
             let environment_variables = EnvironmentVariables {
                 install_dir_env_var,
@@ -715,7 +714,7 @@ impl DistManifest {
                 disable_update_env_var,
                 no_modify_path_env_var,
                 github_base_url_env_var,
-                github_enterprise_base_url_env_var,
+                ghe_base_url_env_var,
             };
 
             self.releases.push(Release {
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -364,6 +364,10 @@ pub struct EnvironmentVariables {
     pub disable_update_env_var: String,
     /// Environment variable to disable modifying the path
     pub no_modify_path_env_var: String,
+    /// Environment variable to set the GitHub base URL
+    pub github_base_url_env_var: String,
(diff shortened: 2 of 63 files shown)


mistydemeo:
Set up a stub HTTP server locally and then set the three environment variables independently of each other to watch what it does. For shell, all three did the right thing:

```
$ set -x INSTALLER_DOWNLOAD_URL http://localhost:8000/blah/blah
$ curl --proto '=https' --tlsv1.2 -LsSf https://github.com/mistydemeo/cargodisttest/releases/download/v0.2.263/axolotlsay-installer.sh | sh
# requests http://localhost:8000/blah/blah/axolotlsay-aarch64-apple-darwin.tar.xz
$ set -e INSTALLER_DOWNLOAD_URL
$ set -x INSTALLER_GITHUB_ENTERPRISE_BASE_URL http://localhost:8000
$ curl --proto '=https' --tlsv1.2 -LsSf https://github.com/mistydemeo/cargodisttest/releases/download/v0.2.263/axolotlsay-installer.sh | sh
# requests http://localhost:8000/mistydemeo/cargodisttest/releases/download/v0.2.263/axolotlsay-aarch64-apple-darwin.tar.xz
$ set -e INSTALLER_GITHUB_ENTERPRISE_BASE_URL
$ set -x INSTALLER_GITHUB_BASE_URL http://localhost:8000
$ curl --proto '=https' --tlsv1.2 -LsSf https://github.com/mistydemeo/cargodisttest/releases/download/v0.2.263/axolotlsay-installer.sh | sh
# requests http://localhost:8000/mistydemeo/cargodisttest/releases/download/v0.2.263/axolotlsay-aarch64-apple-darwin.tar.xz
```

Will test powershell.

mistydemeo:
Confirmed working on powershell.

cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
```diff
@@ -722,13 +732,23 @@ expression: json_schema
...
+          "description": "The URL of the host for GitHub, usually `\"https://github.com\"` (This can vary for GitHub Enterprise)",
+          "type": "string"
+        },
+        "artifact_download_path": {
+          "description": "The path of the release without the base URL\n\ne.g. `/myowner/myrepo/releases/download/v1.0.0/`",
+          "type": "string"
+        },
         "artifact_download_url": {
-          "description": "The URL of the Github Release's artifact downloads\n\ne.g. `https://github.com/myowner/myrepo/releases/download/v1.0.0/`",
+          "description": "The URL of the GitHub Release's artifact downloads\n\ne.g. `https://github.com/myowner/myrepo/releases/download/v1.0.0/`",
```
fasterthanlime, line 751: ™️

cargo-dist/src/tasks.rs
```diff
@@ -1976,10 +1979,7 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
                     runtime_conditions,
                     platform_support: None,
                     // Not actually needed for this installer type
-                    install_dir_env_var: String::new(),
-                    unmanaged_dir_env_var: String::new(),
-                    disable_update_env_var: String::new(),
-                    no_modify_path_env_var: String::new(),
+                    env_vars: cargo_dist_schema::EnvironmentVariables::default(),
```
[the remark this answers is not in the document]
fasterthanlime: Okay now that I got the nit out the way, I'm actually thinking we should clean up what we pass to installers... the `PlatformSupport` situation is a mess and bit me in #1497.

"Some things are sometimes empty strings" is bad, "some things are sometimes None" is slightly less bad, but ideally we'd get closer to a situation where "if you're holding a value of type HomebrewInstallerInputs, then we know all the fields are set and meaningful".

In that regard, I have Negative Feelings about `EnvironmentVariables` implementing `Default` (which sets everything to empty strings).

I think I would rather `EnvironmentVariables` _not_ implement `Default`, and to change the type of `env_vars` to be of type `Option<EnvironmentVariables>` — that way we have all of them, or none at all.

fasterthanlime (approved):
I'm leaving this as an approve because I don't want to block it shipping — but I have suggested one specific change that I'd like to see happen (don't derive Default for `EnvironmentVariables`, make the `env_vars` field optional, deal with it being None/Some when you actually need it, )

cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
```diff
@@ -660,6 +660,8 @@ expression: json_schema
       "type": "object",
       "required": [
         "disable_update_env_var",
+        "github_base_url_env_var",
+        "github_enterprise_base_url_env_var",
```
fasterthanlime, line 664: @mistydemeo do we also want to rename that to be `ghe` styled instead of `github_enterprise` or?

mistydemeo: I was about to say this is purely internal, but it does show up in the manifest where our end users (packagers) will see it. Yeah okay, it should probably be consistent in that case, huh.

(The `GHE` renaming was because `GITHUB_ENTERPRISE` was deemed too long and unwieldy for our grandusers.)

fasterthanlime:
PowerShell works for me too:

CleanShot 2024-10-30 at 21 35 08@2x

cargo-dist/src/tasks.rs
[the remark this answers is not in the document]
mistydemeo: That seems reasonable, for sure - that does sound a lot cleaner. Given our possible schedule around this, I might defer that to a later PR.

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -50,26 +50,45 @@ param (
...
+  $installer_base_url = $env:{{ env_vars.github_enterprise_base_url_env_var }}
+} elseif ($env:{{ env_vars.github_base_url_env_var }}) {
+  $installer_base_url = $env:{{ env_vars.github_base_url_env_var }}
+} else {
+  $installer_base_url = "{{ hosting.github.artifact_base_url }}"
+}
+if ($env:INSTALLER_DOWNLOAD_URL) {
+  $ArtifactDownloadUrl = $env:INSTALLER_DOWNLOAD_URL
+} else {
+  $ArtifactDownloadUrl = "$installer_base_url{{ hosting.github.artifact_download_path }}"
```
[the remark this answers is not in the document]
mistydemeo: The decision predates me, but I think it comes down to `$ArtifactDownloadUrl` originating as a CLI switch which enforced a particular variable naming scheme. The rest of the file is pretty settled on snake_case.

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -50,26 +50,45 @@ param (
...
+{% else %}
+if ($env:{{ env_vars.ghe_base_url_env_var }}) {
+  $installer_base_url = $env:{{ env_vars.ghe_base_url_env_var }}
+} elseif ($env:{{ env_vars.github_base_url_env_var }}) {
+  $installer_base_url = $env:{{ env_vars.github_base_url_env_var }}
+} else {
+  $installer_base_url = "{{ hosting.github.artifact_base_url }}"
+}
+if ($env:INSTALLER_DOWNLOAD_URL) {
+  $ArtifactDownloadUrl = $env:INSTALLER_DOWNLOAD_URL
```
fasterthanlime, line 66: this will help me in the future because I was about to file "$env:INSTALLER_DOWNLOAD_URL is ignored by ps1 installer" x) now I don't have to. I should have Simply Logged Off!

cargo-dist/src/tasks.rs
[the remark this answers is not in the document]
fasterthanlime: Yep sounds good, there's bigger messes I'm more worried about anyway.

cargo-dist/templates/installer/installer.ps1.j2
mistydemeo: Hahaha. Yes, we had a few params that were historically inconsistent between the two installers. I think this was the last one? We had at least one other where it was env var primarily for shell but CLI switch only for powershell, and I wasn't happy about the inconsistency

axoproject/src/repo.rs
```diff
@@ -20,9 +20,20 @@ pub struct GithubRepo {
 }
 
 impl GithubRepo {
+    /// Returns the domain. At the moment this is hardcoded to github.com, but
+    /// it may support more options in the future.
+    pub fn domain(&self) -> String {
+        "https://github.com".to_owned()
+    }
+
+    /// Similar to `web_url`, but returns just the path component without the domain.
```
ashleygwilliams, line 29:
```suggestion
    /// Path component. Used with `domain` to construct `web_url`.
```

cargo-dist-schema/src/lib.rs
```diff
@@ -772,7 +780,14 @@ pub struct Hosting {
...
+    /// (This can vary for GitHub Enterprise)
+    pub artifact_base_url: String,
+    /// The path of the release without the base URL
+    ///
+    /// e.g. `/myowner/myrepo/releases/download/v1.0.0/`
+    pub artifact_download_path: String,
+    /// The URL of the GitHub Release's artifact downloads
     ///
     /// e.g. `https://github.com/myowner/myrepo/releases/download/v1.0.0/`
     pub artifact_download_url: String,
```
ashleygwilliams, line 793: do we want to keep this in the struct if we are also storing it's components?

cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
ashleygwilliams: i was about to comment this lol

cargo-dist/src/host.rs
```diff
@@ -350,11 +353,15 @@ pub(crate) fn select_hosting(
     let repo = raw_repository_url
         .github_repo()
         .map_err(|e| DistError::CantEnableGithubUrlNotGithub { inner: e })?;
+    let domain = repo.domain();
     let repo_url = repo.web_url();
```
ashleygwilliams, line 357: do we keep this if we are constructing the others?

ashleygwilliams (commented):
my main feedback is that we seem to be storing both the constructed url and it's parts, which potentially could cause a consistency issue- could we consolidate?

cargo-dist-schema/src/lib.rs
mistydemeo: We have a few places that read the full URL, but we can probably have them just construct the URL from the parts instead of needing a single preconstructed URL.

mistydemeo: Thought about it again, and this is in the manifest - it's technically public data. Do we want to trust that none of our users used this?

ashleygwilliams: i am willing to take the heat if we are wrong!

cargo-dist/src/host.rs
mistydemeo: Looks like it's fully unused now, yeah. I'll cut it.

PR #1507, opened 2024-10-30
Have snapshots files be collapsed in GitHub by default

fasterthanlime:
See https://github.com/github-linguist/linguist/blob/main/docs/overrides.md

When looking at a PR like https://github.com/axodotdev/cargo-dist/pull/1503/files there's a _lot_ of scrolling involved just for .snap files —  sure we probably should open up a couple and check what's in there, but having 60+ of them opened by default just makes the GitHub UI sluggish and makes the review experience poorer.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,6 @@
 # Denote all files that are truly binary and should not be modified.
 *.png binary
 *.jpg binary
+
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Interesting - I hadn't seen this feature before! Makes sense to me.

PR #1508, opened 2024-10-30
Tune the 'dev' profile

fasterthanlime:
This has been extracted out of https://github.com/axodotdev/cargo-dist/pull/1470 and it's part of making a faster "edit-compile-test-in-ci" cycle when working on cross-compilation.

The rationale is included as inline comments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,15 @@ tag-name = "v{{version}}"
 pre-release-commit-message = "release: {{version}}"
 publish = false
 
+[profile.dev]
+# level true/2 is usually overkill, cf. <https://doc.rust-lang.org/cargo/reference/profiles.html#debug>
+debug = "limited"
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Looking good to me. If the `xz` compression bump is that significant, maybe it's worth considering a more targeted version of this for release builds too?

fasterthanlime:
> maybe it's worth considering a more targeted version of this for release builds too?

cargo defaults `opt-level` to 3 (!) for the release target: https://doc.rust-lang.org/cargo/reference/profiles.html?highlight=opt-level#release (for all dependencies). And has lto on. Which is the main reason release builds take forever.

2 is a good compromise for "dev" I think. We could introduce a separate "actual-debug" profile which has _full_ debug info, disables all optimizations, etc. — but in Rust, you really rarely need an actual debugger, unless the problem is in some C dependency.

This is going to make initial (cold/no cache) dev builds slower, but I don't think any of us will see a difference past the initial whole-project recompile. If we do, let's discuss that and make a follow-up change!

PR #1509, opened 2024-10-30
Add a `dist_url_override` config key

fasterthanlime:
...for quick iteration on cargo-dist itself.

The rationale is explained in the code comments itself.

I have tested this on Linux, via <https://github.com/fasterthanlime/bye/releases/tag/v0.6.0> — I have not tested the powershell version yet, lemme do that before taking this PR out of draft.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
         # we specify bash to get pipefail; it guards against the `curl` command
         # failing. otherwise `sh` won't catch that `curl` returned non-0
         shell: bash
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.24.0-prerelease.2/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.24.1/cargo-dist-installer.sh | sh"
       - name: Cache dist
         uses: actions/upload-artifact@v4
         with:
@@ -122,7 +122,8 @@ jobs:
         with:
           submodules: recursive
       - name: Install dist
-        run: ${{ matrix.install_dist }}
+        shell: ${{ matrix.install_dist.shell }}
+        run: ${{ matrix.install_dist.run }}
       # Get the dist-manifest
       - name: Fetch local artifacts
         uses: actions/download-artifact@v4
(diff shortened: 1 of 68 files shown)


cargo-dist/src/backend/ci/mod.rs
```diff
@@ -21,66 +25,133 @@ pub struct CiInfo {
...
+    version: &'a Version,
+    url_override: Option<&'a CargoDistUrlOverrideRef>,
 }
 
-/// Get the command to invoke to install dist via ps1 script
-fn install_dist_ps1_for_version(version: &Version) -> String {
-    if let Some(git) = install_dist_git(version) {
-        return git;
+/// Generates github steps to install a tool
+pub trait InstallStrategy {
```
mistydemeo, line 36: Nice call on making this a trait!

mistydemeo (approved):
Looking good. This'll be handy for testing, and may end up being useful for other people too.

cargo-dist/src/backend/ci/mod.rs
fasterthanlime: Thanks! It wasn't part of the original plan but the cargo auditable stuff happened!

fasterthanlime:
Note that the config key is named `cargo_dist_url_override`, following `cargo_dist_version`, we can add aliases for `dist_`-prefix versions later (serde makes that easy).

.github/workflows/release.yml
```diff
@@ -122,7 +122,8 @@ jobs:
         with:
           submodules: recursive
       - name: Install dist
-        run: ${{ matrix.install_dist }}
+        shell: ${{ matrix.install_dist.shell }}
```
mistydemeo, line 125: We probably have to regenerate our own release YAML now.

mistydemeo (approved):
Thanks for the quick turnaround on that fix. Just running a test now. https://github.com/mistydemeo/cargodisttest/actions/runs/11711074752

fasterthanlime:
> Thanks for the quick turnaround on that fix. Just running a test now. https://github.com/mistydemeo/cargodisttest/actions/runs/11711074752

I'm also running a test using the `recross` branch (cf. #1529)

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -284,7 +284,8 @@ jobs:
           cache-provider: ${{ matrix.cache_provider }}
       {{%- endif %}}
       - name: Install dist
-        run: ${{ matrix.install_dist }}
+        shell: ${{ matrix.install_dist.shell }}
```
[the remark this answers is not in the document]
fasterthanlime: Filed https://github.com/axodotdev/cargo-dist/issues/1532 for the bigger future change

fasterthanlime:
> Thanks for the quick turnaround on that fix. Just running a test now. https://github.com/mistydemeo/cargodisttest/actions/runs/11711074752

That test was successful and the tests are green. Merging!

PR #1516, opened 2024-11-01
Add cargo-auditable config option.

duckinator:
Moved to #1528 because the slash in the `duckinator/auditable-builds` branch name is incompatible with testing dist from a GitHub branch.

----

deferred:
- local auto-installation of cargo-auditable (moved to #1527)
- throw error when using cargo-auditable=true with non-rust projects (moved to #1526)

complete:
- [x] use `cargo auditable build` instead of `cargo build` if `cargo-auditable=true`
- [x] install cargo-auditable in workflows if `cargo-auditable=true`
- [x] tests for both of those
- [x] real-world local test
- [ ] real-world CI test

### DIFF
--- a/cargo-dist/templates/ci/github/release.yml.j2
+++ b/cargo-dist/templates/ci/github/release.yml.j2
@@ -133,6 +133,13 @@ jobs:
         with:
           name: cargo-dist-cache
           path: ~/.cargo/bin/dist
+      {{%- if need_cargo_auditable %}}
+      - name: Install cargo-auditable
+        # we specify bash to get pipefail; it guards against the `curl` command
+        # failing. otherwise `sh` won't catch that `curl` return non-0
+        run: {{{ install_cargo_auditable_sh }}}
+        shell: bash
+      {{%- endif %}}
       # sure would be cool if github gave us proper conditionals...
       # so here's a doubly-nested ternary-via-truthiness to try to provide the best possible
       # functionality based on whether this is a pull_request, and whether it's from a fork.
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -313,6 +313,9 @@ pub struct GithubMatrixEntry {
     /// what cache provider to use
     #[serde(skip_serializing_if = "Option::is_none")]
     pub cache_provider: Option<String>,
+    /// Expression to execute to install cargo-auditable
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub install_cargo_auditable: Option<String>,
 }
 
 /// Type of job to run on pull request
--- a/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
+++ b/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
@@ -1,6 +1,7 @@
 ---
 source: cargo-dist-schema/src/lib.rs
 expression: json_schema
+snapshot_kind: text
 ---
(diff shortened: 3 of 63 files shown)


duckinator:
image

Just needs tests and we're good to go.

duckinator:
This is the current failure mode for `auditable = true` without `cargo-auditable`. I have mixed feelings about it &mdash; it's not terrible, but something like this is probably better:

```
Setting `auditable = true` requires `cargo auditable`. You can install it with `cargo install cargo-auditable`
```

image

mistydemeo:
Since our main build environment is GitHub Actions at the moment, and it appears cargo auditable isn't preinstalled there, we should also make sure we handle installing it before doing local builds if it's in use. That'd be more impactful for our users than just telling them they need to install it. (We should also do that for non-CI builds though!)

ashleygwilliams:
cargo-auditable uses dist and so we can rely on their installers for our generated workflows and locally https://github.com/rust-secure-code/cargo-auditable/releases/tag/v0.6.4

mistydemeo:
I suppose then we should also consider: do we install a floating "whatever's latest" version? Or do we pick a known-good version and update it periodically?

ashleygwilliams:
@mistydemeo yeah it's a good question- i could see us going either way re pin latest or pick a known good release- both of those options beg the question of a config to override version which seems prudent, because we'll likely need it either way

ashleygwilliams:
@Shnatsel do you have thoughts on how we handle the version of cargo-auditable we install for folks?

mistydemeo (commented):
We discussed checking for if there are no cargo workspaces - you can handle that by checking the `DistGraphBuilder.workspaces` property, which has a list of all workspaces. Each of those has a `WorkspaceKind` you can key off of. You can see us using that here in `DistGraphBuilder::compute_build_steps`: https://github.com/axodotdev/cargo-dist/blob/d8ec1f9890f84e34f8a2f7b85f1c233003cf408a/cargo-dist/src/tasks.rs#L2435-L2448

`self.inner` on `DistGraphBuilder` is `DistGraph`, so you've got access to the config there. It's probably good for us to handle this stuff early, so when we're computing the builds - or something similar early in the dist graph construction process - is probably a good place for us to check and potentially error out.

Shnatsel:
@ashleygwilliams I treat the CLI as the public API for semver purposes, so installing the latest version in 0.6.x series is a safe bet. That should get you bug fixes without any kind of breakage (unless I mess up really badly).

You can revisit this if/when I ship 0.7.x, which right now is not even on the horizon.

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -133,6 +133,18 @@ jobs:
         with:
           name: cargo-dist-cache
           path: ~/.cargo/bin/dist
+      {{%- if need_cargo_auditable %}}
+      - name: Install cargo-auditable
+        # we specify bash to get pipefail; it guards against the `curl` command
+        # failing. otherwise `sh` won't catch that `curl` return non-0
+        run: {{{ install_cargo_auditable_sh }}}
+        shell: bash
+      - name: Cache cargo-auditable
```
mistydemeo, line 142: Feel free to skip this cache. We use this for cargo-dist itself because we reuse the same binary in multiple different jobs running on the same architecture, but the same isn't true for cargo-auditable.

duckinator:
Moving to #1528 because of branch naming making it harder to test.

Issue #1527, opened 2024-11-05
Handle local auto-installation of cargo-auditable


duckinator:
The `cargo-auditable` option is added in #1516. That PR ensures dist uses cargo-auditable and that cargo-auditable is installed in the generated workflows.

However, this leads the situation of running `dist build` locally.

I think there's agreement that this _should_ be at least offered, but we need to figure out the details of how/when to do it.

Issue #1526, opened 2024-11-05
Error when cargo-specific options used for non-cargo projects


duckinator:
If using a cargo-specific option &mdash; such as `cargo-auditable` &mdash; in a workspace with no Rust projects (or specifically on a non-Rust project), dist should raise an error.

(This is a follow-up to #1516.)

PR #1528, opened 2024-11-05
 Add cargo-auditable config option

duckinator:
deferred:
- local auto-installation of cargo-auditable (moved to #1527)
- throw error when using cargo-auditable=true with non-rust projects (moved to #1526)

complete:
- [x] use `cargo auditable build` instead of `cargo build` if `cargo-auditable=true`
- [x] install cargo-auditable in workflows if `cargo-auditable=true`
- [x] tests for both of those
- [x] real-world local test
- [x] real-world CI test

Moved from #1516 because the `duckinator/auditable-builds` branch name was incompatible with testing dist from a GitHub branch.

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -312,8 +312,8 @@ impl GithubCiInfo {
         for (runner, targets) in local_runs {
             use std::fmt::Write;
             let install_dist =
-                install_dist_for_targets(&targets, &install_dist_sh, &install_dist_ps1);
-            let install_cargo_auditable = install_dist_for_targets(
+                install_package_for_targets(&targets, &install_dist_sh, &install_dist_ps1);
+            let install_cargo_auditable = install_package_for_targets(
                 &targets,
                 &install_cargo_auditable_sh,
                 &install_cargo_auditable_ps1,
@@ -638,8 +638,8 @@ fn github_runner_for_target(
     }
 }
 
-/// Select the dist installer approach for a given Github Runner
-fn install_dist_for_targets<'a>(
+/// Select the (dist-produced) installer approach for a given Github Runner
+fn install_package_for_targets<'a>(
     targets: &'a [&'a TargetTripleRef],
     install_sh: &'a str,
     install_ps1: &'a str,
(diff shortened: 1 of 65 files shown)


cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -133,6 +133,13 @@ jobs:
         with:
           name: cargo-dist-cache
           path: ~/.cargo/bin/dist
+      {{%- if need_cargo_auditable %}}
+      - name: Install cargo-auditable
```
mistydemeo, line 137: Hm, I realized we're putting this in the wrong place. This is before `plan`, which plans out the job but doesn't actually perform the builds. What we actually want to do here is have `cargo auditable` installed for the local builds, which is the area starting L178 in this file.

You'll want to ensure that the github matrix entry which has the `install_dist` key gets a new `install_cargo_auditable` key, and that that gets assigned with a platform-appropriate value the same way it gets assigned for dist.

mistydemeo (changes requested):
Realized there's one mistake in the template setup - something needs to get moved.

duckinator:
```
$ cargo audit bin axolotlsay                                                                                                              
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 664 security advisories (from /home/puppy/.cargo/advisory-db)
    Updating crates.io index
       Found 'cargo auditable' data in axolotlsay (23 dependencies)
```

cargo-dist/src/backend/ci/github.rs
```diff
@@ -42,6 +42,10 @@ pub struct GithubCiInfo {
     pub install_dist_sh: String,
     /// expression to use for installing dist via powershell script
     pub install_dist_ps1: String,
+    /// expression to use for installing cargo-auditable via shell script
+    pub install_cargo_auditable_sh: String,
+    /// expression to use for installing cargo-auditable via powershell script
+    pub install_cargo_auditable_ps1: String,
```
mistydemeo, line 48: I believe we can drop these two fields since we're exclusively installing it via the GitHub matrix now?

mistydemeo (approved):
One minor, non-blocking comment. Otherwise, I think we're looking great here! The snapshot tests show the changes we'd want to see in the outputs, too.

PR #1529, opened 2024-11-06
Add cross-compilation support via cargo-zigbuild and cargo-xwin

fasterthanlime:
This PRs makes dist aware of the host/target divide, and teaches it to install and use tools like `cargo-zigbuild` and `cargo-xwin`.

Additionally, it teaches dist to run local jobs in containers (like the `manylinux` ones for compat reason, or the cargo-xwin ones, for "fast cross-compile" reasons).

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -403,7 +403,7 @@ impl GithubRunnerConfig {
         }
     }
 
-    /// cf. [`Self::GithubRunnerConfig`], but parsed
+    /// cf. [`Self::real_triple_name`], but parsed
     pub fn real_triple(&self) -> Triple {
         self.real_triple_name().parse().unwrap()
     }
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -9,19 +9,26 @@
 //! The root type of the schema is [`DistManifest`][].
 
 pub mod macros;
+pub use target_lexicon;
 
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, str::FromStr};
 
 use schemars::JsonSchema;
 use semver::Version;
 use serde::{Deserialize, Serialize};
+use target_lexicon::Triple;
 
 declare_strongly_typed_string! {
-    /// A rust target-triple (e.g. "x86_64-pc-windows-msvc")
-    pub struct TargetTriple => &TargetTripleRef;
+    /// A rustc-like target triple/tuple (e.g. "x86_64-pc-windows-msvc")
+    pub struct TripleName => &TripleNameRef;
 }
 
-impl TargetTripleRef {
+impl TripleNameRef {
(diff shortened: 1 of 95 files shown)


cargo-dist/src/build/cargo.rs
```diff
@@ -66,7 +68,8 @@ impl<'a> DistGraphBuilder<'a> {
         }
 
         let mut builds = vec![];
-        for (target, binaries) in targets {
+        for (target_triple, binaries) in targets {
+            let target = target_triple.parse().unwrap();
```
mistydemeo, line 72: Given we're still in the plan step, and this returns a `DistResult`, we can safely return an error here if we get an unparseable triple. Just need to make sure we return a user-legible error.

cargo-dist/src/build/cargo.rs
```diff
@@ -156,34 +169,68 @@ impl<'a> DistGraphBuilder<'a> {
...
-    target: &CargoBuildStep,
+/// Generate a `cargo build` command
+pub fn make_build_cargo_target_command(
+    cargo_cmd: &str,
+    rustflags: &str,
+    step: &CargoBuildStep,
     auditable: bool,
 ) -> Cmd {
-    let mut command = Cmd::new(cargo_cmd, "build your app with Cargo");
+    let host = cargo_dist_schema::target_lexicon::HOST;
```
mistydemeo, line 179: Note for myself: we can replace the `current_platform` usage with this once this PR is merged.

cargo-dist-schema/src/lib.rs
```diff
@@ -905,6 +917,11 @@ pub enum PackageManager {
     Apt,
 }
 
+declare_strongly_typed_string! {
+    /// A homebrew package name, cf. <https://formulae.brew.sh/>
+    pub struct HomebrewPackageName => &HomebrewPackageNameRef;
```
mistydemeo, line 922: Just a note we can unify this with the other definitions alone.

mistydemeo (commented):
Approach is looking really good, and quite understandable!

cargo-dist/src/build/cargo.rs
fasterthanlime: So this is delicate — `target_lexicon::HOST` is "what the currently-running dist binary was built for".

On the same Linux system, you can get either `x86_64-unknown-linux-gnu` or `x86_64-unknown-linux-musl` depending on which dist binary you're running. That was surprising to me when the codepath suddenly started considering this a cross 🙃

fasterthanlime:
I'm done with this PR (cc @mistydemeo) — there's docs in there, and `fasterthanlime/bye` is showcasing several ways of using cross-compilation: https://github.com/fasterthanlime/bye/actions/runs/11747958257/job/32731139531

(It's not hidden behind flags or whatever, it just has some extra knowledge and lets you specify even more precisely what you want).

mistydemeo (approved):

PR #1531, opened 2024-11-06
Add cargo-cyclonedx config option.

duckinator:


cargo-dist/src/tasks.rs
```diff
@@ -1400,6 +1400,11 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
         }
     }
 
+    fn add_cyclonedx_sbom_file(&mut self, to_release: ReleaseIdx) {
+        // FIXME: Figure out the location
+        eprintln!("add_cyclonedx_sbom_file()");
```
mistydemeo, line 1405: You mentioned this earlier, but looks like it places the output in a file named `packagename/packagename.cdx.xml`.

The relationship between release/package/etc. is kind of thorny here, so let me outline the shape of what you need here. You're going to want this to have  `to_package: PackageIdx` in addition to  `to_release: ReleaseIdx` (I'll get into why the release is necessary later). You can get that index from `info.package_idx` within `compute_releases`, like on this line: https://github.com/axodotdev/cargo-dist/blob/f8e0688ce296ed262d6d4c63c676f62fb5e3e48b/cargo-dist/src/tasks.rs#L2629

With access to that, you can `let package = self.workspaces.package(package_idx)`. That gives you an `axoproject::PackageInfo`: https://github.com/axodotdev/cargo-dist/blob/f8e0688ce296ed262d6d4c63c676f62fb5e3e48b/axoproject/src/lib.rs#L404

Which itself has the `true_name` and `package_root` fields you need to construct the artifact path. (The `true_name` here is because the name can be overridden in dist config, but we're talking to Cargo and so we need the genuine Cargo name.)

The method itself can then access config that can conditionally enable or disable itself. We'll need to grab the release so we can get the release-specific/package-scoped config (since I believe you've set the cyclonedx stuff to be something you can enable/disable per-package), then return if it's disabled.

```rust
fn add_cyclonedx_sbom_file(&mut self, to_package: PackageIdx, to_release: ReleaseIdx) {
  let release = self.release(to_release);

  if !self.global_artifacts_enabled() || !release.config.builds.cargo.cargo_cyclonedx {
    return
 }

  // From here you can safely proceed with constructing the artifact
}
```

duckinator:
Okay, it's adding the file to the archive, but I believe it needs to put it in `target/distrib/` so that `target/distrib/` is treated as the top-level of the generated archive.

Working build without SBOMs:

> The least common ancestor is /home/runner/work/dist-sometest/dist-sometest/target/distrib. This will be the root directory of the artifact

Failed build with SBOMs:

> The least common ancestor is /home/runner/work/dist-sometest/dist-sometest. This will be the root directory of the artifact

This is what the SBOM gave me:

image

duckinator:
https://github.com/duckinator/dist-sometest/releases/tag/v0.0.9 it's included in an actual release so this should be good for review

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -402,12 +409,30 @@ jobs:
           echo "EOF" >> "$GITHUB_OUTPUT"
 
           cp dist-manifest.json "$BUILD_MANIFEST_NAME"
+      {{%- if need_cargo_cyclonedx %}}
+      - id: cargo-cyclonedx
+        shell: bash
+        run: |
+            # Generate SBOM (.cdx.xml) files.
+            cargo cyclonedx -v
```
mistydemeo, line 417: At some point, it may be useful for us to move this into Rust so that we can also run this locally as a part of global builds. I don't think that's a priority for us right now, given we don't install it for users outside of CI and most people are going to be doing this in our actions runs.

mistydemeo (approved):
Looking good! The one thing I think we could do some extra testing on is a multi-crate workspace published in the same release - we don't AFAICT have anything like that in our test suite at the moment, so I don't think it needs to block merging, but it would be useful given it produces package-specific artifacts.

cargo-dist/templates/ci/github/release.yml.j2
duckinator: It might make sense to ask the `cargo-cyclonedx` folks about an option to dump all the generated files in a specific directory. That'd _greatly_ simplify doing this from Rust, because we could avoid having to move files around after running it.

PR #1537, opened 2024-11-07
Allow per-target glibc version overrides.

duckinator:


book/src/reference/config.md
```diff
@@ -38,6 +38,7 @@ We're currently in the middle of [a major config migration](https://github.com/a
     * [`all-features`](#all-features)
     * [`default-features`](#default-features)
     * [`features`](#features)
+    * [`minimum-glibc-version`](#minimum-glibc-version)
```
ashleygwilliams, line 41: i might shorten this to `min` to reduce possibility of mispellings

cargo-dist/src/init.rs
```diff
@@ -1428,3 +1428,25 @@ fn apply_optional_mac_pkg(
         table.remove(key);
     }
 }
+
+/// Similar to [`apply_optional_value`][] but specialized to `MinimumGlibcVersion`, since we're not able to work with structs dynamically
```
ashleygwilliams, line 1432: can you paste what the init experience is for this?

cargo-dist/src/platform.rs
```diff
@@ -829,33 +832,49 @@ fn max_of_min_libc_versions(
...
+
     if let Some(artifact) = manifest.artifacts.get(artifact_id) {
         for asset in &artifact.assets {
             let asset_conditions = native_runtime_conditions_for_asset(manifest, &asset.id);
             runtime_conditions.merge(&asset_conditions);
         }
     };
+
     // FIXME: in our test suite we're running bare artifacts=global so we're missing
     // all artifact/linkage info, preventing basic glibc bounds
```
ashleygwilliams, line 849: is this true? if so, does this change affect this comment?

cargo-dist/tests/integration-tests.rs
```diff
@@ -37,16 +37,19 @@ cargo-dist-version = "{dist_version}"
 installers = ["shell", "powershell", "homebrew", "npm", "msi", "pkg"]
 tap = "axodotdev/homebrew-packages"
 publish-jobs = ["homebrew", "npm"]
-targets = ["x86_64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows-msvc", "aarch64-apple-darwin"]
+targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-pc-windows-msvc", "aarch64-apple-darwin"]
```
ashleygwilliams, line 40: so one reason we didn't use arm linux in our integration tests is because we need to pay for those runners- is it necessary to use this target to test what you want to test?

duckinator: that can be changed, although it needs to be another linux-gnu target. it looks like there's an i686-unknown-linux-gnu. i can use that instead. :+1:

book/src/reference/config.md
duckinator: I'm okay with this suggestion. :+1:

cargo-dist/src/init.rs
duckinator: It's used like this at line 1310:

```rust
    apply_optional_min_glibc_version(
        table,
        "minimum-glibc-version",
        "# The minimum glibc version supported by the package (overrides auto-detection)\n",
        minimum_glibc_version.as_ref(),
    );
```

book/src/reference/config.md
duckinator: I just realized all the existing code for auto-detecting glibc stuff was called `min_glibc_version`, so shortening it improves the internal consistency of the codebase, too.

cargo-dist/src/init.rs
ashleygwilliams: ah my bad- i meant what an end user would see

duckinator: Since I had to change the field names for some of the v1 config to make this work, I changed all of it to avoid weird mismatches between "minimum" and "min" being scattered all over the place.

book/src/reference/config.md
duckinator: Since I had to change the field names for some of the v1 config to make this work, I changed all of it to avoid weird mismatches between "minimum" and "min" being scattered all over the place.

cargo-dist/src/platform.rs
duckinator: Yes, it looks like we are in fact using artifacts=global in some situations.

As I suspected, this change (and technically the PR adding minimum-glibc-version) means the comment should be moved.

cargo-dist/src/init.rs
duckinator: Oh! There's two variants.

First, as its own table:

```
[dist.min-glibc-version]
"*" = "2.17"
i686-unknown-linux-gnu = "2.18"
```

Second, this is the equivalent when put directly under `[dist]`. I don't recommend it unless you're only overriding one value, but TOML allows it:

```
[dist]
# ...
min-glibc-version."*" = "2.17"
min-glibc-version.i686-unknown-linux-gnu = "2.18"
```

duckinator:
Okay, I believe I addressed all your comments, and also rebased off main. :+1:

cargo-dist/src/init.rs
ashleygwilliams: haha yeah i saw this in the code and it's so wild looking haha

ashleygwilliams (approved):

PR #1538, opened 2024-11-08
Add patches to installer to support transition from `cargo-home` to `flat` install layouts

zanieb:
This adds patches to the installer to address https://github.com/axodotdev/axoupdater/issues/210 — these patches are more robust versions of the patches described in that issue.

It's important to have a compatibility shim like this at the installer level because the `axoupdater` cannot be fixed for previous versions of uv. ~Arguably, this is pretty specific to us since nobody else has switched layouts and complained that updates were broken. If there's a reasonable way for us to insert the patches on our end, that's fine with me too.~ Even if an `axoupdater` fix is released soon, this change will be relevant to users upgrading from _older_ versions of an application to the latest version.

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -302,6 +302,22 @@ function Invoke-Installer($artifacts, $platforms) {
     $install_layout = "flat"
   }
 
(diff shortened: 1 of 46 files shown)


zanieb:
I tested similar patches in production for uv, but **these changes are not tested yet**.

ashleygwilliams (approved):

PR #1551, opened 2024-11-13
Avoid discarding tag information, so "dist plan --tag={name}-{version}" works.

duckinator:


mistydemeo (approved):
Good find here, and it'll be useful to see where this ends up exposing slight inconsistencies we hadn't considered.

PR #1566, opened 2024-11-19
Explicitly set manifest path when building MSI with WiX

pnehrer:
Use an explicit manifest path when building MSI with WiX, in case Cargo.toml is in a sub-folder.

Unfortunately I missed this in https://github.com/axodotdev/cargo-dist/pull/1454.


mistydemeo (approved):
Good catch, thanks!

Sorry for the late response; I've been out of town for a couple of weeks.

PR #1567, opened 2024-11-19
Have workflows run `dist plan` on all known tags.

duckinator:
General idea:
- add a command (or flag for dist plan) that prints a JSON representation of all possible tags
- in the generated workflow, use that new command to determine what tags there are, and run `dist plan --output-format=json --tag=${TAG} > plan-dist-manifest-${TAG}.json` for each one

Replaces #1560.

### DIFF
--- a/cargo-dist/src/cli.rs
+++ b/cargo-dist/src/cli.rs
@@ -134,6 +134,8 @@ pub enum Commands {
     /// Report on the dynamic libraries used by the built artifacts.
     #[clap(disable_version_flag = true)]
     Linkage(LinkageArgs),
+    /// List all version tags known by dist.
+    ListTags(ListTagsArgs),
     /// Generate the final build manifest without running any builds.
     ///
     /// This command is designed to match the exact behaviour of
@@ -331,6 +333,9 @@ pub struct LinkageArgs {
     #[clap(long)]
     pub from_json: Option<String>,
 }
+
+#[derive(Args, Clone, Debug)]
+pub struct ListTagsArgs {}
 
 #[derive(Args, Clone, Debug)]
 pub struct HelpMarkdownArgs {}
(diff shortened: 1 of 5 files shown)


cargo-dist/src/cli.rs
```diff
@@ -134,6 +134,8 @@ pub enum Commands {
     /// Report on the dynamic libraries used by the built artifacts.
     #[clap(disable_version_flag = true)]
     Linkage(LinkageArgs),
+    /// List all version tags known by dist.
+    ListTags(ListTagsArgs),
```
ashleygwilliams, line 138: i wonder if we shouldn't make this a tag command with the subcommand list, e.g. dist tag list

duckinator:
As discussed on Discord, currently setting this aside because of the complexity caused by this: https://github.com/axodotdev/cargo-dist/blob/b487ee9c61681f4532b2d4bd1d09ca32484bffb9/cargo-dist/templates/ci/github/release.yml.j2#L143-L149

duckinator:
@alilleybrinker unfortunately the scope on this grew pretty significantly because of how much it interacts with other parts of the codebase. I ran into a thing blocking me from making any progress at all today, so I was asked to switch to a task I can actually make progress on. I'm not sure what the plan for this task is after today; I'd need to discuss that with @ashleygwilliams.

If you do try your own temporary fork and would like my input on something, I tend to be on the axodotdev Discord 10am-4pm eastern (7am-1pm pacific) on weekdays.

alilleybrinker:
No problem! I'm talking with Ashley as well and will see what we can do. Thanks so much for the help so far on this, including the draft changes here! Totally understand about this hitting some complex parts of the system.

duckinator:
We went with #1579 instead.

PR #1568, opened 2024-11-19
Add option to generate omnibor artifact IDs.

duckinator:


duckinator:
Current state:

- test suite runs locally
- real-world test ([duckinator/dist-sometest v0.0.14](https://github.com/duckinator/dist-sometest/releases/tag/v0.0.14)) works
- CI fails, I believe because omnibor-cli needs to be installed in the ci.yml workflow

duckinator:
Alright, this should be ready to go!

- test suite passes locally
- real-world test ([duckinator/dist-sometest v0.0.14](https://github.com/duckinator/dist-sometest/releases/tag/v0.0.14)) works
- CI passes

ashleygwilliams:
@alilleybrinker would be curious if you have any thoughts on this!

alilleybrinker:
> @alilleybrinker would be curious if you have any thoughts on this!

Taking a look!

alilleybrinker (approved):
LGTM! I see you settled on a `.omnibor` file containing the Artifact ID, which seems fine to me. I don't think the OmniBOR project has a consensus on the naming of this file, and this seems as good a solution as any!

cargo-dist/src/lib.rs
```diff
@@ -416,6 +424,23 @@ fn generate_fake_pkg(
     Ok(())
 }
 
+fn generate_omnibor_artifact_id(src_path: &Utf8Path, dest_path: &Utf8Path) -> DistResult<()> {
+    let mut cmd = Cmd::new("omnibor", "generate an OmniBOR Artifact ID");
```
mistydemeo, line 428: How would you feel about adding `omnibor` to the `Tools` struct? That would make it easier for us to handle the "omnibor not installed" case: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/src/tasks.rs#L264-L282

cargo-dist/src/tasks.rs
```diff
@@ -1524,6 +1537,69 @@ impl<'pkg_graph> DistGraphBuilder<'pkg_graph> {
...
+        let artifacts: Vec<_> = global_artifacts
+            .into_iter()
+            .chain(local_artifacts)
+            .map(|a| self.artifact(a).clone())
+            .collect();
+
+        for artifact in &artifacts {
+            let id = artifact.id.clone();
+            let src_path = artifact.file_path.clone();
+            match &artifact.kind {
```
mistydemeo, line 1561: Nit: if this is written as an `if matches!(artifact.kind, ...)` condition, you could skip the unwanted `ArtifactKind` cases and only handle the desired ones.

cargo-dist/src/config/v0.rs
```diff
@@ -467,6 +467,11 @@ pub struct DistMetadata {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
     pub cargo_cyclonedx: Option<bool>,
+
+    /// Whether to generate OmniBOR artifact IDs.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    pub omnibor: Option<bool>,
```
mistydemeo, line 474: I wonder if we want to name this something more specific? Like `generate-omnibor-id`.

mistydemeo (commented):
Looks good, just a couple comments.

cargo-dist/src/tasks.rs
duckinator: This was actually deliberate, because it means if another `ArtifactKind` is added the compiler will tell us we need to decide where it goes, instead of just assuming all new `ArtifactKind`s aren't relevant.

If you think I should change it anyway, lmk and I will. Otherwise I'll add a comment explaining why I did it.

mistydemeo: Ah, good call! That makes sense.

mistydemeo (approved):
Looking good, though we should decide on the final config option before this is merged. Also looks like the tests need to be adjusted for the new field in the `Tools` struct.

PR #1579, opened 2024-11-27
feat: Permit tag incoherence for `dist plan`

alilleybrinker:
This is part of a broader goal of permitting `dist plan` to run in a `dist` workspace with multiple packages which are independently versioned. Today, `dist plan` fails (and thus `dist`'s GitHub Actions on PRs fail) in such a workspace because it can't resolve a coherent tag.

Under the hood, `dist plan`'s `cmd_plan` function delegates to `cmd_manifest` after configuring some settings, and `cmd_manifest` itself further configures the system to demand tag coherency.

This commit splits out the body of `cmd_manifest` into a new `generate_manifest` function, which is equivalent to `cmd_manifest` except it accepts a new `needs_coherence` parameter (so named to match the `Cli::tag_settings` method's parameter name). With this new function introduced, `cmd_manifest` is updated to call `generate_manifest` with `needs_coherence` set to `true`, while `cmd_plan` can call it with `needs_coherence` set to `false`.

### DIFF
--- a/cargo-dist/src/main.rs
+++ b/cargo-dist/src/main.rs
@@ -280,19 +280,8 @@ fn cmd_host(cli: &Cli, args: &HostArgs) -> Result<(), miette::Report> {
 }
 
 fn cmd_manifest(cli: &Cli, args: &ManifestArgs) -> Result<(), miette::Report> {
-    let config = cargo_dist::config::Config {
-        tag_settings: cli.tag_settings(true),
-        create_hosting: false,
-        artifact_mode: args.build_args.artifacts.to_lib(),
-        no_local_paths: cli.no_local_paths,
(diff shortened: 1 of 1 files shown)


alilleybrinker:
Note that this does not include checking the `pr_run_mode` and producing an error if `pr_run_mode == "upload"`. I'm happy to add that if you'd like, and I've also set "allow edits by maintainers" so you can add it if you'd like.

duckinator (approved):
This looks good, and worked both in local testing and some real-world tests.

Only known issue is that if you have incoherent tags and use `pr-run-mode = "upload"`, PRs will fail at the build-local-artifacts job. I'm planning to make a separate PR to address that.

PR #1586, opened 2024-12-02
feat: add pc-windows-gnu to powershell installers

### LINKED ISSUE #1585: PowerShell installer not generated for pc-windows-gnu
We've been improving support for `pc-windows-gnu` targets recently, but a major exception: the powershell installers skip all `pc-windows-gnu` builds. We should make sure this gets fixed.

mistydemeo:
Before this, even though we could build `pc-windows-gnu` installers, we wouldn't actually use them in installers. There were two issues:

* The code deciding when to build powershell installers only selected `pc-windows-msvc`, and always skipped `pc-windows-gnu`. If `pc-windows-gnu` binaries were the only ones available, then the installer wouldn't be generated.
* The installer itself hardcoded `-msvc` and needed to be updated to detect `-gnu`. I've updated it to actually look at the list of targets - if both `-msvc` and `-gnu` are present, we pick `-msvc`, but we now correctly pick `-gnu` from the list if it's all we have.

Fixes #1585.


mistydemeo:
This doesn't quite work - I need to add it to the platform fallback code too.

duckinator (approved):
This looks good, with the caveat that I don't currently have the means to manually test it on Windows.

mistydemeo:
Oh yes, forgot to comment: this works now, and I've added the -gnu packages to the platforms list. Hypothetically we could handle the msvc/gnu disambiguation in the platform code, but it's already a little messy about that. If we're going to make a change around that, it should get done later.

PR #1587, opened 2024-12-02
If pr-run-mode="upload", fail in the "plan" job instead of the "build-local-artifacts" job.

duckinator:


cargo-dist/src/main.rs
```diff
@@ -319,6 +319,24 @@ fn generate_manifest(
         root_cmd: "plan".to_owned(),
     };
     let report = do_manifest(&config)?;
+
+    // FIXME: The build-local-artifacts job in generated workflows can't handle
+    //        can't handle incoherent tags, so this makes it fail in the
+    //        `dist plan` stage instead of waiting.
+    if !needs_coherence {
+        if let Some(pr_run_mode) = report
```
mistydemeo, line 327: Nit: instead of cloning, this can probably be

```suggestion
        if report
           .ci
           .and_then(|ci| ci.github.as_ref())
           .and_then(|gh| gh.pr_run_mode)
            == Some(cargo_dist_schema::PrRunMode::Upload)
```

And then skip the inner `pr_run_mode ==` comparison.

mistydemeo (approved):
Makes sense to me - definitely a better user experience.

cargo-dist/src/main.rs
duckinator: I wasn't able to use your code exactly as-is, but I tweaked it a bit and was able to combine everything into one conditional with no clone.

PR #1588, opened 2024-12-02
Only include cargo-auditable installation commands when it's needed.

### LINKED ISSUE #1569: Update `cargo-auditable` integration to use `.then_some()`
When working on #1568 I figured a way to keep optional settings from being included in serialized forms of `GithubCiInfo` and such.

The gist of it is:
- have them be `Option<...>`,
- use `#[serde(skip_serializing_if = "Option::is_none")]`
- when referencing them, use something along the lines of `need_thing.then_some(thing)`

When adding `cargo-auditable`, it touched almost every insta snapshot, but by using this approach in #1568 it only touches snapshots enabling the omnibor integration.

This will wind up touching all those snapshots again, _but_ it seems like a good practice to adopt.

https://github.com/axodotdev/cargo-dist/blob/f915d8748428a563b9a97ebb532f0599f2032b38/cargo-dist/src/backend/ci/github.rs#L276-L277

https://github.com/axodotdev/cargo-dist/blob/f915d8748428a563b9a97ebb532f0599f2032b38/cargo-dist/src/backend/ci/github.rs#L348-L349

duckinator:
Follow up for #1528, which uses `Option::then_some` to avoid including `cargo-auditable` installation commands when they won't be used.

Closes #1569.

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -477,7 +477,8 @@ pub struct GithubLocalJobConfig {
     pub targets: Option<Vec<TripleName>>,
 
     /// Expression to execute to install cargo-auditable
-    pub install_cargo_auditable: GhaRunStep,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub install_cargo_auditable: Option<GhaRunStep>,
 
     /// Command to run to install dependencies
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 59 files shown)


mistydemeo (approved):
Confirmed no diff from the `axolotlsay_basic` test, which looks right.

mistydemeo:
Also looks like this doesn't trip up the undefined behaviour check we added recently.

PR #1595, opened 2024-12-03
Improve interactive `dist plan` output if there's multiple versions in a workspace

duckinator:
This only affects the "human" output.

Output for an "incoherent" plan with multiple different versions:

image

Output for a "coherent" plan with only one version:

image


mistydemeo (approved):
This output is nice and clear - looks like it should be a good improvement for users.

PR #1597, opened 2024-12-04
feat: disable axoupdater-from-source

mistydemeo:
This was never all that reliable, and it's become less reliable since shipping cross compilation. It's also not really used: we provide binaries for all of the common release targets. We can consider restoring a more functional version in the future, but for now we should turn it off.

### DIFF
--- a/cargo-dist/src/errors.rs
+++ b/cargo-dist/src/errors.rs
@@ -365,6 +365,16 @@ pub enum DistError {
         tool: String,
     },
 
+    /// Unknown target requested
+    #[error(
+        "A build was requested for {target}, but the standalone updater isn't available for it."
+    )]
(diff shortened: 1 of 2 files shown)


cargo-dist/src/errors.rs
```diff
@@ -599,18 +609,6 @@ pub enum DistError {
         /// Version the project uses
         your_version: semver::Version,
     },
-
-    /// axoupdater can't be built for this target from this host
-    #[error("You've requested a cross-compile from {host} to {target}, but axoupdater can't be built for this target")]
-    #[diagnostic(help(
-        "Please either disable this target, or set `install-updater' to `false' in your config."
```
ashleygwilliams, line 606: this isn't really a comment for this PR but i would expect that folks would like the opportunity to turn off the updater for certain targets

mistydemeo: True, that could be a useful feature in the future!

duckinator (approved):
This looks good to me. :+1:

PR #1598, opened 2024-12-04
feat: update axoupdater, fetch known-good version

mistydemeo:
Instead of always fetching the latest axoupdater, fetch the version that this version of dist itself uses as a library. That way, we're always fetching something predictable.

### DIFF
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -251,9 +251,12 @@ fn run_build_step(
     Ok(())
 }
 
-const AXOUPDATER_ASSET_ROOT: &str =
-    "https://github.com/axodotdev/axoupdater/releases/latest/download";
+const AXOUPDATER_ASSET_ROOT: &str = "https://github.com/axodotdev/axoupdater/releases";
 const AXOUPDATER_MINIMUM_VERSION: &str = "0.7.0";
+
+fn axoupdater_asset_root() -> String {
+    format!("{AXOUPDATER_ASSET_ROOT}/download/v{}", axoupdater::VERSION)
+}
 
 /// Fetches an installer executable and installs it in the expected target path.
 pub fn fetch_updater(dist_graph: &DistGraph, updater: &UpdaterStep) -> DistResult<()> {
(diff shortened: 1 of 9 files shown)


cargo-dist/src/lib.rs
```diff
@@ -173,11 +173,14 @@ fn run_build_step(
...
 }
 
-const AXOUPDATER_ASSET_ROOT: &str =
-    "https://github.com/axodotdev/axoupdater/releases/latest/download";
+const AXOUPDATER_ASSET_ROOT: &str = "https://github.com/axodotdev/axoupdater/releases/";
 const AXOUPDATER_MINIMUM_VERSION: &str = "0.7.0";
 const AXOUPDATER_GIT_URL: &str = "https://github.com/axodotdev/axoupdater.git";
 
+fn axoupdater_asset_root() -> String {
+    format!("{AXOUPDATER_ASSET_ROOT}/v{}/download", axoupdater::VERSION)
```
ashleygwilliams, line 181: does this prevent auto-updates?

mistydemeo: No, this is where we'll fetch standalone axoupdater builds from. It just means a given release of dist will always install a specific version of the standalone updater when packaging software, instead of a floating version. dist's own autoupdates will work.

ashleygwilliams: ah sorry i just realized how vague what i wrote was... i meant to ask if this means that a single version of dist will only package a single version of axoupdater and to use a new version of axo updater we'd need to make a new dist release

mistydemeo: Ah, got it. Yes, I felt that this was probably a good thing at this point - protection against breaking changes without a corresponding dist release, now that axoupdater is out of its initial experimental phase.

duckinator (approved):
This seems like a good improvement to me! As you mentioned in <https://github.com/axodotdev/cargo-dist/pull/1598#r1870440361>, it seems prudent to avoid breaking changes happening where we don't expect them.

mistydemeo:
Added an off-by-default option to always use the latest axoupdater.

ashleygwilliams (approved):
🚀

mistydemeo:
Confirmed the flag works - build with the flag fetching the `latest` updater: https://github.com/mistydemeo/cargodisttest/actions/runs/12383976628/job/34567829105#step:9:77

And build without fetching the fixed version: https://github.com/mistydemeo/cargodisttest/actions/runs/12383684253/job/34566994890#step:9:77

PR #1604, opened 2024-12-04
Migrate deserialize_yaml() to axoasset.

duckinator:


duckinator:
Once a new version of axoasset with axodotdev/axoasset#235 is merged, I can make Cargo.toml not use git for axoasset, and _then_ this will be ready.

mistydemeo (commented):
Looks like a surprisingly big lockfile change - did this incorporate a `cargo update` or something?

duckinator:
Not sure what caused it, but I checked out `Cargo.lock` and re-generated it and the diff is way smaller.

mistydemeo (approved):

PR #1607, opened 2024-12-05
release: 0.26.0

mistydemeo:
We're not quite ready to tag this - we have a few action items before then.

- [x] Ensure all docs for the new features are correct
- [x] Write docs for cross-compilation
- [x] Implement #1526.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Nothing Yet!
 
-# Version 0.26.0 (20XX-XX-XX)
+# Version 0.26.0 (2024-12-12)
 
 It's been slightly longer than usual since our last release, and now we're back with a slightly larger than usual release! This version brings several major new features and improvements, including the long-requested Rust cross-compilation feature and support for a few different Rust dependency version tracking formats.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,77 @@
 
 Nothing Yet!
 
+# Version 0.26.0 (20XX-XX-XX)
+
+It's been slightly longer than usual since our last release, and now we're back with a slightly larger than usual release! This version brings several major new features and improvements, including the long-requested Rust cross-compilation feature and support for a few different Rust dependency version tracking formats.
+
+## Builtin Rust cross-compilation support
+
+You've all asked for it, and it's finally here! Previously, we only supported Rust cross-compilation on macOS. With this release, we've extended Rust cross-compilation support to Linux (using [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) and Windows (using [cargo-xwin](https://github.com/rust-cross/cargo-xwin). For more information, see [the docs](https://opensource.axo.dev/cargo-dist/book/ci/customizing.html#cross-compilation).
+
+We're also making use of this feature ourselves: we now build our aarch64 Linux binaries using this new tooling.
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,77 @@
...
+
+While we've always generated checksum information for binary tarballs/ZIPs, we only actually validated those checksums in the Homebrew installer. That changes with this release: we now embed checksum information into the shell installer and validate the tarball before unpacking it.
+
+- impl @fasterthanlime [Verify checksums in install.sh](https://github.com/axodotdev/cargo-dist/pull/1497)
+
+## cargo-auditable support
+
+We've added integrated support for the Rust Secure Code Working Group's [cargo-auditable](https://github.com/rust-secure-code/cargo-auditable), which embeds dependency information in your Rust binaries and makes it possible for users to check your binaries for the full dependency tree they were built from with their precise versions. For more information, see [our docs](https://opensource.axo.dev/cargo-dist/book/reference/config.html#cargo-auditable) and the [docs for the cargo-audit tool](https://github.com/rustsec/rustsec/blob/main/cargo-audit/README.md).
+
+(Note: for technical reasons, this feature and cross-compilation are currently mutually exclusive. Users can only enable one or the other.)
```
ashleygwilliams, line 29: this is interesting... should we file an issue to eliminate this?

CHANGELOG.md
```diff
@@ -2,6 +2,77 @@
...
+
+## cargo-cyclonedx support
+
+We've also added support for generating [CyloneDX Software Bill of Materials (SBOM)](https://cyclonedx.org) files for Rust projects. We've implemented this using the [cargo-cyclonedx](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/main/cargo-cyclonedx/README.md) tool. Unlike the cargo-auditable support above, which embeds dependency information directly into your binaries, this data is stored as a standalone `bom.xml` file which is distributed with your software. Users can then validate that SBOM file using [any compatible CycloneDX tool](https://cyclonedx.org/tool-center/).
+
+- impl @duckinator [Add cargo-cyclonedx config option.](https://github.com/axodotdev/cargo-dist/pull/1531)
+
+## omnibor support
+
+Rounding out this release's new security features, we've added support for [OmniBOR](https://omnibor.io) dependency tracking metadata for Rust projects. We implement this using the [omnibor-cli](https://github.com/omnibor/omnibor-rs/tree/main/omnibor-cli) tool. For more information, see [the docs](https://opensource.axo.dev/cargo-dist/book/reference/config.html#omnibor).
```
ashleygwilliams, line 41: so OmniBOR is not actually tracking dependencies so much as providing a way to identify or fingerprint the built software (which includes dependencies, but also includes first party code)

CHANGELOG.md
mistydemeo: Unfortunately, I don't think we can. All three of `cargo-auditable`, `cargo-xwin` and `cargo-zigbuild` are implemented as cargo subcommands that wrap the real cargo builds. It's not possible to run more than one of them in the same build.

CHANGELOG.md
mistydemeo: Oh, thanks for the clarification! I'll fix this.

CHANGELOG.md
messense: IMO this is not strictly true, see https://github.com/rust-cross/cargo-zigbuild/issues/192

mistydemeo: Yep, been looking into it - #1608. Looking like the zigbuild stuff may be incompatible for other reasons, at least in testing so far.

ashleygwilliams (approved):

PR #1611, opened 2024-12-06
Refactor `dist init` and split out `dist migrate`

duckinator:
- `dist init` still works as on main
- `dist migrate` does the migration part of `dist init`
- `dist init`'s migration functionality is now implemented in terms of `dist migrate`

This shows `dist migrate` on its own:

image

And this shows that `dist migrate` is _only_ a migration, and explicitly does not do the re-init part:

image

### DIFF
--- a/cargo-dist/src/init.rs
+++ b/cargo-dist/src/init.rs
@@ -52,6 +52,90 @@ fn theme() -> dialoguer::theme::ColorfulTheme {
     }
 }
 
+/// Copy [workspace.metadata.dist] from one workspace to [dist] in another.
+fn copy_cargo_workspace_metadata_dist(
+    new_workspace: &mut toml_edit::DocumentMut,
+    workspace_toml: toml_edit::DocumentMut,
+) {
+    if let Some(dist) = workspace_toml
+        .get("workspace")
+        .and_then(|t| t.get("metadata"))
+        .and_then(|t| t.get("dist"))
+    {
+        new_workspace.insert("dist", dist.to_owned());
+    }
+}
+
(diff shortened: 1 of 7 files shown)


duckinator:
Once #1612 is merged, I'll rebase this off `main` and those changes won't be in this PR's diff.

cargo-dist/src/init.rs
```diff
@@ -52,6 +52,92 @@ fn theme() -> dialoguer::theme::ColorfulTheme {
...
+
+    // Load in the root workspace toml to edit and write back
+    let workspace_toml = config::load_toml(&root_workspace.manifest_path)?;
+    let mut original_workspace_toml = workspace_toml.clone();
+
+    // Generate a new workspace, then populate it using config from Cargo.toml.
+    let mut new_workspace = new_workspace();
+    copy_cargo_workspace_metadata_dist(&mut new_workspace, workspace_toml);
+
+    let workspace_toml = new_workspace;
```
mistydemeo, line 117: This assign probably isn't doing anything at this point?

mistydemeo (approved):
Looking good. Tested that `dist init` on a new, uninitted workspace works fine, and that `dist migrate` also works on an initted workspace with `Cargo.toml`. Likewise, `dist init` with or without the migration is doing the right thing.

cargo-dist/src/init.rs
duckinator: ah, good catch! i think it's a leftover from refactoring work. i can just use `new_workspace` instead of `workspace_toml` when writing the config file a few lines down.

PR #1612, opened 2024-12-06
Rename save_cargo_toml()->write_toml() and load_cargo_toml()->load_toml().

duckinator:
Neither of them are Cargo.toml-specific in any way.

### DIFF
--- a/cargo-dist/src/config/mod.rs
+++ b/cargo-dist/src/config/mod.rs
@@ -990,15 +990,15 @@ pub fn get_project() -> Result<axoproject::WorkspaceGraph, axoproject::errors::P
     Ok(workspaces)
 }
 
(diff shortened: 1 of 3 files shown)


cargo-dist/src/config/mod.rs
```diff
@@ -991,14 +991,14 @@ pub fn get_project() -> Result<axoproject::WorkspaceGraph, axoproject::errors::P
 }
 
 /// Load a Cargo.toml into toml-edit form
```
mistydemeo, line 993: Can you update the comments too?

duckinator: Done!

mistydemeo (approved):

Issue #1608, opened 2024-12-06
Try adding support for combining cargo auditable with cargo-zigbuild/xwin


mistydemeo:
We currently error out on trying the combination of `cargo auditable` with `cargo zigbuild`/`cargo xwin`, because they're all subcommands. However, according to https://github.com/rust-secure-code/cargo-auditable/issues/167, it looks like this combo might actually be possible so long as `auditable` comes first in the argument list. I'll test this and see if I can confirm.

mistydemeo:
This currently doesn't work for `cargo-zigbuild`. The linker flag auditable tries to force isn't compatible with its linker.

```
error: linking with `/github/home/.cache/cargo-zigbuild/0.19.5/zigcc-aarch64-unknown-linux-gnu-aef7.sh` failed: exit status: 1
  |
[...]

PR #1615, opened 2024-12-09
feat: enable auditable with xwin

mistydemeo:
In testing, I've been able to confirm that cargo-auditable and cargo-xwin are compatible; zigbuild is the only one with a compatibility issue.

Refs #1608.

### DIFF
--- a/cargo-dist/src/build/cargo.rs
+++ b/cargo-dist/src/build/cargo.rs
@@ -185,18 +185,18 @@ pub fn make_build_cargo_target_command(
     let mut command = Cmd::new(cargo_cmd, "build your app with Cargo");
     if auditable {
         command.arg("auditable");
-        if wrapper.is_some() {
-            return Err(DistError::CannotDoCargoAuditableAndCrossCompile {
-                host: host.to_owned(),
-                target,
-            });
-        }
     }
     match wrapper {
         None => {
             command.arg("build");
         }
         Some(CargoBuildWrapper::ZigBuild) => {
+            if auditable {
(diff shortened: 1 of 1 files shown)


mistydemeo:
Did a test and can confirm that this had the desired effect. xwin builds worked fine, while zigbuild builds gave the warning telling users to disable one or other of the features.

```
  × Cross-compilation builds from x86_64-unknown-linux-gnu to aarch64-unknown-linux-gnu cannot be used with cargo-
  │ auditable
  help: set cargo-auditable to false or don't do cross-compilation
```

duckinator (approved):
Code is straightforward enough. I'm not sure how to test this particular thing myself, but you said in https://github.com/axodotdev/cargo-dist/pull/1615#issuecomment-2529970503 that you tested it and it behaved as expected, so this should be good. :+1:

mistydemeo:
Ah yes, should have mentioned what I did to test!

I enabled `cargo-auditable = true`, then added this cross-compile config to my test app:

```toml
[dist.github-custom-runners.aarch64-unknown-linux-gnu.container]
image = "quay.io/pypa/manylinux_2_28_x86_64"
host = "x86_64-unknown-linux-musl"

[dist.github-custom-runners.aarch64-pc-windows-msvc]
container = "messense/cargo-xwin"
```

That sets up two cross-commpiles: zigbuild to aarch64-unknown-linux-gnu, and xwin to aarch64-pc-windows-msvc. I then confirmed that xwin+auditable works as expected, while zigbuild bails out with the error about not being able to use auditable with that specific cross-compile.

PR #1617, opened 2024-12-09
fix: provide default for install updater question

### LINKED ISSUE #1610: `dist init` install-updater question does not respect `--yes`
image

I confirmed this still happens with v0.26.0-prerelease.3.

mistydemeo:
I picked `false` as the default when `--yes` is passed, since that's also the default for keymashing enter.

Best viewed with `?w=1`.

Fixes #1610.


duckinator (approved):
This looks good to me. :3

PR #1629, opened 2024-12-11
fix: omnibor ids are local to tarballs

mistydemeo:
Previously, we were creating these as global artifacts and we created them for a larger group of artifacts. Making them global ended up accidentally skipping them for the executable archives, which is the one category of artifact we actually *did* want to be building them for. This removes most of the other categories, and swaps them to local artifacts so they build alongside their tarballs.

Best viewed with `?w=1`.


ashleygwilliams (approved):
this looks good to me- i am curious- what do we do if omnibor cannot install on a target that is specified for a local build? i am assuming right now that it's a "run time" error, aka the local build will fail on release?

mistydemeo:
That's right, we'd fail at the time we tried to install the tool.

PR #1640, opened 2024-12-13
[WIP] Bail early if required tools can't be found.

duckinator:


duckinator:
the output formatting needs to be improved, but i'm pretty happy with this:

image


```
~/test/sometest main+ dist build

  × The following tools are required to run this task, but are missing:
  │ [
  │     "cargo-auditable",
  │     "cargo-cyclonedx",
  │     "omnibor-cli",
  │ ]
  help: Please install the tools mentioned above and try again.

~/test/sometest main+ dist build --target x86_64-pc-windows-msvc

  × The following tools are required to run this task, but are missing:
  │ [
  │     "cargo-auditable",
  │     "cargo-cyclonedx",
  │     "omnibor-cli",
  │     "cargo-xwin",
  │ ]
  help: Please install the tools mentioned above and try again.

~/test/sometest main+
```

duckinator:
As an aside, this change improves the error messages if you run `cargo test` on dist, without `cargo-auditable` + `cargo-cyclonedx` + `omnibor-cli` installed.

Before:

```
Error: 
  × "cargo-dist dist build -aglobal" failed (exit status: 255)
  ╰─▶ 
      stdout:
      
      stderr:
      building artifacts:
        source.tar.gz
        source.tar.gz.sha256
        source.tar.gz.omnibor
        axolotlsay-installer.sh
        axolotlsay-installer.ps1
        axolotlsay.rb
        axolotlsay-npm-package.tar.gz
        axolotlsay.cdx.xml
        sha256.sum
      
      
        × omnibor-cli, required to run this task, is missing
        help: Ensure omnibor-cli is installed
```

After:

```
Error: 
  × "cargo-dist dist build -aglobal" failed (exit status: 255)
  ╰─▶ 
      stdout:
      
      stderr:
      
        × The following tools are required to run this task, but are missing:
        │ [
        │     "cargo-auditable",
        │     "cargo-cyclonedx",
        │     "omnibor-cli",
        │ ]
        help: Please install the tools mentioned above and try again.
```

cargo-dist/src/lib.rs
```diff
@@ -57,8 +57,90 @@ pub mod tasks;
...
+            _ => {}
+        }
+    }
+
+    // These are all of the tools we can check for.
+    //
+    // bool::then(f) returns an Option, so we start with a
+    // Vec<Option<Result<&Tool, DistResult>>>.
+    let all_tools: Vec<Option<DistResult<&Tool>>> = vec![
+        need_cargo_auditable.then(|| tools.cargo_auditable()),
```
mistydemeo, line 111: We should try to scope these to whether we're building the phase(s) that requires them. eg if we're only doing global builds, we shouldn't fail on `auditable` being missing.

cargo-dist/src/tasks.rs
```diff
@@ -3180,6 +3216,29 @@ fn tool_info() -> DistResult<Tools> {
...
+        // but it does confirm it's installed, which is what we care about.
+        cargo_auditable: find_tool2("cargo", "auditable", "--version"),
+
+        cargo_cyclonedx: find_tool2("cargo", "cyclonedx", "--version"),
+        cargo_xwin: find_tool2("cargo", "xwin", "--version"),
+        cargo_zigbuild: find_tool("cargo-zigbuild", "--version"),
+    })
+}
+
+fn find_tool2(name: &str, arg: &str, test_flag: &str) -> Option<Tool> {
```
mistydemeo, line 3230: Can we get a better description of what this does and why it's a different method from the other one?

mistydemeo (commented):
Looks like this has tripped up the tests - let's take a look and make sure we're not introducing tool dep bounds on things that don't actually need them.

cargo-dist/src/tasks.rs
duckinator: Renamed it to `find_cargo_subcommand`, since that's what it's for.

cargo-dist/src/lib.rs
```diff
@@ -57,8 +57,90 @@ pub mod tasks;
...
+
+    let tools = dist.tools;
+
+    let cargo = tools.cargo()?;
+    let host = cargo.host_target.parse()?;
+
+    for step in dist
+        .local_build_steps
+        .iter()
+        .chain(dist.global_build_steps.iter())
```
mistydemeo, line 80: I don't think `BuildStep::Cargo` can ever appear in `dist.global_build_steps` - should be safe to just iterate on local build steps here.

duckinator: Good catch! At one point I wondered if this was the case, but I forgot to follow up on it.

cargo-dist/src/lib.rs
```diff
@@ -57,8 +57,90 @@ pub mod tasks;
...
+
+    let need_cargo_auditable = builds.cargo.cargo_auditable;
+    let need_cargo_cyclonedx = builds.cargo.cargo_cyclonedx;
+    let need_omnibor = builds.omnibor;
+    let mut need_xwin = false;
+    let mut need_zigbuild = false;
+
+    let tools = dist.tools;
+
+    let cargo = tools.cargo()?;
```
mistydemeo, line 74: We can't return an error here - we might be building non-Cargo software, and it's valid for users not to have Cargo if they're not building any Rust software.

Instead, we could consider adding cargo to the list of needed tools below, and checking it if we have any `BuildStep::Cargo` jobs. (Can still call `let cargo =` here, but defer checking if we got `Ok` or `Err` until we're inside `BuildStep::Cargo`.)

cargo-dist/src/lib.rs
duckinator: I made `needs_cargo_auditable` and `needs_cargo_cyclonedx` only true if building local artifacts, which should address this.

cargo-dist/src/lib.rs
```diff
@@ -57,8 +57,88 @@ pub mod tasks;
 #[cfg(test)]
 mod tests;
 
+/// dist env test -- make sure we have everything we need for a build.
+pub fn do_env_test(cfg: &Config) -> DistResult<()> {
+    let (dist, _manifest) = tasks::gather_work(cfg)?;
+
+    let local_builds = cfg.artifact_mode == ArtifactMode::Local;
```
mistydemeo, line 64: We've got two more artifact modes to check for too - `ArtifactMode::Host` (build whatever the current host can build, including local and global builds) and `ArtifactMode::All` (build everything).

cargo-dist/src/lib.rs
duckinator: I couldn't find a way to hoist `let cargo = tools.cargo()` out of the loop, but keep the check in the loop, without either ownership problems or adding more code than felt reasonable.

For now, I just moved both the `let cargo = ...` and `let host = ...` lines into the loop. If you want, I can look into having it check for Cargo tomorrow.

cargo-dist/src/lib.rs
duckinator: Changed to checking for global builds and changed `&& local_builds` to `&& !global_builds`, since that seems like what's actually wanted here.

cargo-dist/src/lib.rs
```diff
@@ -57,8 +57,88 @@ pub mod tasks;
...
+
+    for step in dist.local_build_steps.iter() {
+        // Can't require cross-compilation tools if we aren't compiling.
+        if cfg.artifact_mode == ArtifactMode::Lies {
+            break;
+        }
+
+        match step {
+            BuildStep::Cargo(step) => {
+                let cargo = tools.cargo()?;
```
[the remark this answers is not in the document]
duckinator: I switched to `tools.host_target`. As a note: the code still uses question mark operators in the form of `tools.host_target.parse()?` and `step.target_triple.parse()?`, but I believe if it can't parse the host _or_ build target, it makes sense to just throw an error at that point? Those don't seem recoverable.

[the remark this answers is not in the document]
duckinator: _As best I can tell_, the second of those should be guaranteed to succeed here: if `step.target_triple.parse()?` _would_ fail, it fails in `tasks::gather_work(cfg)?` at the start of the function anyway.

Also it requires specifying an invalid target-triple in `dist-workspace.toml` and running `dist build --target aforementioned-invalid-target`, so I think a hard error makes sense.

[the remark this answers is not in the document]
mistydemeo: Yup, I think that's fine - both of those are genuine error cases. "Cargo is missing" comes up in supported usecases, so that was more my concern.

mistydemeo (approved):
Looking good. It might be nice to handle Cargo itself at some point, for users building non-Rust software, but this all looks great.

duckinator:
Turns out Cargo is handled separately, and `tasks::gather_work()` can't succeed unless we either _don't need_ or _need and have_ Cargo, so checking for it in `do_env_test` doesn't make much sense.


As a note for the future, in case that changes, the approach I found that would theoretically work (if the code ever got run) is adding `Tools::cargo_tool() -> DistResult<&Tool>` alongside `Tools::cargo() -> DistResult<&CargoInfo>`, then in `do_env_test()` you can treat it as a normal tool but use `.cargo_tool()` instead of `.cargo()`.

<details>
<summary>diff</summary>

```diff
diff --git a/cargo-dist/src/lib.rs b/cargo-dist/src/lib.rs
index b02525e7..7c1906ed 100644
--- a/cargo-dist/src/lib.rs
+++ b/cargo-dist/src/lib.rs
@@ -70,6 +70,7 @@ pub fn do_env_test(cfg: &Config) -> DistResult<()> {
     let need_omnibor = builds.omnibor;
     let mut need_xwin = false;
     let mut need_zigbuild = false;
+    let mut need_cargo = false;
 
     let tools = dist.tools;
     let host = tools.host_target.parse()?;
@@ -82,6 +83,8 @@ pub fn do_env_test(cfg: &Config) -> DistResult<()> {
 
         match step {
             BuildStep::Cargo(step) => {
+                need_cargo = true;
+
                 let target = step.target_triple.parse()?;
                 let wrapper = tasks::build_wrapper_for_cross(&host, &target)?;
 
@@ -104,6 +107,7 @@ pub fn do_env_test(cfg: &Config) -> DistResult<()> {
     // bool::then(f) returns an Option, so we start with a
     // Vec<Option<Result<&Tool, DistResult>>>.
     let all_tools: Vec<Option<DistResult<&Tool>>> = vec![
+        need_cargo.then(|| tools.cargo_tool()),
         need_cargo_auditable.then(|| tools.cargo_auditable()),
         need_cargo_cyclonedx.then(|| tools.cargo_cyclonedx()),
         need_omnibor.then(|| tools.omnibor()),
diff --git a/cargo-dist/src/tasks.rs b/cargo-dist/src/tasks.rs
index 4c83ec88..2dfad37e 100644
--- a/cargo-dist/src/tasks.rs
+++ b/cargo-dist/src/tasks.rs
@@ -289,9 +289,20 @@ pub struct Tools {
     pub cargo_xwin: Option<Tool>,
     /// cargo-zigbuild, for some cross builds
     pub cargo_zigbuild: Option<Tool>,
+
+    // This is a kludge to avoid type issues. See cargo_tool() docs.
+    cargo_tool: Option<Tool>,
 }
 
 impl Tools {
+    /// Return the cargo info, as a Tool instead of a CargoInfo, or an error.
+    /// This is a kludge to avoid needing to handle type issues.
+    pub fn cargo_tool(&self) -> DistResult<&Tool> {
+        self.cargo_tool.as_ref().ok_or(DistError::ToolMissing {
+            tool: "cargo".to_owned(),
+        })
+    }
+
     /// Returns the cargo info or an error
     pub fn cargo(&self) -> DistResult<&CargoInfo> {
         self.cargo.as_ref().ok_or(DistError::ToolMissing {
@@ -3224,6 +3235,8 @@ fn tool_info() -> DistResult<Tools> {
         cargo_cyclonedx: find_cargo_subcommand("cargo", "cyclonedx", "--version"),
         cargo_xwin: find_cargo_subcommand("cargo", "xwin", "--version"),
         cargo_zigbuild: find_tool("cargo-zigbuild", "--version"),
+
+        cargo_tool: find_tool("cargo", "-V"),
     })
 }
```

</details>

PR #1656, opened 2024-12-18
feat: migrate dist.toml to dist-workspace.toml

mistydemeo:
We treat dist.toml-alone generic projects a bit awkwardly because of our migration to the new metadata structure. The easiest solution for us is to migrate those proejcts to dist-workspace.toml, where they can be structured and interpreted the same as Cargo.toml projects with a dist-workspace.toml file.

~~Currently, this produces a pair of `dist-workspace.toml` and `dist.toml`. The `dist.toml` retains the package-level metadata, while the `dist-workspace.toml` contains a workspace entry with a single `dist:.` member and the `[dist]` metadata.~~

At @ashleygwilliams's suggestion, I've updated axoproject so it now supports generic projects whose `dist-workspace.toml` contains a `[workspace]` and `[package]` in the same file, and we produce that single metadata when migrating solo-`dist.toml`.

### DIFF
--- a/axoproject/src/generic.rs
+++ b/axoproject/src/generic.rs
@@ -165,18 +165,17 @@ fn workspace_from(manifest_path: &Utf8Path) -> Result<WorkspaceStructure> {
     let source = SourceFile::new(manifest_path.as_str(), String::new());
     let span = source.span_for_line_col(1, 1);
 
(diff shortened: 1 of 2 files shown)


duckinator (approved):
Tried this on a little `dist.toml`-only C project and it migrated it perfectly. Very nice. :3

PR #1658, opened 2024-12-19
Rework `Add-Path` to avoid expanding and request an environment refresh

### LINKED ISSUE #1614: Windows: Modify PATH without requiring a system restart?

DavisVaughan:
Closes #1614
Closes #1657

A rework of #1657 after everything we have learned from https://github.com/axodotdev/cargo-dist/pull/1657#issuecomment-2555684533. Please read that comment / PR as it describes everything in detail.

I will once again require some help with snapshot updating!

Key points:

- Powershell installers no longer forcibly expand "expandable variables" in your `PATH`, like `%EXPAND%`
- Powershell installers no longer change your user-level `Path` registry property from `REG_EXPAND_SZ` to `REG_SZ`
- Powershell installers no longer require a _system_ restart, and now only require a _shell_ restart, like Unix shell installers

I tested this myself by:
- Completely resetting my `Path` to a state before dist touched it
- Adding `%TEMP%` into my `Path` as a test "expandable" variable 
- Running a patched powershell installer with these changes, resulting in:
    - The `Path` stayed a `REG_EXPAND_SZ` data type
    - The `Path` retained my unexpanded `%TEMP%`
    - On a _shell_ restart, I could load the cli tool I installed
- Running the patched powershell installer _again_, nothing happened because it detected the existing install directory in the `Path`

I even went so far as to wipe out my `Path` property in the registry to see if it could create it "from scratch" if it wasn't there already, and it worked and was created with the right `REG_EXPAND_SZ` type.

Screenshot 2024-12-19 at 4 34 21 PM

### DIFF
--- a/cargo-dist/templates/installer/installer.ps1.j2
+++ b/cargo-dist/templates/installer/installer.ps1.j2
@@ -469,7 +469,7 @@ function Invoke-Installer($artifacts, $platforms) {
     Add-Ci-Path $dest_dir
     if (Add-Path $dest_dir) {
         Write-Information ""
-        Write-Information "To add $dest_dir to your PATH, either restart your system or run:"
+        Write-Information "To add $dest_dir to your PATH, either restart your shell or run:"
         Write-Information ""
         Write-Information "    set Path=$dest_dir;%Path%   (cmd)"
         Write-Information "    `$env:Path = `"$dest_dir;`$env:Path`"   (powershell)"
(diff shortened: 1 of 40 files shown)


duckinator:
FYI: We're probably not going to get a chance to review this until after the holidays.

Thank you again for the help, by the way! I know lots of folks are looking forward to not having to reboot after installing stuff, and looks like this PR improves other Path-related stuff too. :slightly_smiling_face:

mistydemeo:
Thanks again for the comprehensive writeup! I've updated the snapshots, and will do some testing before we merge this.

mistydemeo (approved):
Tested, and confirmed working! Thanks so much for your work on this. I was also able to confirm that this preserves the expected string type for the PATH in the registry.

I did notice that if the user had an existing PATH entry that's a `REG_SZ`, this will replace it with a `REG_EXPAND_SZ`. Since we were previously unconditionally creating it as `REG_SZ`, this is IMO still a step up from before and not a blocker.

PR #1659, opened 2024-12-19
feat: check for shadowed binaries in shell installer

mistydemeo:
This was a user request. In some cases, the user may have multiple copies of something installed, and won't realize the copy installed by the shell installer has been shadowed in their `PATH`. It's useful for us to test that and provide an explicit warning.

Tested and working:

```
$ ./axolotlsay-installer.sh
downloading axolotlsay 0.2.320 aarch64-apple-darwin
installing to /Users/mistydemeo/.cargo/bin
  axolotlsay
  axolotlsay-update
everything's installed!
WARNING: The following commands are shadowed by other commands in your PATH: axolotlsay
```


duckinator (approved):
Tested locally, and it works well.

PR #1661, opened 2024-12-20
fix: Avoid 'dist migrate' deleting dist-workspace.toml.

duckinator:
A condition like the following was intended to be switched from `==` to `!=`:

    A == B && C == D

The correct way to do this would be:

    A != B || C != D

However, instead, what we had was this inverted logic:

    A != B && C != D

In practice, because of the specific things we were checking, this meant instead of *never* deleting dist-workspace.toml, we were *always* deleting it.


duckinator:
Fortunately, a duplicate check in `dist init` prevent that from having the same bug.

mistydemeo (approved):
Good catch.

PR #1662, opened 2024-12-20
Update changelog for 0.27.1

duckinator:
This is waiting on #1661.


mistydemeo (approved):
We can update the date when the release is cut.

PR #1684, opened 2025-01-08
release: 1.0.0-rc.1

mistydemeo:
This release contains the rename but no other content. Any remaining 1.0 content will come after the first prerelease.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,7 +183,7 @@ dependencies = [
 
 [[package]]
 name = "axoproject"
-version = "0.28.0"
+version = "1.0.0-rc.1"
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
q- can we call this a release candidate? e.g. rc.1?

mistydemeo:
I *think* so. Let me double check our prerelease handling to make sure we'll correctly handle it like with the other naming scheme.

mistydemeo:
Yes, it should be fine. We'll handle `.rc1` as a prerelease. I'll swap it over.

ashleygwilliams (approved):

PR #1695, opened 2025-01-13
docs: Fix incorrect option name

ThiefMaster:
The option is actually called `install-success-msg`; using `custom-success-msg` does not work.

Unrelated thing I noticed while testing this option: Backticks inside that message actually behave as shell backticks. Not sure if intentional or not (can't think of any reason why someone would want this). Maybe that deserves a warning so people are not surprised by it, or to be fixed so the message can contain anything w/o being interpreted by the shell?


ashleygwilliams (approved):

ashleygwilliams:
thank you so much for this and apologies for the misleading docs/confusion!

PR #1698, opened 2025-01-14
(config-migration) Implement `Display` for `ChecksumStyle`, derive `Default` for `ArtifactLayer` and `InstallerLayer`

duckinator:
This is needed for implementing the v0->v1 config migration.


ashleygwilliams (approved):
this looks good assuming we are happy with all Nones as the defaults?

duckinator:
Yeah, all `None` as defaults is what we want here.

PR #1699, opened 2025-01-14
(config-migration) Add the ability to determine and track the config version.

duckinator:


ashleygwilliams:
this looks like it has a failing test, not sure if you want to address that before i review or not

duckinator:
I'll look into that first. I _think_ it was a transient error from pnpm (ah, the joys of tests that require the network to function), but I'm re-running to verify that.

cargo-dist/src/config/version.rs
```diff
@@ -0,0 +1,67 @@
+//! For determining which configuration version we're using.
```
ashleygwilliams, line 1: A NEW FILE??!!! (yessssssss)

cargo-dist/src/config/version.rs
```diff
@@ -0,0 +1,67 @@
...
+    let root_workspace = workspaces.root_workspace();
+
+    get_version_for_manifest(root_workspace.manifest_path.to_owned())
+}
+
+/// Given a path to a dist manifest (e.g. `dist-workspace.toml`), returns
+/// the config version being used.
+pub fn get_version_for_manifest(dist_manifest_path: Utf8PathBuf) -> DistResult<i64> {
+    if dist_manifest_path.file_name() != Some("dist-workspace.toml") {
+        // If the manifest is in Cargo.toml or dist.toml, we're
```
ashleygwilliams, line 44: this may not be strictly true because there was a brief period of time when folks put all config in a dist.toml (including workspace). i think practically tho, this is true, but wanted to call out the formal exception.

cargo-dist/src/config/version.rs
```diff
@@ -0,0 +1,67 @@
+//! For determining which configuration version we're using.
+
+use axoasset::SourceFile;
+use camino::Utf8PathBuf;
+use serde::Deserialize;
+
+use crate::{DistError, DistResult};
+
+const MAX_VERSION: i64 = 1;
```
ashleygwilliams, line 9: thought- instead of a const i64... do we want to make this a semver version and/or a enum? the validity check with "greater than" feels non specific and picking a different type for this value feels like the right call. personally i am leaning towards just using semver

ashleygwilliams (commented):
in general i think this is good, one thing to decide is the type of the version field (not sure i64 is the right call)

cargo-dist/src/config/mod.rs
```diff
@@ -18,6 +18,10 @@ use crate::{
     METADATA_DIST,
 };
 
+mod version;
+pub use version::ConfigVersion;
+pub use version::get_version;
```
ashleygwilliams, line 23: could collapse this into the above i think?

ashleygwilliams (approved):
lgtm modulo small use statement nit

cargo-dist/src/config/mod.rs
duckinator: Fixing that also fixed the failed `rust fmt` in CI, conveniently.

PR #1703, opened 2025-01-16
(config-migration) Add `validate_install_paths()` and `merge_workspace_config()` to `TomlLayer`

duckinator:


cargo-dist/src/config/v1/mod.rs
```diff
@@ -481,6 +483,173 @@ impl TomlLayer {
             }
         }
     }
+
+    /// Determines whether the configured install paths are compatible with each other
+    pub fn validate_install_paths(&self) -> DistResult<()> {
```
ashleygwilliams, line 488: do we want this to return a vec of install paths maybe? (this is something people often want visibility into in the plan step)- this may not be the place for it, but seeing the unit type return made me think of it

cargo-dist/src/config/v1/mod.rs
```diff
@@ -481,6 +483,173 @@ impl TomlLayer {
...
+                if paths.len() > 1 && paths.contains(&InstallPathStrategy::CargoHome) {
+                    return Err(DistError::IncompatibleInstallPathConfiguration {});
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    // We say "value is being ignored", but it seems like we use the "ignored" values anyway?
```
ashleygwilliams, line 500: can you give more details on how it is used? that def seems bad!! but agree we should repro for now, might leave a TODO in this comment to return to this

cargo-dist/src/config/v1/mod.rs
```diff
@@ -481,6 +483,173 @@ impl TomlLayer {
...
+        warn!("dist.{} is set, but this is only accepted in workspace.metadata (value is being ignored): {}", name, package_manifest_path);
+    }
+
+    /// Merge a workspace config into a package config (self)
+    pub fn merge_workspace_config(
+        &mut self,
+        workspace_config: &Self,
+        package_manifest_path: &Utf8Path,
+    ) {
+        // This is intentionally written awkwardly to make you update it
```
ashleygwilliams, line 513: do we think this "awkwardness" strategy is still useful? (may still be worth not changing atm but i am curious your thoughts)

cargo-dist/src/config/v1/mod.rs
```diff
@@ -481,6 +483,173 @@ impl TomlLayer {
...
+
+        // Check for global settings on local packages
+        if dist_version.is_some() {
+            Self::merge_warn("dist-version", package_manifest_path);
+        }
+        if dist_url_override.is_some() {
+            Self::merge_warn("dist-url-override", package_manifest_path);
+        }
+
+        // Arguably should be package-local for things like msi installers, but doesn't make sense for CI,
```
ashleygwilliams, line 537: can you say more about how this is helpful for msi installers? agree on maybe not changing this yet- honestly allow dirty is a messy feature haha so i'd rather see it phased out in favor of heavy handed "custom" features that can inline needed changes, etc

cargo-dist/src/config/v1/mod.rs
```diff
@@ -481,6 +483,173 @@ impl TomlLayer {
...
+        // All of `hosts.github` is global-only.
+        if let Some(hosts) = &hosts {
+            if let Some(github) = &hosts.github {
+                if github.is_not_false() {
+                    Self::merge_warn("hosts.github", package_manifest_path);
+                }
+            }
+        }
+
+        // All of `installers` (InstallerLayer) is package-only, so no need to check it.
```
ashleygwilliams, line 594: i wish this was in the type system somehow

ashleygwilliams (approved):
a few comments, i think in general this is fine- probably some ways to clean up/optimize but not necessarily urgent atm

cargo-dist/src/config/v1/mod.rs
duckinator: This comment was copied verbatim from `v0.rs` and, in retrospect, I have no idea what it's trying to communicate.

Leaning towards just removing it.

cargo-dist/src/config/v1/mod.rs
duckinator: I think there's probably a better way to do it, but I'm not sure what that is. But it _has_ actively caught problems for me a few times, so I think until we find a proper alternative, it's better than not destructuring it.

cargo-dist/src/config/v1/mod.rs
duckinator: I don't think it makes sense. If we add a function like `get_install_paths(&self) -> Vec<...>`, we could use `if let Some(paths) = self.get_install_paths()`, but I'm not sure that's worth the effort for this PR.

PR #1704, opened 2025-01-16
(config-migration) Don't modify `[profile.dist]` until after migrations; consolidate migrations under one prompt.

duckinator:


ashleygwilliams (approved):
looks good/uncontroversial to me!

PR #1705, opened 2025-01-16
(config-migration) Make `dist::init` and `dist::tasks` into folder-based modules

duckinator:


duckinator:
This conflicts with #1704 and I really should've realized that sooner, so setting this aside until after that's reviewed + merged.

ashleygwilliams (approved):
YESS

PR #1717, opened 2025-01-21
Move src/init/{mod -> v0}.rs, add new src/init/mod.rs to re-export functions

duckinator:
Also have src/tests/config.rs use dist::init::v0 directly.


duckinator:
The diff looks messier than this actually is. If you go through one commit at a time, it's a bit clearer:

- rename `src/init/mod.rs` -> `src/init/v0.rs`
- add new `src/init/mod.rs` which re-exports some stuff from `dist::init::v0`
- update a reference in `src/tests/config.rs`

ashleygwilliams (approved):

PR #1718, opened 2025-01-21
(config-migration) Split `dist::migrate` out from `dist::init`

duckinator:


cargo-dist/src/init/v0.rs
```diff
@@ -258,7 +99,7 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
             .interact()?;
 
         if is_migrating {
-            do_migrate()?;
+            migrate::do_migrate()?;
```
ashleygwilliams, line 96: this makes me think we could have `migrate::do` but this type of renaming isn't super important atm

cargo-dist/src/migrate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+fn prune_cargo_workspace_metadata_dist(workspace: &mut toml_edit::DocumentMut) {
+    workspace
+        .get_mut("workspace")
+        .and_then(|ws| ws.get_mut("metadata"))
+        .and_then(|metadata_item| metadata_item.as_table_mut())
+        .and_then(|table| table.remove("dist"));
+}
+
+/// Create a toml-edit document set up for a cargo workspace.
+pub(crate) fn new_cargo_workspace() -> toml_edit::DocumentMut {
```
ashleygwilliams, line 45: does this want a version attached to it?

ashleygwilliams: e.g. `new_v1_cargo_workspace`

cargo-dist/src/migrate/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+        array.push("cargo:.");
+        t["members"] = toml_edit::value(array);
+    }
+    new_workspace.insert("workspace", table);
+
+    new_workspace
+}
+
+/// Create a toml-edit document set up for a cargo workspace.
+fn new_generic_workspace() -> toml_edit::DocumentMut {
```
ashleygwilliams, line 61: same as above

ashleygwilliams (approved):
main comments are around naming, i do think in the migrate mod it would be useful to talk about versions but i won't block if you disagree

cargo-dist/src/migrate/mod.rs
ashleygwilliams: actually just realized that maybe migrate just wants a version namespace?

duckinator:
I think I'm going to leave the naming as is for now, since other stuff builds on top of this.

PR #1719, opened 2025-01-21
(config-migration) Move root_workspace + workspace_toml to where they're used; remove unneeded desired_workspace_kind

duckinator:


cargo-dist/src/init/v0.rs
```diff
@@ -148,38 +130,26 @@ pub fn do_init(cfg: &Config, args: &InitArgs) -> DistResult<()> {
...
-        filename = "dist-workspace.toml";
-        destination = root_workspace.workspace_dir.join(filename);
-    } else {
-        filename = root_workspace
-            .manifest_path
-            .file_name()
-            .expect("no filename!?");
-        destination = root_workspace.manifest_path.to_owned();
-    };
+    let filename = root_workspace
```
ashleygwilliams, line 144: ahhh so much nicer

duckinator: as a fun aside, this has actually been unneeded since we started having `dist init` call `dist migrate`. i just wasn't as confident of it until i was more familiar with this part of the codebase.

PR #1723, opened 2025-01-23
(config-migration) Add dist::config::load() and friends

duckinator:


cargo-dist/src/errors.rs
```diff
@@ -626,6 +626,10 @@ pub enum DistError {
         /// Version the project uses
         your_version: semver::Version,
     },
+
+    /// No dist manifest
+    #[error("No configuration file (e.g. dist-workspace.toml) was found")]
```
ashleygwilliams, line 631: we may want to give some more help here- could people on v0 who havent yet migrated get this?

ashleygwilliams (approved):
one comment on improving an error message otherwise looks great. love the small functions

cargo-dist/src/errors.rs
duckinator: people on v0 can't get this -- it's only returned by `dist::v1::loader::*`, which is only used if we explicitly do not have a v0 config.

duckinator: added help text

PR #1724, opened 2025-01-23
(config-migration) Add `config::version::want_v1()` and empty `do_migrate_from_v0()`

duckinator:


cargo-dist/src/config/mod.rs
```diff
@@ -18,8 +18,8 @@ use crate::{
     METADATA_DIST,
 };
 
-mod version;
-pub use version::{get_version, ConfigVersion};
+pub mod version;
```
ashleygwilliams, line 21: we want to pub the mod and the stuff within it? that seems odd to me

ashleygwilliams (approved):
question on visibility, otherwise looks good

cargo-dist/src/config/mod.rs
duckinator: good catch -- made some minor changes so the mod can be private.

PR #1726, opened 2025-01-23
(config-migration) Split `init_dist_profile()` out into its own file

duckinator:


cargo-dist/src/init/init_dist_profile.rs
```diff
@@ -0,0 +1,45 @@
+use crate::{config::Config, errors::DistResult, PROFILE_DIST};
+use axoasset::toml_edit;
+
+pub fn init_dist_profile(
```
ashleygwilliams, line 4: since this is in a folder called init do we want to drop the init here?

duckinator: For things yoinked from `init/v0.rs` (previously `init.rs`), I'm trying to avoid renaming things right now.

Once I'm done splitting things apart, I'm open to renaming stuff.

duckinator: The file can def be renamed to just `init/dist_profile.rs`, though. I'll go ahead and do that.

ashleygwilliams (approved):

PR #1734, opened 2025-01-29
Update docs related to #1658

DavisVaughan:
Follow up to #1658

The powershell docs were technically wrong before, because they seemed to imply that only a shell restart was required, when really you had to restart your whole machine. Luckily that means that updating the docs to reflect the current state of the world was easy, because now you really do only need the shell restart 😄


duckinator (approved):
Thank you for following up on this!

PR #1735, opened 2025-01-29
fix: Do not report shadowed binaries when no binary exists

geofft:


geofft:
The uv installer is giving me

```
downloading uv 0.5.25 aarch64-unknown-linux-gnu
no checksums to verify
installing to /home/nixos/.local/bin
  uv
  uvx
everything's installed!

To add $HOME/.local/bin to your PATH, either restart your shell or run:

    source $HOME/.local/bin/env (sh, bash, zsh)
    source $HOME/.local/bin/env.fish (fish)
WARNING: The following commands are shadowed by other commands in your PATH: uv uvx
```

If I apply this patch to https://astral.sh/uv/install.sh and run the patched installer (after deleting ~/.local), then I don't get that last message.

duckinator (approved):
Thank you for the PR! This is a good catch. :slightly_smiling_face:

PR #1743, opened 2025-02-04
Support --options in mac code signing

jackkleeman:
This is necessary to produce hardened binaries, which are required if the binaries are to be notarized.


mistydemeo (approved):
This makes sense, thanks! Do you mind rebasing this to let the tests run?

jackkleeman:
Done @mistydemeo

mistydemeo (approved):
Thanks!

PR #1745, opened 2025-02-04
Support publishing prereleases even if there are only custom publish jobs

### LINKED ISSUE #1744: publish-prereleases = true only takes effect if publish jobs contains `npm` or `homebrew`
Currently the logic to determine if publish-prereleases = true makes it into the plan depends on the prerelease=true flag being set for both of the npm and homebrew v1 config blocks. If both of those are None, we default to false. However, its possible to have custom publish jobs which would *also* be skipped on prerelease. I feel that if both are None, we should still respect the value set in the dist config, if its set, before defaulting to false. I think this could be a breaking change, if there are custom publish jobs that users are expecting not to execute on prerelease.

jackkleeman:
Fixes #1744

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -302,7 +302,7 @@ impl GithubCiInfo {
         }
 
         let mut publish_jobs = vec![];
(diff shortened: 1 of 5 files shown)


mistydemeo:
Merged via #1838. Thank you!

PR #1798, opened 2025-03-17
Explicitly configure wix properties not defaulted from metadata

pnehrer:
WIP


mistydemeo:
Sorry for the late review. Makes sense to me - I'll look into adding this to the tests so we can be sure it works.

mistydemeo:
In testing now, we *are* bringing these values in from Cargo.toml. For example, in the axolotlsay test, we're getting the axolotlsay package's description and URL. Is your goal here to override those values with something else?

pnehrer:
Yes, we were hoping to be able to override these explicitly.

PR #1811, opened 2025-04-03
initial fork of cargo-dist to astral-sh org

Gankra:


Gankra:
oops, opened this against the repo. but uh, yeah we need to fork this to keep fixing bugs and cutting releases. we're going off the 0.28.0 branch and not 1.0.0.

PR #1834, opened 2025-07-05
feat: synchronize runners with github

mistydemeo:
Based on current docs:

* https://docs.github.com/en/actions/how-tos/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners
* https://docs.github.com/en/actions/concepts/runners/about-larger-runners

### DIFF
--- a/book/src/ci/customizing.md
+++ b/book/src/ci/customizing.md
@@ -89,8 +89,8 @@ By default, dist uses the following runners:
 
 * Linux (x86_64): `ubuntu-22.04`
 * macOS (x86_64): `macos-13`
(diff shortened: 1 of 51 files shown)


reneleonhardt:
Coincidentally, those are the exact latest versions 😄

Wouldn't it be easier to just use macos-latest and windows-latest?
What are the problems with ubuntu-latest (ubuntu-24.04)?

mistydemeo:
We try to use the oldest supported version for compatibility. On Linux in particular, linking to an older glibc improves compatibility across different distributions and OS versions.

PR #1860, opened 2025-07-10
chore(deps): bump target-lexicon to 0.13.2

hpouillot:
Fix a build error

This error was caused by a version mismatch of the target-lexicon crate in the workspace dependency graph.

The dist-schema crate (and thus cargo-dist) depended on target-lexicon v0.12.16, while other dependencies (such as guppy via axoproject) pulled in target-lexicon v0.13.2.

Rust treats types from different versions as distinct, so the ParseError type from target-lexicon v0.12.16 was not compatible with the one from v0.13.2, leading to trait bound errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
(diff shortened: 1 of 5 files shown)


mistydemeo:
Hm, looks like this introduced dependencies that aren't buildable on the version of Rust we're targeting. I'll have to look into an upgrade.

mistydemeo:
Do you mind specifying where you saw the newer target-lexicon being pulled in? Were you doing an unlocked `cargo install`? It looks like the lockfile specifies the same version for both crates.

mistydemeo:
I pulled your commit into a guppy update branch was able to get it to work after tweaking the lockfile. Thanks!

PR #1948, opened 2025-07-23
feat: add persist-credentials: false to checkouts

mistydemeo:


mistydemeo:
@Dr-Emann - I've applied a fix for the Homebrew publishing issue you noted with the `persist-credentials` change.

Issue #1644, opened 2024-12-17
Cross build `aarch64-pc-windows-msvc` from `x86_64-pc-windows-msvc` without using `cargo-xwin`


Myriad-Dreamin:
First, thanks for the awesome project. It does save me a lot of time to setup and release targets and the generated announcement looks beautiful. However, I had a hard time to build with cargo-dist, I put some feedback here.

**Problem**

My first problem is that `cargo-xwin` is too old to build my binaries:

https://github.com/Myriad-Dreamin/tinymist/actions/runs/12358673816/job/34489776246

Here is the critical log line:

```log
error: rustc 1.79.0 is not supported by the following packages:
  sync-lsp@0.12.12 requires rustc 1.82
```

**My Temporary Solution**

However, I also find that there is no problem to just run `cargo build --target aarch64-pc-windows-msvc` from `x86_64-pc-windows-msvc` for my binaries. I therefore changed `cargo-dist/src/tasks.rs`:

```patch
@@ -486,6 +506,10 @@ pub fn build_wrapper_for_cross(
// compiling for Windows (making PE binaries, .dll files, etc.)
OperatingSystem::Windows => match host.operating_system {
+    OperatingSystem::Windows => {
+        // from win to win, it is okay for my binaries
+        Ok(None)
+    }
  OperatingSystem::Linux | OperatingSystem::Darwin => {
  // cargo-xwin is made for that
}
```

**Proposed Improvement**

It should be great if cargo-dist can provide me some twist to not using `cargo-xwin` if I don't have to. If we can already have a nice way to solve it, please let me know.

Myriad-Dreamin:
Sorry, just to add a note. When I back to read the issue, I may encode some wrong knowledge. The reason that my first shoot uses Windows-to-Windows method was is just because we were doing that in our old CI script. I have no idea about reason we don't use Linux-to-Windows. If you are interested, you could also check Linux-to-Windows, to make it clearer.

Myriad-Dreamin:
Also put some information, the [cargo.lock](https://github.com/Myriad-Dreamin/tinymist/blob/b1bb3196b718da0c3df97020877a343a7b36c488/Cargo.lock). These crates should work pretty, because the software using it has been used on thousands of windows PC. Some heavy crates:
- `reqwest+rustls`: network
- `tokio`: async
- `rayon`: multi-thread functions

messense:
[...]

PR #1962, opened 2025-07-24
fix: allow cross compiling from windows to windows

baszalmstra:
This PR enables cross-compilation from windows to windows. This is needed to compile `aarch64-pc-windows-msvc` targets from another windows machine which is generally supported. This change was discussed in https://github.com/axodotdev/cargo-dist/issues/1644 but never materialized in a PR.

Fixes https://github.com/axodotdev/cargo-dist/issues/1644


mistydemeo:
Makes sense to me! Have you had a chance to test this yet?

baszalmstra:
Yes, works like a charm!

mistydemeo (approved):
Sounds good, thanks! I'll check if the docs need updating. I appreciate the PR!

PR #1980, opened 2025-07-28
use GitHub ARM-specific runners for aarch64 Linux targets by default

domenkozar:
GitHub now provides Linux ARM64 hosted runners for free in public repositories. This change updates the default runner for aarch64-unknown-linux-gnu targets to use ubuntu-22.04-arm instead of the x86_64 runner.

See: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Addresses https://github.com/axodotdev/cargo-dist/issues/74

### DIFF
--- a/cargo-dist/src/backend/ci/github.rs
+++ b/cargo-dist/src/backend/ci/github.rs
@@ -671,7 +671,14 @@ fn github_runner_for_target(
     // where random system dependencies can creep in and be very
     // recent. This helps with portability!
     let result = Some(match target_triple.operating_system {
-        OperatingSystem::Linux => runner_to_config(GithubRunnerRef::from_str("ubuntu-22.04")),
+        OperatingSystem::Linux => {
+            // Use ARM-specific runner for aarch64 Linux targets
+            if matches!(target_triple.architecture, Architecture::Aarch64(_)) {
+                runner_to_config(GithubRunnerRef::from_str("ubuntu-22.04-arm"))
+            } else {
+                runner_to_config(GithubRunnerRef::from_str("ubuntu-22.04"))
+            }
+        }
         OperatingSystem::Darwin(_) => runner_to_config(GithubRunnerRef::from_str("macos-13")),
         OperatingSystem::Windows => {
             // Default to cargo-xwin for Windows cross-compiles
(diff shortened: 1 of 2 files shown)


mistydemeo:
Nice, thank you! I missed that these had been added. I'll have to give this a test.

cargo-dist/src/backend/ci/github.rs
```diff
@@ -671,7 +671,14 @@ fn github_runner_for_target(
     // where random system dependencies can creep in and be very
     // recent. This helps with portability!
     let result = Some(match target_triple.operating_system {
-        OperatingSystem::Linux => runner_to_config(GithubRunnerRef::from_str("ubuntu-22.04")),
+        OperatingSystem::Linux => {
+            // Use ARM-specific runner for aarch64 Linux targets
+            if target_triple.architecture == Architecture::Aarch64 {
```
mistydemeo, line 676: `Architecture::Aarch64` (which [comes from target_lexicon](https://docs.rs/target-lexicon/0.13.2/target_lexicon/enum.Architecture.html)) takes an architecture argument which is necessary here for the `==` comparison to work. Probably easier to do this with a `matches!()` check so you can check against any subarchitecture.
```suggestion
            if matches!(target_triple.architecture, Architecture::Aarch64(_)) {
```

mistydemeo (changes requested):
Noticed one issue flagged when CI ran.

cargo-dist/src/backend/ci/github.rs
domenkozar: Thanks! Rebased and pushed.

mistydemeo:
The test failure here is "expected", in that it's reflecting the change in default runner that's not in the snapshots yet. I can update the snapshots for you, or if you'd like to do it, you can:

* Install [cargo insta](https://insta.rs/docs/cli/)
* Run the tests with `cargo test`
* Run `cargo insta accept`
* Commit and push the changes

domenkozar:
> The test failure here is "expected", in that it's reflecting the change in default runner that's not in the snapshots yet. I can update the snapshots for you, or if you'd like to do it, you can:
> 
> * Install [cargo insta](https://insta.rs/docs/cli/)
> * Run the tests with `cargo test`
> * Run `cargo insta accept`
> * Commit and push the changes

I've pushed the changes, if you want to provide a good experience out the box, you can add `devenv.nix` like:

```nix
{ pkgs, ... }: {
  languages.rust.enable = true;

  packages = [ pkgs.cargo-insta ];
}
```

mistydemeo:
Nice, I'll do that. Would be good to provide a more convenient experience for nix-based devs.

Issue #1754, opened 2025-02-10
Support attestations in GitHub host phase


samypr100:
Feature Request: I would like to have support github attestations during the `host` phase with ability to filter on what the subjects should be. See example working workflow https://github.com/astral-sh/uv/pull/11357

Problem: Even though there's support already for github attestations via `github-attestations = true`, it only works when `build-local-artifacts` is used, which is not the case in the above scenario/workflow. In addition it would be desirable to be able to configure what gets attested and becomes part of the subject versus what does not, as shown in the above workflow.

Proposal:
1. Support `actions/attest-build-provenance@v2` on `host` phase. This could be introduced by adding a `github_attestations_phase` configuration variable that takes either `host` or `build-local-artifacts`, defaulting to `build-local-artifacts` for backwards compatibility and modifying `publish_github.yml.j2` to support it.
2. Support filters for attestations in host phase. This could also be done by adding a `github_attestations_filters` configuration variable that is an array of strings that will be passed down to `subject-path` in `attest-build-provenance` in the `host` phase. This can default to `None` for backwards compatibility. For example, in case of `build-local-artifacts` it should result in the current `target/distrib/*${{ join(matrix.targets, ', ') }}*"`, but when used with `host` it will result in `artifacts/*` by default or when adding more filters `['*.json', '*.sh', '*.ps1', '*.zip', '*.tar.gz']` it would work for example as shown below.

    ```yml
      - name: Generate artifact attestations
        uses: actions/attest-build-provenance@v2
        with:
          subject-path: |
            artifacts/*.json
            artifacts/*.sh
            artifacts/*.ps1
            artifacts/*.zip
            artifacts/*.tar.gz
    ```

I'd appreciate any thoughts/guidance here if this is achievable or something you'd be happy to support from an external contribution 😄

PR #2000, opened 2025-08-02
Add support for attestations in the host phase

samypr100:
Closes https://github.com/axodotdev/cargo-dist/issues/1754

Two integration tests were added to confirm functionality behaves as expected.

Originally opened in https://github.com/astral-sh/cargo-dist/pull/45

### DIFF
--- a/cargo-dist/src/init/v0.rs
+++ b/cargo-dist/src/init/v0.rs
@@ -236,6 +236,8 @@ fn get_new_dist_metadata(
             ssldotcom_windows_sign: None,
             macos_sign: None,
             github_attestations: None,
+            github_attestations_filters: None,
+            github_attestations_phase: None,
             msvc_crt_static: None,
             hosting: None,
             extra_artifacts: None,
@@ -738,6 +740,8 @@ fn apply_dist_to_metadata(metadata: &mut toml_edit::Item, meta: &DistMetadata) {
         ssldotcom_windows_sign,
         macos_sign,
         github_attestations,
+        github_attestations_filters,
+        github_attestations_phase,
         msvc_crt_static,
         hosting,
         tag_namespace,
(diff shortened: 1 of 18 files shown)


mistydemeo:
This makes sense to me, and thank you for adding tests!

Question for you - do you think these make sense as two separate settings? It feels like they're pretty linked, since the `github-attestations-filters` setting doesn't do anything if you don't also change the other one. Maybe it makes sense to just have the `github-attestations-filters` setting, and have that switch the phase if you choose it?

samypr100:
> This makes sense to me, and thank you for adding tests!
> 
> Question for you - do you think these make sense as two separate settings? It feels like they're pretty linked, since the `github-attestations-filters` setting doesn't do anything if you don't also change the other one. Maybe it makes sense to just have the `github-attestations-filters` setting, and have that switch the phase if you choose it?

Good question, happy to change it.
I chose two separate entries to allow more flexibility in the future, e.g. to support filters in the `build-local-artifacts` phase as well.
I didn't implement that in this PR though as I was thinking that could be a follow up.

mistydemeo:
Good point - leaving it open for other expansion could be useful. I was thinking about what else you might want to attest in that step, but now that I think about it, the custom build steps do mean that there are hypothetically extra artifacts produced in local builds that the user might want attested. That being the case, I'm happy for these to be separate steps.

samypr100:
@mistydemeo Anything I could do to help move this forward? 🙏

mistydemeo:
Sorry to have taken so long! I'll try to get this merged tonight.

Just for a little transparency, I'm preparing for a big move in a few weeks, and it turns out dealing with all of that stuff takes up most of my after-work energy and brainpower. 😅

samypr100:
> Sorry to have taken so long! I'll try to get this merged tonight.
> 
> Just for a little transparency, I'm preparing for a big move in a few weeks, and it turns out dealing with all of that stuff takes up most of my after-work energy and brainpower. 😅

No worries, moving is a nightmare. Take your time 📦.
I asked as I wasn't sure there were other changes you'd like for me to do 😄.
On the avenue of total transparency I am also going to be dissapearing 👻 for a couple of weeks sometime next week 😆.

cargo-dist/src/init/v0.rs
```diff
@@ -1120,6 +1124,20 @@ fn apply_dist_to_metadata(metadata: &mut toml_edit::Item, meta: &DistMetadata) {
         *github_attestations,
     );
 
+    apply_string_or_list(
+        table,
+        "github_attestations-filters",
```
mistydemeo, line 1129: Looks like a typo in the name here - using an underscore for the first separator instead of a dash. This would make `init` generate an incorrect config.
```suggestion
        "github-attestations-filters",
```

mistydemeo (changes requested):
Looking good! Sorry it took me so long to review. Just one quick request, then I think we're good to go.

PR #2005, opened 2025-08-03
fix: windows use powershell to extract zip file

### LINKED ISSUE #2004: Npm install does not work on windows
## Description

The npm install script depends on commands like "unzip" which as far as I can see are not provided with windows by default (At least they're not apart of my install)

Image

This results in the following error when trying to install the package on windows:

`npm error Error fetching release: spawnSync unzip ENOENT`

jacobtread:
## Description

For windows platforms the unzip utility is not on systems by default, this changes it so that on windows powershell is used in combination with Expand-Archive to extract the zip file.

## Changes
- Extend "Package" to take in the platform
- Add case in zip extraction for the windows platform to use powershell Expand-Archive instead of unzip

## Related Issues
- Closes #2004

### DIFF
--- a/cargo-dist/tests/snapshots/axolotlsay_attestations_filters.snap
+++ b/cargo-dist/tests/snapshots/axolotlsay_attestations_filters.snap
@@ -2521,7 +2521,7 @@ const error = (msg) => {
 };
 
 class Package {
-  constructor(name, url, filename, zipExt, binaries) {
+  constructor(platform, name, url, filename, zipExt, binaries) {
     let errors = [];
     if (typeof url !== "string") {
       errors.push("url must be a string");
@@ -2555,6 +2555,8 @@ class Package {
         '\n\nCorrect usage: new Package("my-binary", "https://example.com/binary/download.tar.gz", {"my-binary": "my-binary"})';
       error(errorMsg);
     }
+
+    this.platform = platform;
     this.url = url;
     this.name = name;
     this.filename = filename;
@@ -2630,12 +2632,30 @@ class Package {
                   );
                 }
               } else if (this.zipExt == ".zip") {
-                const result = spawnSync("unzip", [
-                  "-q",
-                  tempFile,
-                  "-d",
-                  this.installDirectory,
-                ]);
+                let result;
+                if (this.platform.includes("windows")) {
+                  // Windows does not have "unzip" by default on many installations, instead
+                  // we use Expand-Archive from powershell
+                  result = spawnSync("powershell.exe", [
+                    "-NoProfile",
+                    "-NonInteractive",
+                    "-Command",
+                    `& {
+                        param([string]$LiteralPath, [string]$DestinationPath)
+                        Expand-Archive -LiteralPath $LiteralPath -DestinationPath $DestinationPath -Force
+                    }`,
+                    tempFile,
+                    this.installDirectory,
+                  ]);
+                } else {
+                  result = spawnSync("unzip", [
+                    "-q",
+                    tempFile,
+                    "-d",
+                    this.installDirectory,
+                  ]);
+                }
+
                 if (result.status == 0) {
                   resolve();
                 } else if (result.error) {
@@ -2798,7 +2818,7 @@ const getPackage = () => {
   const url = `${artifactDownloadUrl}/${platform.artifactName}`;
   let filename = platform.artifactName;
   let ext = platform.zipExt;
-  let binary = new Package(name, url, filename, ext, platform.bins);
+  let binary = new Package(platform, name, url, filename, ext, platform.bins);
 
   return binary;
 };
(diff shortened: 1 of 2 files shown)


mistydemeo:
Thank you very much! I appreciate this! I'll give it a test and see how it works.

It looks like `prettier` is flagging a couple of requested formatting changes. Are you up for tweaking that? I can take care of it if you prefer.

The snapshots we use for tests will need to be updated with these changes. I can do that for you if you'd like, or you can do it by following these steps:

* Install [cargo insta](https://insta.rs/docs/cli/)
* Run the tests with `cargo test`
* Run `cargo insta accept`
* Commit and push the changes

jacobtread:
Hi, I have updated the formatting :) I will need you to please handle the test snapshots updating I think If I run them on my device it will cause lots of problems, as running the tests on my machine ends up failing a majority of unrelated integration and snapshot tests (I'd assume because I'm using windows and the tests don't take that into account) thank you!

Edit: The errors I was getting might've just been me running from the top level project, running the tests and cargo insta accept from the cargo-dist sub directory got me more acceptable outputs, I've commited the new snapshots for this

mistydemeo:
Thanks! Looks like one snapshot didn't take, so I've pushed a commit with that.

mistydemeo:
Looking good! I'll just do some testing on my Windows machine later to make sure all works as I expect, but I think this should be good.

jacobtread:
Awesome thank you :)

jacobtread:
Is there anything I can do to push this one forward for a new release? I'm hoping to use this in one of my projects some time soon  :) 

If not thats okay I can look into manually having it go up to npm separately 

Edit: found the cargo-dist-url-override option :) so I'm all good no need to rush with this

mistydemeo:
Sorry for the delay! I’ve just been very busy lately, and hope to have time to integrate and release this soon.

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -122,12 +124,25 @@ class Package {
...
-                ]);
+                let result;
+                if (this.platform.includes("windows")) {
+                  // Windows does not have "unzip" by default on many installations, instead
+                  // we use Expand-Archive from powershell
+                  result = spawnSync("powershell.exe", [
+                    "-NoProfile",
+                    "-NonInteractive",
+                    "-Command",
+                    `Expand-Archive -LiteralPath '${tempFile}' -DestinationPath '${this.installDirectory}' -Force`,
```
mistydemeo, line 135: Just had a realization. Does this introduce the possibility of shell injection? Previously that wasn't possible because `spawnSync` is being run with `shell=false`, but in this version the unzip command is being passed to powershell as a single command string. Could mean that install directories or tempfile names with powershell parameters in them could do things here they don't in the other paths.

jacobtread: Yes I think you are right there, I will have a look into other possible ways of executing the command

jacobtread:
> Sorry for the delay! I’ve just been very busy lately, and hope to have time to integrate and release this soon.

It's no worries!

I have modified the command to use args to prevent shell injection in the install dir / temp file name. 

I've updated the snapshots that would run for me the "axolotlsay_basic" test is failing on my machine due to "failed to generate an OmniBOR Artifact ID (status: exit code: 2)" so If you could please run that on your machine for me if I've missed any snapshots 😄

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -122,12 +124,27 @@ class Package {
...
-                ]);
+                let result;
+                if (this.platform.includes("windows")) {
+                  // Windows does not have "unzip" by default on many installations, instead
+                  // we use Expand-Archive from powershell
+                  result = spawnSync("powershell.exe", [
+                    "-NoProfile",
+                    "-NonInteractive",
+                    "-Command",
+                    "& { Expand-Archive -LiteralPath $args[0] -DestinationPath $args[1] -Force }",
```
StartAutomating, line 135: Expand-Archive is safe, but this technique of embedding isn't always so secure.

Essentially, because you're passing down an unquoted string in thru, it could theoretically get expanded into something nasty.  In practice, you'll probably be fine if the inputs are sanitized first (say, by confirming the file exists).

I'd personally feel more comfortable if the command were something like:

~~~PowerShell
& {param([string]$LiteralPath, [string]$DestinationPath) Expand-Archive @psboundParameters } 
~~~

This would force each parameter to be a single string, and pass the parameters directly down to Expand-Path.

mistydemeo: Great, thank you!

jacobtread:
Thank you!

mistydemeo:
I'll try to get a new release out soon.

jacobtread:
> I'll try to get a new release out soon.

No rush from me :) I'm using a fork I published for the time being thank you for merging though!

jacobtread:
@mistydemeo Hold off on deploying this to a release, I miss understood the .platform didn't realize that was an object, missed the line that pulled the object for it in from the package.json so the .includes is failing for zip installs I'll open a PR shortly

PR #2036, opened 2025-08-25
Add cargo-install instructions to install.md

egranata:
I was not aware of cargo-binstall and assumed it to be a typo. cargo install seems to work, and cargo binstall requires an extra step, worth documenting IMHO

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -52,17 +52,17 @@ nix-env -i cargo-dist
 
 ### Other Options
 
-#### cargo-install
-
-```sh
-cargo install cargo-dist
-```
-
 #### cargo-binstall
 
 ```sh
 cargo install cargo-binstall # if you don't have binstall available already
 cargo binstall cargo-dist
+```
+
+#### cargo-install
+
+```sh
+cargo install cargo-dist
 ```
 
 ## Build From Source


mistydemeo:
`cargo install` installs from source and from non-locked dependencies (which has broken before!), so I think I'd prefer it if `cargo-binstall` came first. Make sense to you?

egranata:
> `cargo install` installs from source and from non-locked dependencies (which has broken before!), so I think I'd prefer it if `cargo-binstall` came first. Make sense to you?

Yeah makes sense. I have updated the PR with swapping the two. Of course, I can't seem to squash the two commits into one, so that's something you may wanna do when you merge this?

mistydemeo (approved):

mistydemeo:
Actually, I just noticed the section right below this is for `cargo install`. I get that with `binstall` coming first there was some confusion around what this was meant to be - maybe we can reword the header for the `cargo install` section that's already there so it's clearer what it's suggesting?

book/src/install.md
```diff
@@ -55,9 +55,16 @@ nix-env -i cargo-dist
...
 
 ```sh
+cargo install cargo-binstall # if you don't have binstall available already
 cargo binstall cargo-dist
 ```
 
+#### cargo-install
+
+```sh
+cargo install cargo-dist
```
JadedBlueEyes: `cargo install --locked cargo-dist`, as well as mentioning that this will install from source

PR #2046, opened 2025-08-27
fix: platform.includes bug

jacobtread:
Sorry introduced a bug in my last PR thats crashing when trying to extract .zip missed the fact that the platform tuple string gets turned into an object from the package.json contents so the .includes I was using expecting a string here is failing

Thankfully looks like windows is the only one that uses the .zip path so I haven't broken anything that wasn't working already but my bad 😓

### DIFF
--- a/cargo-dist/tests/snapshots/axolotlsay_basic.snap
+++ b/cargo-dist/tests/snapshots/axolotlsay_basic.snap
@@ -2672,7 +2672,7 @@ class Package {
                 }
               } else if (this.zipExt == ".zip") {
                 let result;
-                if (this.platform.includes("windows")) {
+                if (this.platform.artifactName.includes("windows")) {
                   // Windows does not have "unzip" by default on many installations, instead
                   // we use Expand-Archive from powershell
                   result = spawnSync("powershell.exe", [


jacobtread:
@mistydemeo  Will want this to be merged before trying to release my previous PR if you could please handle any missing snapshots for me thank you sorry for any inconvience

mistydemeo (approved):

mistydemeo:
Thank you for the quick fix! I appreciate you catching this. The feature (and fix) are available in 0.30.0-prerelease.2 and .3.

jacobtread:
Missed it completely during my testing on my republished fork due to a fail on my part at self hosting cargo-dist I hadn't changed the git repo when I republished it so all the installers were just downloading normal cargo-dist instead of my fork 🤦‍♂️ 

Thank you for the timely release!

jacobtread:
@mistydemeo 0.30.0-prerelease.2 and .3. does not seem to include this fix😅

jacobtread:
I have confirmed the fix is all working as intended now though :) using my republished fork the setup is working on windows without issue after the fixes are applied

mistydemeo:
Argh, you're right. I missed backporting that commit to the 0.x branch. Doing that now, and will cut new prereleases.

jacobtread:
Thank you!

PR #2070, opened 2025-09-03
feat: optionally apply `github-build-setup` steps to additional jobs

### LINKED ISSUE #2065: Supporting additional environment setup steps (pre-planning)
I have a private crate that is included as a Git dep in my project. In order for `cargo` operations against the project to succeed I have to configure Github Actions to use an access token to pull the repository.

```yaml
- run: git config --global url."https://x-access-token:${{ secrets.PRIVATE_REPO_ACCESS_TOKEN }}@github.com/".insteadOf "ssh://git@github.com/"
```

This works as expected in my other CI workflows. However, `dist`s plan phase runs `cargo metadata`, which requires that repo be cloneable. Unfortunately there doesn't appear to be a hook available that lets me inject this as a prelude to the various jobs in the release workflow. `github-build-setup` gets close, but that only affects the `build-local-artifacts` job.

Is there a way to express this with a non-dirty `release.yml`?

arusahni:
For some use cases, such as using private git crate dependencies in a project's Cargo.toml manifest, Github Actions requires a certain amount of ceremony to alias the private repo's Git URL to to something that can be cloned using an access token. For the `build-local-artifacts` job, this can be accomplished by adding the authentication step in the `github-build-setup` YML file.

There are additional jobs that necessitate these credentials be set, mainly the `plan`, `build-global-artifacts`, and `host`. This appears to be due to the fact they invoke `cargo metadata` (which attempts to resolve dependencies).

Support these use-cases by introducing a `github-build-setup-jobs` collection which allows users to opt-in to including the `github-build-setup` steps in various jobs.

Fixes #2065

### DIFF
--- a/cargo-dist/templates/ci/github/release.yml.j2
+++ b/cargo-dist/templates/ci/github/release.yml.j2
@@ -128,6 +128,9 @@ jobs:
         with:
           name: cargo-dist-cache
           path: ~/.cargo/bin/dist
+      {{%- if "plan" in github_build_setup_jobs %}}
+{{%- include 'ci/github/partials/build_setup_jobs.yml' %}}
+      {{%- endif %}}
       # sure would be cool if github gave us proper conditionals...
       # so here's a doubly-nested ternary-via-truthiness to try to provide the best possible
       # functionality based on whether this is a pull_request, and whether it's from a fork.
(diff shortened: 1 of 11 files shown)


mistydemeo:
Thanks for the submission!

This makes sense, and I understand the motivation, but I'm not sure always applying this is the best solution. The original motivation for this feature was to install package dependencies for builds - for example, a user needed to fetch a custom copy of Lua. This is a bit too heavy to apply to all steps if it's only neeed for builds; it would significantly add to the time it takes to run the full action without any benefit.

How about adding a new setting where we configure what phases to add the `github-build-setup` code to? It would default to `local-build` but could be extended to list other phases too, like what you're hoping for.

arusahni:
@mistydemeo that suggestion makes sense to me. Happy to make those changes.

mistydemeo:
Thank you! I appreciate it. Happy to review when you’re ready.

arusahni:
Alright, I think this is ready for review @mistydemeo. I appreciate your assistance!

mistydemeo:
I'll review soon - sorry for the delay! I've been busy.

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -128,6 +128,53 @@ jobs:
         with:
           name: cargo-dist-cache
           path: ~/.cargo/bin/dist
+      {{%- if "plan" in github_build_setup_jobs %}}
+      {{%- for step in github_build_setup %}}
```
mistydemeo: Now that this is being repeated in a number of places, how about turning it into a partial pulled in via an include statement? That way we can avoid inlining exactly the same logic in a number of places. You can see a sample `include` statement here: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/templates/ci/github/release.yml.j2#L660

And the partials go in this directory: https://github.com/axodotdev/cargo-dist/tree/main/cargo-dist/templates/ci/github/partials

mistydemeo (changes requested):
Thanks again for your understanding on my delay getting to this. Your PR came in just as I was moving to a new home, and it took me awhile after the big stressful move to have the energy for OSS again. 😅 

We've made some changes to the base branch, so this will need a rebase. I've made one other suggestion too.

arusahni:
@mistydemeo no worries at all. Moving can be a lot, and OSS maintenance can be thankless at times.

I have rebased and eliminated the duplicated rules.

arusahni:
Whoops, missed that format issue. I have fixed it and rebased on `main`.

arusahni:
I noticed this PR was pretty stale, so I rebased it and addressed the conflicts against main.

PR #2079, opened 2025-09-07
feat: use canonical TitleCase for LoongArch64 Linux targets

xen0n:
The architecture's name is better spelled as "LoongArch" in documents, instead of "LOONGARCH" which looks like an identifier.


mistydemeo (approved):
Thanks!

Issue #2133, opened 2025-10-09
Enabling GitHub attestations results in overly broad permissions


samypr100:
Currently, when `github-attestations = true`, `release.yml` ends up having a permission structure like shown below

```yaml
name: Release

permissions:
  "attestations": "write"
  "contents": "write"
  "id-token": "write"

...

jobs:

...
```

In practice, having `attestations` and `id-token` write at the workflow level is considered bad practice per [zizmor](https://docs.zizmor.sh/audits/#excessive-permissions).

Ideally, these permissions should be set at the corresponding job that generates attestations instead. For example,

```yaml
...
build-local-artifacts:
...
    permissions:
      "attestations": "write"
      "contents": "read"
      "id-token": "write"
    steps:
...
```

or

```yaml
host:
...
    permissions:
      "attestations": "write"
      "contents": "write"
      "id-token": "write"
    steps:
...
```

Since these are builtin jobs, I believe these permissions would not need to be exposed to the end-user for configurability.

PR #2134, opened 2025-10-09
fix: avoid overly broad permissions when running attestations

samypr100:
Closes https://github.com/axodotdev/cargo-dist/issues/2133


mistydemeo:
Since there are only two possible options here, it looks like it'd be simpler to just inline the two options into the template instead of putting most of the logic in Rust.

samypr100:
> Since there are only two possible options here, it looks like it'd be simpler to just inline the two options into the template instead of putting most of the logic in Rust.

@mistydemeo done in dab8f027fa9d25e5dcc8fead578c7cac8fde438e

mistydemeo (approved):
Thank you! Looks good.

Issue #2130, opened 2025-10-06
Add support for `x86_64-unknown-freebsd`


DenuxPlays:
my usecase would be to write software for OPNSense.

The cross compilation from linux is fairly easy (and effortless using cross).

Is there a chance that support would be added for this target?

mistydemeo:
That seems reasonable! Are you interested in doing it, or would you like me to implement it? If I do, are you able to test some test binaries I produce and see if they work?

DenuxPlays:
I would love to implement it but as always I dont have the time sadly :(

But I would be able to test it.
I would also be able to supply builds for a small cli to test it

DenuxPlays:
Hey,

is it possible to make something like this work in the mean time:
```toml
[dist]
targets = ["x86_64-unknown-freebsd"]

[dist.github-custom-runners.x86_64-unknown-freebsd]
runner = "ubuntu-latest"
container = { image = "ghcr.io/cross-rs/x86_64-unknown-freebsd:latest" }
```

PR #2139, opened 2025-10-13
FreeBSD Support

DenuxPlays:
closes: #2130 

Adds support for FreeBSD cross compilation using cross.

This PR also adds the ability to cross compile with cross.

TODOs:

- [x] add tests
- [x] install cross
- [ ] manual testing

### DIFF
--- a/cargo-dist/tests/integration-tests.rs
+++ b/cargo-dist/tests/integration-tests.rs
@@ -1946,7 +1946,7 @@ fn axolotlsay_cross1() -> Result<(), miette::Report> {
 [workspace.metadata.dist]
 cargo-dist-version = "{dist_version}"
 installers = ["shell", "powershell"]
-targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "aarch64-apple-darwin", "x86_64-pc-windows-msvc", "aarch64-pc-windows-msvc"]
+targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "aarch64-apple-darwin", "x86_64-pc-windows-msvc", "aarch64-pc-windows-msvc", "x86_64-unknown-freebsd"]
 ci = ["github"]
 unix-archive = ".tar.gz"
 windows-archive = ".tar.gz"
--- a/cargo-dist/tests/snapshots/axolotlsay_cross1.snap
+++ b/cargo-dist/tests/snapshots/axolotlsay_cross1.snap
@@ -267,6 +267,18 @@ download_binary_and_run_installer() {
             _updater_name=""
             _updater_bin=""
             ;;
+        "axolotlsay-x86_64-unknown-freebsd.tar.gz")
+            _arch="x86_64-unknown-freebsd"
+            _zip_ext=".tar.gz"
+            _bins="axolotlsay"
+            _bins_js_array='"axolotlsay"'
+            _libs=""
+            _libs_js_array=""
+            _staticlibs=""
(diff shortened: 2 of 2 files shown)


DenuxPlays:
@mistydemeo 

Hey,
I found some time implementing freebsd & cross support.
I've tested it locally and building works.

I would really like to add some automated tests can you tell me how to do that?
I would also like to know if I could test the github workflow locally somehow?

Also I found out during debugging that your debug profile is really bad for debugging.
Due to opt-level being 2 a lot of variables are getting inlined & removed so debugging using lldb gets kind of impossible.
Just wanted to say that for I know I commented out the custom debug profile override.

mistydemeo:
Wonderful, thank you!

We won't be able to do automated tests that actually produce the binaries, but we can test the CI and installer setup by adding more tests to the integration tests here. Maybe add freebsd as a platform to one of the existing tests and see if the diff from `cargo test` looks right? https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/tests/integration-tests.rs

If you'd like to do some manual testing, you can do something like my cargodisttest repo with a toy app you have set up to build for freebsd using a git branch as the source for fetching cargo-dist: https://github.com/mistydemeo/cargodisttest

Let me know if you need help with any of that.

DenuxPlays:
I have another question.

Was it correct to start from main or should I have started my work from the stable branch?

DenuxPlays:
I also could not test the release.

I've created cargodisttest repo and tried to use my version of dist but I get this error:
https://github.com/DenuxPlays/cargodisttest/actions/runs/18706582046/job/53345700511

Maybe you can help me figure this out?

mistydemeo:
I'm very sorry for the late reply!

Taking a look at your build, it looks like the problem is that the changes you've made to your local generated config for testing aren't reflected in the code in this branch.

When dist runs, it checks to see that your generated CI config is identical to what dist would generate. In this case, you've hand-edited the CI config to pull down your build of dist, but you haven't changed the template in your branch to have those same changes.

You can use the `allow-dirty` setting to disable this check. https://axodotdev.github.io/cargo-dist/book/ci/customizing.html#hand-editing-releaseyml Keep in mind that *does* mean that if you make any new changes to the template as a part of this branch, you'll have to remember to make those changes in your CI config. Alternately, you can update the template in your branch here to point at your repo temporarily while you're testing and only back that out when we're ready to merge.

Again, I'm sorry for the inconvenience. :/ I know these aren't super great answers. I'll also take a look at our from-git dev config and see about setting up a way to make it possible to configure a different remote for the repo instead of hardcoding axodotdev.

DenuxPlays:
Hey,

all good thanks for the reply.
I tried it now i get a different error but I think I caused that.
Gonna take a look at it soon

DenuxPlays:
Hey,
this has to wait a bit sorry :/

I am currently fully invested in preparing for the spring boot 4 migration.

If you want you can a look at the error I meant:
https://github.com/DenuxPlays/cargodisttest/actions/runs/18866003066/job/53833673844

We plan to finish the migration in novemeber so I'll have time again at the end of november.

mistydemeo:
Hm, I think you might have just forgotten to bump the version in your Cargo.toml! It looks like the tag you pushed is 0.2.1, but the program itself doesn't have version 0.2.1. cargo-dist validates that the version you're trying to release actually exists in the Cargo package.

mistydemeo:
And no worries - please take your time! I appreciate you taking the time to work on this.

DenuxPlays:
Did some testing I will hopefully find time to finish next week...

Note to me:
Current Probem: cargo binstall is getting installed but cross is not

PR #2178, opened 2025-10-29
Release 0.30.1

Gankra:


mistydemeo:
May also be worth seeing if we can get #2070 in.

Gankra:
Hm interesting, that PR seems like it needs some non-trivial fixes, and I want to unblock the use of the attestations feature (the zizmor linter isn't happy with the overly broad permissions). 

I'm happy to review that and do a second release this weekend though.

mistydemeo:
Sounds good, don't think it has to get done immediately.

PR #2181, opened 2025-10-30
Support Attestations in Announce Phase

samypr100:
When https://github.com/axodotdev/cargo-dist/pull/2134 fixed the issues reported by [zizmor](https://github.com/zizmorcore/zizmor) from https://github.com/astral-sh/uv/pull/11357 were merged, the attestation permissions were no longer at the root level of the release workflow as they previously were traditionally. This introduced a side-effect of needing to have a mechanism to differentiate between attesting in host vs announce phase to know where to add the permissions.

This PR provides support when `github-release = "announce"` is set, and you needed to attest in the `host` phase. You can now explicitly set the attest phase to `announce` via `github-attestations-phase = "announce"` to ensure the permissions are correct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Unreleased
 
-Nothing Yet!
-
+* When customizing when artifacts are attested, you can now use the `announce` phase.
 
 # Version 0.30.1 (2025-10-29)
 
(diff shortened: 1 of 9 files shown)


Gankra (approved):

Gankra:
Looks great!

Gankra:
I've rebased it but FYI the "real" main branch of this repo rn is actually the "stable" one -- it's on my TODO list to swap main and stable (current main was for a now defunct 1.0.0 push).

samypr100:
Good to know

PR #2185, opened 2025-11-01
chore: reset to stable branch

mistydemeo:
This resets the state of `main` to the current tip of `stable`. I've made this a commit instead of force-pushing `main` so that contributors can just `git pull` smoothly. I've also pushed a copy of the old `main` to the `config_1_0` branch for posterity.


Gankra (approved):
Thank you!

Issue #2191, opened 2025-11-07
[Feature PR] Improve ci configuration for `.github/workflows/release.yml`


Mooling0602:
I've found shortcomings in the current `release.yml` file: in the "Install Rust non-interactively if not already installed" section, if `rustc` and `cargo` tools are already installed, the installation step is skipped, ultimately causing the build of the product to fail when the pre-installed tool version is lower than the minimum required by the project.

It caused some errors in my project's Actions tasks:
- https://github.com/Mooling0602/Daemon_Console_Rust/actions/runs/19165853500
- https://github.com/Mooling0602/Daemon_Console_Rust/actions/runs/18746468997

I wrote a better `release.yml` template file, it reads `rust-version` option in `Cargo.toml`, and if current rust (cargo/rustc) version lower than it, then reinstall rust tools to latest version.

I want to make a pull request to merge the `release.yml` file I modified with ChatGPT. I'll do it later and refer to this issue there.

I want to make a contribution to this great project and hope it gets better!

Gankra:
I believe your problem would be better solved by adding a rust-toolchain.toml to your project, which is the community-standard way to get CI to use the toolchain you want: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Mooling0602:
> I believe your problem would be better solved by adding a rust-toolchain.toml to your project, which is the community-standard way to get CI to use the toolchain you want: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Yes, thanks for your reply.
I'm thinking if force install latest rust tools works better than "Install Rust non-interactively if not already installed", I think it couldn't affect rust-toolchain.toml file's work.

wade-cheng:
I'm getting the compilation error `error[E0658]: let expressions in this position are unstable` in release CI for _only_ aarch64-pc-windows-msvc. The error mentions https://github.com/rust-lang/rust/issues/53667, which leads me to believe dist is supplying an outdated rust somehow?

I have 
```
       "artifacts": [
        "source.tar.gz",
        "source.tar.gz.sha256",
        "compile-typst-site-installer.sh",
        "compile-typst-site-installer.ps1",
        "sha256.sum",
        "compile-typst-site-aarch64-apple-darwin.tar.xz",
        "compile-typst-site-aarch64-apple-darwin.tar.xz.sha256",
        "compile-typst-site-aarch64-pc-windows-msvc.zip",
[...]

PR #2192, opened 2025-11-07
feat: auto detect needed rust tool version before building release

Mooling0602:
Implemention for: https://github.com/axodotdev/cargo-dist/issues/2191


Gankra:
This PR is editing the release infra for our own project and not cargo-dist itself.

Also I think just using a rust-toolchain.toml is the fix for this problem.

Mooling0602:
> This PR is editing the release infra for our own project and not cargo-dist itself.
> 
> Also I think just using a rust-toolchain.toml is the fix for this problem.

Yes, I think so.
Anyway thanks for your reply.

PR #2195, opened 2025-11-10
fix: don't preserve file ownership when extracting tarball

atoerien:
This is a simple fix for https://github.com/astral-sh/uv/issues/12942, where if you run the installer as root the resulting binaries end up with weird UID/GID values.

When running as root, tar uses `--same-owner` by default, which causes the issue. So just specifying `--no-same-owner` (the default for non-root users) fixes the problem.

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -305,7 +305,7 @@ download_binary_and_run_installer() {
             ;;
 
         ".tar."*)
-            ensure tar xf "$_file" --strip-components 1 -C "$_dir"
+            ensure tar xf "$_file" --no-same-owner --strip-components 1 -C "$_dir"
             ;;
         *)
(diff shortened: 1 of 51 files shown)


mistydemeo:
Good catch, thank you!

The test fixtures will need updating; would you like me to do that for you?

atoerien:
@mistydemeo Er I don't really know where I need to be changing things, so yes that would be much appreciated :)

Also I just did a search for `--strip-components` in github and I see there's a `binary-install.js` and `installer.ps1` - I'm not sure if this change needs to be made there as well.

mistydemeo:
`installer.ps1` runs on Windows and I don't think Windows handles user permissions the same way. We should check if the CLI flag exists on Windows `tar`. As for `binary.js`... that's the npm installer. That one should be possible to run as root, so we might also need it there.

mistydemeo (approved):

PR #2227, opened 2025-12-19
 fix: installer receipt for windows with posix shell #51 

eegli:
Addresses https://github.com/astral-sh/uv/issues/12686.

## Notes
- The detection of POSIX shells on Windows is inspired by CPython's `venv` std lib implementation: https://github.com/python/cpython/blob/9f8ec95bb079c86c82a397516ea5da0b2cc1033d/Lib/venv/scripts/common/activate#L40
- Not really sure what's the best way to test this, having a dedicated MSYS2 test environment or similar would require some additional effort. There's https://github.com/msys2/setup-msys2 if CI tests are sufficient.

Supersedes https://github.com/astral-sh/cargo-dist/pull/51.

### DIFF
--- a/cargo-dist/tests/snapshots/axolotlsay_basic.snap
+++ b/cargo-dist/tests/snapshots/axolotlsay_basic.snap
@@ -84,6 +84,29 @@ get_home() {
         getent passwd "$(id -un)" | cut -d: -f6
     fi
 }
+
+# Check if running on Windows with POSIX-compliant shell (CYGWIN, MSYS, MINGW)
+is_windows_posix() {
+    case "$(uname)" in
+        CYGWIN*|MSYS*|MINGW*)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+# Convert POSIX path to Windows path
+convert_path_for_receipt() {
+    local _path="$1"
+    if is_windows_posix && check_cmd cygpath; then
+        # Use cygpath to convert path, then escape backslashes for JSON
+        cygpath -w "$_path" | sed 's/\\/\\\\/g'
+    else
+        echo "$_path"
+    fi
+}
+
 # The HOME reference to show in user output. If `$HOME` isn't set, we show the absolute path instead.
 get_home_expression() {
     if [ -n "${HOME:-}" ]; then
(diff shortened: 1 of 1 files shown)


eegli:
Summoning @konstin

mistydemeo:
~~Are we sure `cygpath` is always available in MSYS2/MinGW environments?~~ Sorry, just noticed you check for its presence before running.

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -79,6 +79,29 @@ get_home() {
...
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+# Convert POSIX path to Windows path
+convert_path_for_receipt() {
+    local _path="$1"
+    if is_windows_posix && command -v cygpath > /dev/null 2>&1; then
```
mistydemeo: We've already got a helper in this script for checking for the presence of commands: `check_cmd`. https://github.com/axodotdev/cargo-dist/blob/c53b7d91e9daf36bbedf8a37a60a3e9863142862/cargo-dist/templates/installer/installer.sh.j2#L1311C1-L1314

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -79,6 +79,29 @@ get_home() {
...
+        # Use cygpath to convert path, then escape backslashes for JSON
+        cygpath -w "$_path" | sed 's/\\/\\\\/g'
+    else
+        echo "$_path"
```
mistydemeo: In this case we know the path is likely to be wrong but we're not going to be able to convert it, right?

eegli:
Snapshots are now up to date, took me a while to figure out `omnibor-cli@0.7.0` is required in the path to make `axolotlsay_basic` pass.

mistydemeo:
Thanks! Sorry for the inconvenience with getting the tests running.

I see in your comment on the original issue that there were two problems - that the install location was different than expected, but also the path issue you're fixing here. Will the updater still not work properly if we fix the path issue with this PR?

eegli:
@mistydemeo no worries.

> Will the updater still not work properly if we fix the path issue with this PR?

The PR fixes both: On Windows, we now store the receipt in `$LOCALAPPDATA` (which is usually `C:\Users\<user>\AppData\Local`), and the path is converted via `cygpath`, a utility that is available in all Cygwin-based environment/shells. Since - very roughly - Git Bash ⊂ MSYS2 ⊂ Cygwin, the updater shell script has great support, especially since most Windows people likely work in Git Bash (myself included). And it's of course still compatible with Linux/MacOS.

eegli:
In the worst case, (e.g., `$LOCALAPPDATA` is not set and `cygpath` does not exist (not sure if/when this is possible), we just fall back to the current behavior (which is exactly what is outlined in the original issue).

eegli:
Hey @mistydemeo, do you plan to merge this at some point?

mistydemeo (approved):
Sorry for the delay! It's been a busy couple of weeks. Thank you for this!

Looks like there's one other PR up for something touching Astral's installers, so we'll likely look to get these out in the same release.

eegli:
No worries. Happy we could make this happen!

PR #2223, opened 2025-12-15
fix: typo in variable name

20jasper:
Fixes a typo in the `builderGlibcMInorVersion`. The `I` in MInor shouldn't be capitalized

PR #2234, opened 2025-12-28
chore(docs): update contributing guidelines to include omnibor

20jasper:
While working on #2223, I found `omnibor` is required to run integration tests. This PR includes a script to install required testing tools

However, there are a few things I'd like to clear up before going further

### README/CONTRIBUTING.md overlap

Both the README.md and CONTRIBUTING.md have information about contributions—some is overlapping and some is not. Would y'all be interested in any of these options?
1. move contributing info to CONTRIBUTING.md and link in the README
2. move contributing info fully to the README

### Nix 

Would y'all prefer updating the devenv.nix be 
- made into a separate issue
- note the current failures are expected when using the devenv.nix

As of now, the integration test using omnibor fails in the nix devenv, and it seems like not an insignficant amount of work to add omnibor to the devenv.nix, though of course I could be missing something obvious since this is the first time I've used nix

From what little I know of nix, running an install script for something without a deterministic lockfile is not the preferred way to do things. For reference, [omnibor-rs does not provide a `Cargo.lock`](https://github.com/omnibor/omnibor-rs). I have not checked if there is another deterministic build for this beyond directly downloading the artifacts

### Troubleshooting

I was unable to install omnibor via the devenv.nix

It did not seem that omnibor was already available to use like cargo-insta was https://search.nixos.org/packages?channel=25.11&query=omnibor-cli

Had trouble locally installing nix, so ran a dockerfile with the following configuration
```
FROM nixos/nix 

RUN nix-env --install --attr devenv -f https://github.com/NixOS/nixpkgs/tarball/nixpkgs-unstable

COPY . /workdir
WORKDIR /workdir

ENTRYPOINT [ "bash" ]
```

After dropping into the shell with `devenv shell` and running tests via `cargo test`, reproduced the integration test failure
```sh
test axolotlsay_basic ... FAILED
```

I got the git hash from the omnibor releases [chore: Release omnibor-v0.7.0 · omnibor/omnibor-rs@80ff804 · GitHub](https://github.com/omnibor/omnibor-rs/commit/80ff804c1048ea0ded03d84510b1d1cbdaf98229) 

As mentioned above, no Cargo.lock complicates building from source

It seems possible to set different download URLs per OS and architecture but wanted to check in before spending too much time here

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,12 +20,12 @@ that your fix works.
 **3. Overcommunicate**
 
 In all scenarios, please provide as much context as possible- you may not think
-it's important but it may be! 
+it's important but it may be!
 
 **4. Patience**
 
 Axo is a very small company, it's possible that we may not be able to
-immediately prioritize your issue. We are excite to develop a community of 
+immediately prioritize your issue. We are excited to develop a community of
 contributors around this project, but it won't always be on the top of our to-do
 list, even if we wish it could be.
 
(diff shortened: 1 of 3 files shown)


OpenSauce:
I certainly got caught out with the omnibar issue so this would be great (and for what my opinion is worth I think option 1 would be fine :stuck_out_tongue:)

mistydemeo:
Sorry for the delay responding!

> move contributing info to CONTRIBUTING.md and link in the README

I think this is probably the better approach - I think the two had gotten out of sync due to being separate, so having them in one place will probably help.

I'm fine with making updating devenv.nix a separate issue from this. It sounds trickier, anyway.

Issue #236, opened 2023-05-03
artifact-download-url customization


Gankra:
Currently we only support hardcoded support for github artifact download urls, which makes installers gated behind github. We can solve this by just letting the user specify. This should possibly support some amount of templating, as seen in [the config for cargo-binstall](https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md).

It might also help [github enterprise users](https://github.com/axodotdev/cargo-dist/issues/185).

Theoretical usecases:

* `artifact-download-url = "{repo}/releases/downloads/{tag}/" (what we currently do for github)
* `artifact-download-url = "https://mywebsite.com/{tag}/"`

Exposing things like "app-name" and "app-version" are sketchier with workspaces but we can maybe allow them for single-app workspaces?

ashleygwilliams:
interesting. this feels complicated- but complicated is fine if the end users are advanced. can we offer a way for an end user to validate/test what they've set up?

Gankra:
Ideally this would use the same solution we pick for https://github.com/axodotdev/oranda/pull/507

PR #2250, opened 2026-01-11
Add support for customizing artifact-download-url

charliermarsh:
## Summary

This PR allows users to customize the download URL, to support installing from a mirror. Fallback to GitHub Actions is also supported.

We're looking to use this in Ruff, uv, and ty to allow us to host our artifacts on a non-GitHub host with fallback.

Closes https://github.com/axodotdev/cargo-dist/issues/236.


charliermarsh:
\cc @Gankra

Gankra:
Replaced by #2285

PR #2261, opened 2026-01-21
Reduce risk of an interrupted installation leading to a partial installation

EliteTK:
Since the installer script uses `mktemp -d` to create a directory to store the downloaded release and extract it, the files might end up on a different filesystem than the target installation directory. This means that the `mv` basically turns into a `cp` followed by an `rm` which means that if it gets interrupted at this point, the result could be a partially written file, or just in general an installation with only some of the files written.

There's no real way to solve this without support from the filesystem, so instead this PR aims to reduce the window in these cases by first using `mv` to copy/move the files into a subdirectory of the target and then `mv` the files out. The second round of `mv` should just be `rename` under the hood so this should happen quickly. As an additional side effect, while you might end up with a partial upgrade, you will never end up with a partially written binary.

Also, since the temporary directory creation was successful, this means that the final round of `mv` is likely to also be successful as both require the same permissions to the parent directory.

### DIFF
--- a/cargo-dist/tests/snapshots/akaikatana_basic.snap
+++ b/cargo-dist/tests/snapshots/akaikatana_basic.snap
@@ -756,10 +756,10 @@ install() {
     # permissions.
 
     for _bin_name in $_bins; do
-            ensure mv "$_install_temp/$_bin_name" "$_install_dir"
-            for _dest in $(aliases_for_binary "$_bin_name" "$_arch"); do
-                    ln -sf "$_install_dir/$_bin_name" "$_install_dir/$_dest"
-            done
+        ensure mv "$_install_temp/$_bin_name" "$_install_dir"
+        for _dest in $(aliases_for_binary "$_bin_name" "$_arch"); do
+            ln -sf "$_install_dir/$_bin_name" "$_install_dir/$_dest"
+        done
     done
     for _lib_name in $_libs $_staticlibs; do
         ensure mv "$_lib_install_temp/$_lib_name" "$_lib_install_dir"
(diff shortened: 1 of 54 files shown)


cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -1456,7 +1471,7 @@ verify_checksum() {
             local _test_blake2s
             _test_blake2s="$(printf "can do blake2s" | b2sum -a blake2s | awk '{printf $1}')" || _test_blake2s=""
 
-            if [ "X$_test_blake2s" = "X$_well_known_blake2s_checksum" ]; then
+            if [ "$_test_blake2s" = "$_well_known_blake2s_checksum" ]; then
```
Gankra: Heh, I added a shellcheck ignore in my PR because I didn't want to think about if this is indeed safe to drop on all our supported platforms (if you have a strong sense I'm happy to land this).

EliteTK:
I couldn't figure out how to run the `_basic` test and wasn't sure if there was a snapshot for it. I'll try to get it working or manually edit the snapshots.

Gankra:
Yeah I discovered this week that the basic test is subject to some unfortunate bitrot, the only way to run it is to install exactly this copy of omnibor:

https://github.com/axodotdev/cargo-dist/blob/main/.github/workflows/ci.yml#L124

(I should rip out the omnibor stuff if it doesn't even work...)

mistydemeo:
I was checking GitHub and it looks like at least one project out there is using the omnibor support. It's unfortunate they changed the CLI options in a backwards-incompatible way; makes it harder for us to support older and newer versions.

Issue #1335, opened 2024-08-17
bun build drops bins in root of dir, hard to gitignore


ashleygwilliams:

Gankra:
I swear i've filed this bug like 5 times but apparently I never do.

Right now we require you to dump your build output next to your manifest file which is whack.

mistydemeo:
I filed an issue for this a bit ago: #874

mistydemeo:
Implemented by https://github.com/axodotdev/cargo-dist/pull/2271.

Issue #874, opened 2024-03-24
Generic builds: provide config for artifact subdir


mistydemeo:
Right now, we assume that generic build artifacts are located at the root of the build directory. We should add a configuration value that configures an alternate root to locate artifacts.

I'm experimenting with packaging an autotools-based package, and for packages using this and cmake, it would be much simpler to be able to root cargo-dist's binaries inside a `build` or `build/bin` directory.

mistydemeo:
Implemented by #2271.

PR #2271, opened 2026-01-31
feat: support out_dir in config

CatBraaain:
dist-workspace.toml now supports package.out-dir.
Previously, for non-Rust projects, binaries were assumed to be placed directly under the root. With this change, build artifacts can be stored in an arbitrary folder.

Related issues:
- https://github.com/axodotdev/cargo-dist/issues/874
- https://github.com/axodotdev/cargo-dist/issues/1335

### DIFF
--- a/cargo-dist/src/build/generic.rs
+++ b/cargo-dist/src/build/generic.rs
@@ -51,11 +51,16 @@ impl<'a> DistGraphBuilder<'a> {
             }
             for (pkg_idx, expected_binaries) in builds_by_pkg_idx {
                 let package = self.workspaces.package(pkg_idx);
+                let out_dir = if let Some(dir) = package.out_dir.as_ref() {
+                    package.package_root.join(dir)
+                } else {
+                    package.package_root.clone()
+                };
                 builds.push(BuildStep::Generic(GenericBuildStep {
                     target_triple: target.clone(),
                     expected_binaries,
                     working_dir: package.package_root.clone(),
-                    out_dir: package.package_root.clone(),
+                    out_dir,
                     build_command: package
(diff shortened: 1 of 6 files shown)


CatBraaain:
@ashleygwilliams
This PR introduces a change that is very important for non-Rust projects.
Once this is merged, I’m planning to adopt it in my own project.
If possible, I’d really appreciate it if you could take a look at this with priority.
When you have time, I’d really appreciate a prioritized review.

Have a good day!

mistydemeo:
Hi! Sorry, I haven't had the time to review this, but I'm going to try to get to it soon.

cargo-dist/src/build/generic.rs
```diff
@@ -55,7 +55,9 @@ impl<'a> DistGraphBuilder<'a> {
                     target_triple: target.clone(),
                     expected_binaries,
                     working_dir: package.package_root.clone(),
-                    out_dir: package.package_root.clone(),
+                    out_dir: package
```
mistydemeo: I think it would make sense to have an assignment that checks if `package.out_dir` is defined, and branch based on that. So, up earlier in the file, you might do something like:

```rust
let out_dir = if let Some(dir) = package.out_dir {
  package.package_root.join(dir)
} else {
  package.package_root.clone()
};
```

And then pass that `out_dir` here.

CatBraaain: Thanks for the reviewing. I’ve updated the code to assign `out_dir` to a variable and branch with if let Some(...) as recommended.

mistydemeo (approved):
Thanks!

PR #2285, opened 2026-02-12
Add "simple" hosting style

Gankra:
This is an alternative to #2250 that more thoroughly applies the URL by treating it like a hosting style. This makes the config... much more gross, but ensures the proper value is propagated, and lays groundwork for a potential `mirror-upload-url` (right now astral plans to do this via a custom job).

It does this primarily by introducing a `simple-download-url = "https://static.mysite.com/{tag}"` value. Priority between the sources can be specified by using `hosting = ["simple", "github"]` vs `hosting = ["github", "simple"]`. One can also specify to only use the mirror with `hosting = ["simple"]`. If no ordering is specified, both are used and the mirror is preferred.

This PR intends to implement only support for this in the shell and powershell installers, so notably it punts on:

* npm installer (just uses the highest priority one)
* brew installer (just uses the highest priority one)

~~I'm still testing the shell/powershell changes.~~

### DIFF
--- a/cargo-dist-schema/src/lib.rs
+++ b/cargo-dist-schema/src/lib.rs
@@ -1096,7 +1096,7 @@ pub struct GithubHosting {
     pub repo: String,
 }
 
-/// Github Hosting
+/// Simple Hosting
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct SimpleHosting {
     /// The URL to download artifacts from
--- a/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
+++ b/cargo-dist-schema/src/snapshots/cargo_dist_schema__emit.snap
@@ -1225,7 +1225,7 @@ expression: json_schema
       ]
     },
     "SimpleHosting": {
-      "description": "Github Hosting",
+      "description": "Simple Hosting",
       "type": "object",
       "properties": {
         "download_url": {


Gankra:
let the record show that writing powershell and bash scripts you Really don't want to run locally, while also sick, is "hard"

Gankra:
Alright I need to tidy up the commits and add some docs but this *seems* to be working well now.

Some testing of the fallback behaviour is merited, in particular I saw some mysterious behaviour in CI where the download would fail but in a way that the download command considered a success. Not sure if that was a transient effect of my work or a legit issue (wouldn't reproduce locally -- not that surprising with how complex that curl/wget selection stuff is).

cargo-dist/src/config/v0_to_v1.rs
```diff
@@ -306,11 +323,15 @@ impl DistMetadata {
...
-                })
+                if install_location.is_some() || identifier.is_some() {
+                    Some(PkgInstallerLayer {
+                        common: CommonInstallerLayer::default(),
+                        identifier,
+                        install_location,
+                    })
+                } else {
+                    None
+                }
```
Gankra: This is a completely drive-by fix I ended up doing while debugging some apparent bitrot in the tests (as written it would always enable the pkg installer)

cargo-dist/templates/installer/installer.ps1.j2
```diff
@@ -49,20 +44,22 @@ param (
...
-}
-if ($env:{{ env_vars.download_url_env_var }}) {
-  $ArtifactDownloadUrl = $env:{{ env_vars.download_url_env_var }}
-} elseif ($env:INSTALLER_DOWNLOAD_URL) {
-  $ArtifactDownloadUrl = $env:INSTALLER_DOWNLOAD_URL
-} else {
-  $ArtifactDownloadUrl = "$installer_base_url{{ hosting.github.artifact_download_path }}"
+  $ArtifactDownloadUrls = @({% for url in base_urls -%}
+    "{{ url }}"{{ ", " if not loop.last else "" }}
+  {%- endfor %})
```
Gankra: This code concerned me for a bit, because it turns out the shell installers don't actually use base_url anymore after various features. However I was able to rearrange the conditions the preserve the same defacto behaviour for overridden env-vars, while still defaulting to the new cascading list of options.

Gankra:
Not convinced I actually fixed the pkg thing, it seems to hauntedly come and go on my machine which is... distressing.

mistydemeo:
The `pkg` issue sounds... mystifying. I can see if I can repro?

Gankra:
Yeah I was trying to find a good term for "static file server", and mirror felt the best, but I agree it's weird when it's the primary (which is the expected/intended case).

mistydemeo:
Arg, sorry about introducing a merge conflict with the `npm-shrinkwrap` branch.

PR #2302, opened 2026-02-20
feat(homebrew): mirror support with multi-urls

mistydemeo:
This adds Homebrew support for simple download URLs, and any other secondary URLs we'll be adding in the future. Homebrew has first-class mirror support via the `mirror` keyword. It allows arbitrary numbers of mirrors, which works perfectly for us.


Gankra:
Oh just realized the docs are now slightly wrong (they claim only sh and pwsh installers fully support this).

PR #2311, opened 2026-02-28
installer: add retry for transient network errors

### LINKED ISSUE #943: Add retry for axoupdater fetching
When we fetch axoupdater during the build, we should retry a few times on failure in case the issue is a transient network blip. Given how common these are on actions, it would help us avoid retrying entire broken builds.

adeebshihadeh:
[Example failure](https://github.com/commaai/openpilot/actions/runs/22524031665/job/65253149581) doing a uv install in our openpilot CI.

Also may close https://github.com/axodotdev/cargo-dist/issues/943 if I understand correctly?

---

from `man curl`:
```
       --retry <num>
              If  a transient error is returned when curl tries to perform a trans‐
              fer, it retries this number of times before giving  up.  Setting  the
              number  to  0  makes curl do no retries (which is the default). Tran‐
              sient error means either: a timeout, an FTP 4xx response code  or  an
              HTTP 408, 429, 500, 502, 503 or 504 response code.
```

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -1424,9 +1424,9 @@ downloader() {
     then need_cmd "$_dld"
     elif [ "$_dld" = curl ]; then
         if [ -n "${AUTH_TOKEN:-}" ]; then
-            curl -sSfL --retry 3 --retry-delay 3 --retry-all-errors --header "Authorization: Bearer ${AUTH_TOKEN}" "$1" -o "$2"
+            curl -sSfL --retry 3 --retry-delay 3 --header "Authorization: Bearer ${AUTH_TOKEN}" "$1" -o "$2"
         else
-            curl -sSfL --retry 3 --retry-delay 3 --retry-all-errors "$1" -o "$2"
+            curl -sSfL --retry 3 --retry-delay 3 "$1" -o "$2"
         fi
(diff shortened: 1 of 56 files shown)


mistydemeo:
Thanks! Sorry for the delay reviewing here.

Checking rustup's implementation, they check if the `--retry` flag is supported by curl - I believe it may be missing in some old versions. https://github.com/rust-lang/rustup/pull/2869 If we want to start using the flag, we'll want to do the same thing here. We'll also want to check into availability of the wget flag, too.

mistydemeo:
It appears that `--retry` first appeared in 7.12.3 [(ca 2004)](https://curl.se/ch/7.12.3.html), and `--retry-delay` first appeared in 7.14.0 [(ca 2005)](https://curl.se/ch/7.14.0.html). So... it's been around for a long time, but it does seem feasible that some of the more exotic systems people want to use could be missing it.

marandaneto:
+1 on this, getting very often `socket hang up` transient errors from GH releases, happy to take on this if we want to add this built in cargo-dist
right now the workaround is to patch the generated installer

suggestions:
- Adds four download attempts with exponential backoff: 1s, 2s, 4s.
 - Retries HTTP 408, 429, 5xx, socket resets, timeouts, and similar transient errors.
 - Does not retry permanent errors such as HTTP 404.

PR #2313, opened 2026-03-02
fix: double escape windows path in receipt

eegli:
Follow-up fix for #2227. Good news: We're halfway there with supporting Windows. However, on Git Bash/MSYS/MINGW environments, the `uv-receipt.json` now contains unescaped backslashes in the `install_prefix` field, which is invalid JSON:
```json
{"install_prefix":"C:\Users\username\.local\bin",...}
```

### Root Cause
The `convert_path_for_receipt()` function outputs escaped backslashes (`\\`), but these are later inserted via command substitution into a [double-quoted sed command](https://github.com/axodotdev/cargo-dist/blob/73dc6326196427e5d0ac5b6950208be05e8c3b03/cargo-dist/templates/installer/installer.sh.j2#L664):

```bash
RECEIPT=$(echo "$RECEIPT" | sed "s,AXO_INSTALL_PREFIX,$(convert_path_for_receipt "$_receipt_install_dir"),")
```

Inside double quotes, the shell processes `\\` as an escape sequence, which reduces each pair to a single backslash before `sed` executes. This results in unescaped backslashes. With this change, we output 4 backslashes per original backslash, which get reduced to two again. Essentially:

`sed "s,AXO_INSTALL_PREFIX,C:\\\\Users\\\\test"` -> Correct JSON

Refactored it a little so the whole receipt path conversion logic is not spread all over the place

### DIFF
--- a/cargo-dist/templates/installer/installer.sh.j2
+++ b/cargo-dist/templates/installer/installer.sh.j2
@@ -92,17 +92,6 @@ is_windows_posix() {
             ;;
     esac
 }
-# Convert POSIX path to Windows path
-convert_path_for_receipt() {
-    local _path="$1"
-    if is_windows_posix && check_cmd cygpath; then
-        # Use cygpath to convert path, then escape backslashes for JSON
-        # Need 4 backslashes to survive: command substitution + double quotes in sed command
-        cygpath -w "$_path" | sed 's/\\/\\\\\\\\/g'
-    else
-        echo "$_path"
-    fi
-}
 
 # The HOME reference to show in user output. If `$HOME` isn't set, we show the absolute path instead.
 get_home_expression() {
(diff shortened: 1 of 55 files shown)


eegli:
@mistydemeo can we merge this?

mistydemeo:
Yes, sorry for the delay reviewing this. Just need to check something before merging!

eegli:
No worries, let me know if I can help out @mistydemeo

eegli:
Can we merge?

mistydemeo:
Yeah, sorry - just need to get the time to test this before merging.

mistydemeo (approved):
Looks good, thanks.

Sorry again for the delay. Started a new job recently and it's kept me very busy.

PR #2318, opened 2026-03-03
feat(auditable): remove zigbuild limitation

mistydemeo:
We had this limitation in place since zigbuild couldn't be used to cross-compile while cargo-auditable was in use. That limitation no longer exists as of the new cargo-auditable 0.7.3.

https://github.com/rust-secure-code/cargo-auditable/releases/tag/v0.7.3

cc @Shnatsel


Shnatsel:
Due to the way `cargo zigbuild` works, the command should be `cargo auditable zigbuild` (which works) instead of `cargo zigbuild auditable` (which fails). It's worth double-checking that this is what `dist` runs.

mistydemeo:
Got it, I'll check what command order it uses.

PR #2330, opened 2026-03-16
Add failing test for npm installer authenticated proxy bug

tnkuehne:
### Context

We use `@j178/prek`, a Rust CLI distributed via npm through cargo-dist. The generated npm installer uses `axios` + `axios-proxy-builder` to download binaries. We discovered that authenticated proxies don't work. When a proxy (e.g. **Claude Code web sandbox**) requires credentials in the URL (`HTTPS_PROXY=http://user:pass@proxy:8080`), `axios-proxy-builder` silently drops them. It only passes `host` and `port` to `tunnel.httpsOverHttp()`, ignoring `proxyAuth` entirely.

We patched it locally:
```patch
// axios-proxy-builder dist/proxy-builder.js
-  const agent = tunnel.httpsOverHttp({
-    proxy: {
-      host: hostname,
-      port: parseInt(port),
-    },
-  });
+  const proxyConfig = {
+    host: hostname,
+    port: parseInt(port),
+  };
+  if (username && password) {
+    proxyConfig.proxyAuth = `${username}:${password}`;
+  }
+  const agent = tunnel.httpsOverHttp({
+    proxy: proxyConfig,
+  });
```

Patching this in every project isn't sustainable. `axios-proxy-builder` hasn't received an update in years, so fixing it upstream isn't realistic either. This is why  I am here. 

### What this PR does

Adds a failing regression test (`npm_installer_sends_proxy_auth`) that demonstrates the bug. The test:

1. Starts a local HTTP CONNECT proxy requiring Basic auth
2. Sets `HTTPS_PROXY` with credentials
3. Runs the actual installer code path (`binary.js` -> `install()` -> HTTP request)
4. Checks whether the proxy received the `Proxy-Authorization` header

The test fails because `axios-proxy-builder` drops the credentials from the CONNECT request.

### Suggested fix

Replace `axios` + `axios-proxy-builder` with native `fetch()` and `undici`'s `EnvHttpProxyAgent`, which handles authenticated proxies, `NO_PROXY`, and all standard proxy env vars correctly out of the box. This requires bumping the node engine minimum from `>=14` to `>=18` (for native `fetch()` and `Readable.fromWeb()`), but Node 14 has been EOL since April 2023, so that should be safe.

For reference, Node.js v22.21.0+ added [native proxy support for `fetch()`](https://nodejs.org/en/learn/http/enterprise-network-configuration) via `--use-env-proxy` / `NODE_USE_ENV_PROXY`, which would eliminate the `undici` dependency entirely. However, that requires Node 22.21+ and the end user has to opt in themselves. We can't set it from within the installer. `undici`'s `EnvHttpProxyAgent` works on Node 18+ with no user configuration required.

I have a working implementation on [this branch](https://github.com/tnkuehne/cargo-dist/tree/remove-broken-axios-dep). Happy to open a follow-up PR if this direction sounds good.

### DIFF
--- a/cargo-dist/tests/fixtures/test-npm-proxy-auth.js
+++ b/cargo-dist/tests/fixtures/test-npm-proxy-auth.js
@@ -1,0 +1,143 @@
+// Tests that the npm installer correctly forwards proxy authentication
+// credentials when HTTPS_PROXY contains username:password.
+//
+// This exercises the actual installer code path end-to-end:
+// 1. Starts a local HTTP CONNECT proxy requiring Basic auth
+// 2. Sets HTTPS_PROXY=http://testuser:testpass@127.0.0.1:<port>
+// 3. Patches package.json so binary.js has a real HTTPS download URL
+// 4. Requires binary.js (which sets up whatever proxy mechanism it uses)
+// 5. Calls install() — the same function npm postinstall runs
+// 6. Checks whether the proxy actually received Proxy-Authorization
+//
+// On the buggy code (axios-proxy-builder): FAILS — credentials are dropped
+// On the fixed code (undici EnvHttpProxyAgent): PASSES — credentials are sent
+
+"use strict";
+
+const http = require("http");
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+
+// Kill the test if it hangs
+setTimeout(() => {
+  console.error("FAIL: Test timed out");
+  process.exit(1);
+}, 15000);
+
+async function main() {
+  let authReceived = false;
+
+  // Start a minimal CONNECT proxy that inspects the Proxy-Authorization header
+  const proxy = http.createServer();
+  proxy.on("connect", (req, clientSocket) => {
+    const auth = req.headers["proxy-authorization"];
+    if (auth) {
+      const decoded = Buffer.from(
+        auth.replace("Basic ", ""),
+        "base64",
+      ).toString();
+      if (decoded === "testuser:testpass") {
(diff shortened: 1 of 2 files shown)


mistydemeo:
It also appears axios itself has builtin `proxy-from-env` support. A simpler patch here might be to just rip out `axios-proxy-builder` and see if it works with axios's automatic feature?

tnkuehne:
> It also appears axios itself has builtin `proxy-from-env` support. A simpler patch here might be to just rip out `axios-proxy-builder` and see if it works with axios's automatic feature?

Actually, I hadn't thought of this in the first place but I tested it now and removing axios-proxy-builder and letting axios handle proxies natively does fix the auth issue. **But** axios sends HTTPS requests as plain HTTP through the proxy instead of using CONNECT tunneling. Some proxies reject that and require proper CONNECT for HTTPS (see https://github.com/axios/axios/issues/5256). So we'd fix authenticated proxies but break strict CONNECT-only proxies. The undici approach handles both correctly.

rafaeelaudibert:
See #2350, I'm proposing `axios` to be dropped, which means I had to reimplement the proxy in this library. Can you check whether the implementation makes sense for you?

tnkuehne:
> See #2350, I'm proposing `axios` to be dropped, which means I had to reimplement the proxy in this library. Can you check whether the implementation makes sense for you?

yes perfect, thats actually the cleanest approach! Thanks

tnkuehne:
Closed by #2400

Issue #2224, opened 2025-12-15
Deps: Is console.table necessary?


20jasper:
[In the npm installer `binary.js` file, it references console.table](https://github.com/axodotdev/cargo-dist/blob/c53b7d91e9daf36bbedf8a37a60a3e9863142862/cargo-dist/templates/installer/npm/binary.js#L3)

This seems to be unused since there are no calls to console.table beyond this import 

Beyond that, the minimum node version is marked as 14, and [console.table has been shipped with node since node 10](https://nodejs.org/en/blog/release/v10.0.0)

PR #2350, opened 2026-04-01
chore(npm): replace axios with native node:http/node:https

rafaeelaudibert:
The only reason why @posthog depends on axios is because of `cargo-dist`. We don't want that anymore, so... here's a contribution!

Part of the e18e (https://e18e.dev) ecosystem cleanup effort. Remove axios, axios-proxy-builder, console.table, and rimraf from the npm installer template dependencies, replacing them with Node.js built-in modules. This reduces production dependencies from 42 packages to 1 (detect-libc).

See: https://e18e.dev/docs/replacements/fetch.html

- Replace axios HTTP client with `http`/`https` built-in modules
- Implement redirect following (up to 5 hops) inline
- Implement HTTP CONNECT proxy tunneling for HTTPS through HTTP proxies, with support for proxy authentication and NO_PROXY
- Replace rimraf.sync() with custom `rmSync` function - can use `fs.rmSync()` once we bump to  v14.14+
- Remove dead console.table import (cTable was never referenced)
- Regenerate npm-shrinkwrap.json (537 lines → 49 lines)
- Update test snapshots


cargo-dist/templates/installer/npm/binary-install.js
WarningImHack3r, line 338: You might also want to use the `error` function at the top of the file here

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -83,7 +215,7 @@ class Package {
     }
 
     if (existsSync(this.installDirectory)) {
```
WarningImHack3r: By the way, using exists before acting is often considered a bad practice, as some race condition can make it fail and crash the underlying fs call

Using rm within a try catch is generally better

WarningImHack3r (approved):

cargo-dist/templates/installer/npm/binary-install.js
rafaeelaudibert: Agreed, just ammended

rafaeelaudibert:
CC @mistydemeo since you seem to be the most active maintainer here :)

tnkuehne (approved):
New tests from #2330 pass on this one! LGTM

rafaeelaudibert:
Thanks for the confirmation @tnkuehne! @mistydemeo would love a review here!

20jasper:
Good work! I was just about to open a PR for the same 😄 

rimraf was 5MB of transitive deps removed alone (glob was the main offender)

This PR closes #2224

20jasper:
This PR reduces installed size by 56%, 11M -> 4.8M

rafaeelaudibert:
Nice, even better! 4.8M is cool 😎

mistydemeo:
Sorry for the delay getting to this; will review soon.

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -1,19 +1,151 @@
...
+    : process.env.HTTP_PROXY || process.env.http_proxy;
+
+  if (!proxyEnv) return null;
+
+  const proxyUrl = new URL(proxyEnv);
+  return {
+    hostname: proxyUrl.hostname,
+    port: proxyUrl.port || (proxyUrl.protocol === "https:" ? 443 : 80),
+    auth: proxyUrl.username
+      ? `${proxyUrl.username}:${proxyUrl.password}`
```
mistydemeo: Nit, but it looks like axios allowed password-only auth too - worth caring about preserving the behaviour here exactly? eg if password is set but not username it'd be treated as valid and create an auth string of `":PASSWORD"`.

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -1,19 +1,151 @@
...
+      );
+    } else {
+      doRequest();
+    }
+  });
+}
+
 class Package {
-  constructor(platform, name, url, filename, zipExt, binaries) {
+  constructor (platform, name, url, filename, zipExt, binaries) {
```
mistydemeo: Nit, mind keeping this consistent between the constructor and the other methods?
```suggestion
  constructor(platform, name, url, filename, zipExt, binaries) {
```

mistydemeo (commented):
Thanks again! Got a couple of minor comments, and will get @Gankra to take a look too. I'll also do some testing before we merge anything.

Do you mind bumping the minimum node in the `package.json` too? From what you're describing looks like the new minimum is 18.

rafaeelaudibert:
@mistydemeo changes have been addressed! I've also merged with master to make diff cleaner

WarningImHack3r:
@rafaeelaudibert @mistydemeo why does the minimum Node version needs to be bumped after dropping axios? All the Node modules used in the replacement do work with this old version AFAIK, don’t they? (don’t get me wrong, 18 is very reasonable and even too old itself, but dropping support for versions despite no technical limitations seems weird)

mistydemeo:
@WarningImHack3r The fetch API this PR uses was introduced in Node 18.

WarningImHack3r:
> The fetch API this PR uses

I reviewed the code a good half dozen times since it’s been created, and I don’t see any usage of `fetch` anywhere, as those have (specifically?) been avoided by using `node:http` and `node:https`?

Unless some precise APIs introduced recently have been used, this PR looks good with Node 14 support (or even less)

20jasper:
> > The fetch API this PR uses
> 
> I reviewed the code a good half dozen times since it’s been created, and I don’t see any usage of `fetch` anywhere, as those have (specifically?) been avoided by using `node:http` and `node:https`?
> 
> Unless some precise APIs introduced recently have been used, this PR looks good with Node 14 support (or even less)

This PR will require at least `v14.18.0` since the `node:` scheme was backported to node `v14.18.0`,

> v16.0.0, v14.18.0 | Added node: import support to require(...).
https://nodejs.org/api/modules.html#built-in-modules

WarningImHack3r:
> This PR will require at least `v14.18.0` since the `node:` scheme was backported to node `v14.18.0`,
> 
> > v16.0.0, v14.18.0 | Added node: import support to require(...).
> > [nodejs.org/api/modules.html#built-in-modules](https://nodejs.org/api/modules.html#built-in-modules)

This is for `require()`, `node:` prefix was added to ESM imports in 14.13.1 & 12.20 :D https://nodejs.org/api/esm.html#node-imports

mistydemeo:
> I reviewed the code a good half dozen times since it’s been created, and I don’t see any usage of fetch anywhere

Yep, I read the link from the PR description but missed that the API described there wasn't actually the one used in this PR 😅

rafaeelaudibert:
@mistydemeo @WarningImHack3r I apologize, that's my fault! I changed approach in the middle of it, but then forgot to actually think about it when implementing the node changes. I'll revert it back to the previous node version, this should make this easier to merge.

I'll be able to get to this next week.

WarningImHack3r:
Ok, so I just went ahead and did a manual checkup of every addition and their Node support so we're as fair as possible with the actual needs. Here are the results (once again, I don't care about old Node specifically at all, I even think we should drop everything non-supported, but it's one less breaking change for this PR to get approved and it doesn't close doors for no reason for the 3 people still using old Node):
- `rmSync`: [added in `14.14.0`](https://nodejs.org/docs/latest/api/fs.html#fsrmsyncpath-options)
- `node:` prefix in require (that I think we could get rid of if it introduces compatibility issues, it's not mandatory) (sorry @20jasper, I mentioned ESM but I forgot we're still in CJS here, my reply was irrelevant and you were right!): [added in `14.18.0`](https://nodejs.org/api/modules.html#built-in-modules)

> All the rest has been supported since basically forever (< v1)

So, as far as my research goes, we can either support `>=14.18` if we keep the PR as is, OR go as low as `>=14.14` if we replace `node:http(s)` in the `require` calls with `http(s)`!

cargo-dist/templates/installer/npm/binary-install.js
```diff
@@ -1,17 +1,180 @@
...
+      throw err;
+    }
+    for (const entry of readdirSync(target)) {
+      rmSync(join(target, entry), { recursive: true, force });
+    }
+    rmdirSync(target);
+  } else {
+    unlinkSync(target);
+  }
+}
```
rafaeelaudibert, line 45: @WarningImHack3r @mistydemeo this is my attempt to keep this at `>= 14` since I believe moving to `>= 14.14.0` might be a breaking change.

Let me know how you feel around this, I can see this being a problem, in which case we can just bump to `>= 14.14.0` instead.

rafaeelaudibert:
@WarningImHack3r @mistydemeo I've pushed some changes that mean we can go back to supporting Node `>= 14` like before the changes.

Of course, this means we have to ship some inline code that is native to Node in more modern versions, but I feel like this is a reasonable change to keep the backwards compatibility.

I would suggest the maintainers to consider releasing a new breaking version soon bumping to a newer Node version, there's a lot that can be changed in this repo to make it more modern while only supporting supported Node versions.

WarningImHack3r (approved):
LGTM!

gankra-oai (approved):

Gankra:
I'm mildly inclined to just bump the node version we require since the one you want is like, from 2020? But since you already did the legwork it seems fine.

Confirmed the proxy variables you checked are used right, and that we're not dropping anything. I can't speak super well to the precise node APIs but if it works, it works.

WarningImHack3r:
Kudos to the single misformatted file ruining the CI and preventing merging

mistydemeo:
I'll merge a squashed version with the appropriate style changes.

rafaeelaudibert:
Closed by #2400 :)

tysm!

PR #2352, opened 2026-04-02
feat: add cargo install fallback for dist prebuilt binary installation

### LINKED ISSUE #2351: Add cargo install fallback when prebuilt binary installer is unavailable
## Summary

The generated release workflow installs cargo-dist itself via a prebuilt binary
installer (`curl ... | sh`). On architectures where no prebuilt binary is
published — such as riscv64 — this step fails and blocks the entire workflow.

I'd like to propose adding a `cargo install cargo-dist --locked --version <version>`
fallback so the workflow can proceed even when a prebuilt binary is not available
for the runner's architecture.

## Background

The RISE Project recently announced the [RISE RISC-V Runners][rise] (2026-03-24),
a free, managed GitHub Actions runner service backed by Scaleway bare-metal
RISC-V servers. Any open-source project can now run native RISC-V CI by setting
`runs-on: ubuntu-24.04-riscv` — no emulation, no cross-compilation, no waitlist.

This makes it practical for Rust projects to build and distribute riscv64
binaries through GitHub Actions. However, cargo-dist's own installer only ships
prebuilt binaries for x86_64 and aarch64, so the "Install dist" step in the
generated `release.yml` fails on RISC-V runners.

[rise]: https://riseproject.dev/2026/03/24/announcing-the-rise-risc-v-runners-free-native-risc-v-ci-on-github/

## Proposed Change

In `DistInstallStrategy::dash()` / `powershell()`, append a fallback after the
installer command:

**bash (sh):**
```sh
curl --proto '=https' --tlsv1.2 -LsSf <url>/cargo-dist-installer.sh | sh \
  || cargo install cargo-dist --locked --version <version>
```

**PowerShell:**
```powershell
try { irm <url>/cargo-dist-installer.ps1 | iex }
catch { cargo install cargo-dist --locked --version <version> }
```

This is safe because:
- GitHub Actions bash steps run with `set -eo pipefail`; `-e` does not trigger
  inside an `||` chain, and `pipefail` ensures `curl` failures propagate.
- On architectures where the prebuilt binary exists, the installer succeeds and
  the fallback never runs.
- `cargo install --locked --version` is deterministic and uses the published
  crate's lockfile.

The trade-off is that the fallback compiles from source, which is slower. This is
acceptable because the alternative is a hard failure with no workaround.

ChanTsune:
When a prebuilt binary installer is unavailable for the target architecture (e.g. riscv), the generated CI workflow now falls back to `cargo install cargo-dist --locked --version <version>` instead of failing outright.

resolve #2351

### DIFF
--- a/cargo-dist/src/backend/ci/mod.rs
+++ b/cargo-dist/src/backend/ci/mod.rs
@@ -77,6 +77,8 @@ pub enum DistInstallStrategy {
         installer_url: String,
         /// the installer name, without `.sh` or `.ps1`
         installer_name: String,
+        /// the version to install, used for cargo install fallback
+        version: String,
     },
     /// Run `cargo install --git` (slow!)
     GitBranch {
@@ -97,6 +99,7 @@ impl DistInstallSettings<'_> {
             return DistInstallStrategy::Installer {
                 installer_url: url.as_str().to_owned(),
                 installer_name: "cargo-dist-installer".to_owned(),
+                version: self.version.to_string(),
             };
         }
 
@@ -116,6 +119,7 @@ impl DistInstallSettings<'_> {
             // to compute these values for packages we build *BUT* it's messy and not that important
             installer_url: format!("{BASE_DIST_FETCH_URL}/v{version}"),
             installer_name,
+            version: version.to_string(),
         }
     }
 }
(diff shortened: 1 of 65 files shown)


mistydemeo:
I'm not sure if this is the right approach. This will catch *any* `curl | sh` failure, even ones that are unrelated to the installer being missing for a given architecture. I think it would make more sense to vary which install command we inject based on whether a binary should be available or not.

gounthar:
Hi, I came at this from the other end: I added riscv64 release targets to atuin, they merged, and then had to revert a day later (atuinsh/atuin#3745) because the release workflow can't install dist on a riscv64 runner. So this thread is the blocker for us too, and I'd rather help finish it than open something competing.

@mistydemeo, your objection makes sense to me. `|| cargo install` catches a network blip or a bad checksum just as readily as a missing binary, and turns both into a silent twenty-minute source build.

The pieces for the narrower version look like they're already in place. `for_triple` receives the host triple of the job (`github.rs:331` passes `real_triple`), and the set of hosts dist publishes for is small and known: x86_64 and aarch64 for linux and macOS, x86_64 for Windows. So the strategy can pick `cargo install` only when the host isn't one of those, and keep the installer with no fallback everywhere else. A curl failure on x86_64 then still fails loudly, which I think is what you want.

It also stays out of the way of the existing snapshots. Since the generated command only changes for hosts that have no binary today, the 60-odd .snap files in this PR would go back to unchanged, and it'd add one new case covering a custom runner on a target dist doesn't ship.

@ChanTsune, this is your issue and your PR, so I don't want to step on it. Happy to push that version to your branch if you'd like, or open a separate PR that credits you, or just leave it with you if you were already partway there. Whichever you prefer.

PR #2370, opened 2026-04-20
CI: Hash-pin actions, other minor security chores

woodruffw:
This addresses a handful of CI best practices; namely it pins all action references to a hash (making them more hermetic) and explicitly drops permissions + eliminates credential persistence where possible.

(None of this is directly exploitable ATM, but is good from a defense-in-depth perspective.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,11 @@ updates:
     target-branch: main
     cooldown:
       default-days: 7
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "10:00"
+    target-branch: main
+    cooldown:
+      default-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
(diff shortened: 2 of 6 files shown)


woodruffw:
~~NB: I only pinned existing action versions, without updating them. Updating them (including hashes) could be done pretty easily with a new Dependabot group, but I left that out here since I didn't want to add on more long-tail maintenance work 🙂~~

Gankra (approved):
This is becoming the orthodoxy and is advisable to do, especially since we're part of people's release supply-chains.

Note for misty: the dependabot auto-updater is kinda cursed because it doesn't understand it needs to update dist-workspace.toml, so will always need a manual fixup.

mistydemeo:
If we're going to do this via dependabot, do we have a way we can hook into it to determine what versions it picked and update those in `dist-workspace.toml` so we don't have to manually fix it up every single time?

Gankra:
We do not currently have a way to do that, I'm not familiar enough with dependabot to say if that's easy.

We could also drop the dependabot stuff and deal with likely forgetting to update.

woodruffw:
Yeah, unfortunately there's no easy way to do that with Dependabot at the moment. There _might_ be a way to hack it into place with Renovate instead (since they have custom updaters/pattern matching), but that would involve adopting a third-party tool (and comes with other tradeoffs).

mistydemeo:
Sorry for the late response. That's fine then, maybe at some point we can swap to Renovate and automate it a little better but this will work for now.

mistydemeo (approved):

PR #2392, opened 2026-05-07
ps1: peek the inner exception on download failure

woodruffw:
This tries to make error messages from PowerShell's `DownloadFile` less opaque by peeking through the inner exceptions for the first non-empty one. I'm _pretty_ sure this is the right approach, but I'm by no means a PowerShell expert.

xref https://github.com/astral-sh/uv/issues/16309 as an example of error opacity.


mistydemeo (approved):
Haven't been able to test, but looks right to me.

woodruffw:
Thanks @mistydemeo!

PR #2396, opened 2026-05-17
Add Azure signing support and improve workflow for releases

### LINKED ISSUE #2395: support Azure Artifact Signing for binary signing
Hi, I am the maintainer of [rpx](https://github.com/scalerail-solutions/rpx) and would like to use Azure Artifact Signing as it's considerably cheaper than SSL.com.

Would it be possible to extend cargo-dist to support this provider?

I've made a pretty dirty monkey patch to use it but I think I'm not signing enough of things for it to work properly as evident by windows defender calling my powershell install script a trojan.

blyedev:
closes #2395


mistydemeo:
Thank you! I'm a little busy but I'll try to get to reviewing this soon.

blyedev:
Hey @mistydemeo, I wanted to follow up about the review of this. If you don't have the time that's perfectly fine as well, I can use a fork on rpx until you do, just don't want to duplicate work if it's going to get reviewed soon.

penberg:
We would love to use this for Turso packaging too. Right now we're working around this with hack in `release.yml`: https://github.com/tursodatabase/turso/pull/7618

blyedev:
Hi @mistydemeo, excuse the ping but shipping a workaround for this is pretty complex for me and this feature is becoming a blocker for scalerail-solutions/rpx#45 and a bigger rollout of my cli I'm trying to do. 

Do you think it'd be possible to compensate you for the time it would take to review this / ensuring it's in good enough shape to be merged?

mistydemeo:
Sure, and that's very kind of you to offer!

I would also appreciate any context you can offer on what I'd need to do to validate this is working - for example, is there a straightforward of me to test this without a paid Azure account?

book/src/supplychain-security/signing/windows.md
```diff
@@ -81,3 +81,41 @@ Want support for another vendor? [Drop us a line](mailto:hello@axo.dev) or [file
     If you used a sandbox (test) certificate, SmartScreen won't acknowledge the code signing at all, but you should be able to see a "Digital Signatures" entry in the properties of the file:
 
     ![](../../img/signing-properties.png)
+
+
+## Azure Artifact Signing Quickstart
+
+1. **Create Azure Artifact Signing resources**
+
+    Create an Artifact Signing account, complete identity validation, and create a certificate profile. For real releases, use a Public Trust profile. For testing, use a Public Trust Test profile.
```
mistydemeo: Let's also link to the upstream docs here so users have some extra context to go off. Maybe the quickstart guide would be useful? https://learn.microsoft.com/en-us/azure/artifact-signing/quickstart

book/src/reference/config.md
```diff
@@ -371,6 +372,43 @@ This setting takes one of two values:
...
+> certificate-profile-name = "my-certificate-profile"
+> ```
+
+If configured, dist signs Windows artifacts with Azure Artifact Signing. The selected certificate profile determines whether Azure uses a real Public Trust certificate or a Public Trust Test certificate.
+
+This setting cannot be used with `ssldotcom-windows-sign`.
+
+The table contains these fields:
+
+* `endpoint`: the regional Artifact Signing endpoint
```
mistydemeo, line 399: I think it would also be good to provide an upstream link to the documentation here to define what these values mean.

book/src/supplychain-security/signing/windows.md
```diff
@@ -81,3 +81,41 @@ Want support for another vendor? [Drop us a line](mailto:hello@axo.dev) or [file
...
+
+2. **Grant signing permissions**
+
+    The identity used by CI must have the `Artifact Signing Certificate Profile Signer` role on the certificate profile or signing account.
+
+3. **Add the Azure identity configuration to GitHub**
```
mistydemeo: As in step 5 for SSL.com, let's link to the GitHub secrets docs here to guide users: https://docs.github.com/en/actions/security-guides/encrypted-secrets

book/src/supplychain-security/signing/windows.md
```diff
@@ -81,3 +81,41 @@ Want support for another vendor? [Drop us a line](mailto:hello@axo.dev) or [file
...
+
+    - `AZURE_CLIENT_ID`: the client id of the app registration used for GitHub OIDC
+    - `AZURE_TENANT_ID`: the Azure tenant id
+    - `AZURE_SUBSCRIPTION_ID`: the Azure subscription id
+
+    These values identify the Azure identity; they are not authentication credentials. The generated workflow currently reads them from the GitHub `secrets` context, so add them as repository secrets or secrets on the `release` environment.
+
+    The generated GitHub workflow uses `azure/login` with OpenID Connect and runs the artifact build jobs in the `release` GitHub environment. Configure a federated credential for your repository on the app registration using an environment subject such as `repo:OWNER/REPO:environment:release`.
```
mistydemeo: Likewise, can we link to the docs for federated credentials here?

cargo-dist/src/sign/azure.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn sign(&self, file: &Utf8Path) -> DistResult<()> {
+        info!("Azure Artifact Signing {file}");
+
+        let AzureArtifactSigningConfig {
+            endpoint,
+            account_name,
+            certificate_profile_name,
+        } = &self.config;
+
+        // Match Azure's official GitHub Action integration, which wraps this module.
```
mistydemeo: Mind linking to the action this is matching? Is that https://github.com/Azure/artifact-signing-action?

cargo-dist/src/sign/mod.rs
```diff
@@ -26,16 +31,25 @@ impl Signing {
         host_target: &TripleNameRef,
         dist_dir: &Utf8Path,
         ssldotcom_windows_sign: Option<ProductionMode>,
+        azure_windows_sign: Option<AzureArtifactSigningConfig>,
         macos_sign: bool,
     ) -> DistResult<Self> {
+        if ssldotcom_windows_sign.is_some() && azure_windows_sign.is_some() {
+            return Err(crate::errors::DistError::MultipleWindowsSigningProviders);
```
mistydemeo, line 38: Thanks for handling this case!

mistydemeo (commented):
Thanks for including user documentation! Made this very easy to review. Just a few comments on the docs to make this a little more approachable, and consistent with the existing signing documentation.

I saw the comment that the signing has to run on x86_64 Windows - not even ARM64 Windows, huh? That's unfortunate, but I suppose it makes sense.

Thinking a little bit about handling this, it might be best to avoid overloading the global artifact job for this. If we have this one job that has to run on specific architectures, this may happen again in the future, and I'd like to avoid possible clashes in global runner if we can.

As an alternative, what if we introduced a new signing step that happens in between the global artifacts and hosting? We'd run it once per signing method being used, in between the global build artifacts and hosting; it would download the local *and* global artifacts, sign the ones that are necessary, and then reupload them with an updated manifest that defines the updated hashes.

book/src/reference/config.md
blyedev: 🫡

book/src/supplychain-security/signing/windows.md
blyedev: 🫡

book/src/supplychain-security/signing/windows.md
blyedev: 🫡

book/src/supplychain-security/signing/windows.md
blyedev: 🫡

cargo-dist/src/sign/azure.rs
blyedev: I linked to it both here and updated config.md to explain this in the docs too.

blyedev:
@mistydemeo Thanks for the suggestion. I implemented a dedicated x86_64 Windows signing job between global artifact creation and hosting. Executables and MSIs remain signed during local Windows builds, while the global job stays on Linux and the new job signs the PowerShell installer before replacing the global artifact bundle. Hosting is gated on successful signing.

I verified the approach end-to-end with Azure:

- [Workflow run](https://github.com/scalerail-solutions/rpx/actions/runs/32965471929)
- [Signed rpx release](https://github.com/scalerail-solutions/rpx/releases/tag/v2.0.0-azure.3)

The executable inside the Windows archive, MSI, and PowerShell installer all have the expected signatures. I included the same workflow path for SSL.com with test and production snapshot coverage, but could not verify it live because I do not have an SSL.com account.

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -315,6 +318,19 @@ jobs:
       - name: Install dependencies
         run: |
           ${{ matrix.packages_install }}
+      {{%- if azure_windows_sign %}}
```
mistydemeo, line 321: Since the global artifact job is also pulling down local artifacts, do those end up getting re-signed then? Do we actually need to do this here?

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -195,6 +195,9 @@ jobs:
       # - N "local" tasks that build each platform's binaries and platform-specific installers
       matrix: ${{ fromJson(needs.plan.outputs.val).ci.github.artifacts_matrix }}
     runs-on: ${{ matrix.runner }}
+    {{%- if azure_windows_sign %}}
+    environment: release
```
mistydemeo, line 199: I see that here and in the docs, you're using `environment: release` for the secrets. We're not using `environment:` for anything else in the other secrets; I think we should stick to the same secret scoping we're using for everything else.

cargo-dist/templates/ci/github/release.yml.j2
mistydemeo: Also realized running this in the local builds isn't going to work if there's an aarch64 Windows build. Probably best to save this for the separate signing job?

cargo-dist/templates/ci/github/release.yml.j2
blyedev: The environment is because of how we provision the azure identity. It's tied to the release environment. The variables we provision as secrets here technically should no issue being public it's just a layer of security by obscurity in case someone doesn't secure their identity properly. The secrets can also be global, it's really only the azure workload identity that's tied to release.

cargo-dist/templates/ci/github/release.yml.j2
blyedev: As per the resigning, nope, the dist sign the we use there explicitly filters only for the powershell installer that needs to be signed, everything else is kept as is.

As per the aarch64 builds. Yes I guess this would have to happen. The problem is that artifacts such as msi, are botch archives that contain the executable to be signed and would need to be signed themselves. A signing job would need to unpack sign and repack all the possible artifacts. This will be the case for any type of archive windows produces. I would need a few weeks to really understand the process enough to design this properly if we want to support windows aarch64 signing.

mistydemeo: > I would need a few weeks to really understand the process enough to design this properly if we want to support windows aarch64 signing.

Given the current design, that makes sense. We can ship a V1 that clearly advertises the limitation, and if someone does want aarch64 Windows signing we can work on improvements at a later date.

cargo-dist/templates/ci/github/release.yml.j2
mistydemeo: I see, so this is required on the Azure side. That's fine, but let's take the "secrets on the `release` environment." part out of the docs to avoid confusion and just let users put the secrets in one place.

cargo-dist/templates/ci/github/release.yml.j2
```diff
@@ -481,6 +497,83 @@ jobs:
...
+      - uses: {{{actions["actions/checkout"] | safe }}}
+        with:
+          persist-credentials: false
+          submodules: recursive
+      - name: Install dist
+        run: {{{ dist_install_for_signing.run }}}
+      - name: Fetch artifacts
+        uses: {{{actions["actions/download-artifact"] | safe }}}
+        with:
+          pattern: artifacts-*
```
mistydemeo: If we're only signing the global artifacts, this can also be optimized by only downloading them too.
```suggestion
          pattern: artifacts-build-global
```

mistydemeo:
I've run a build and it appears it worked! Thanks for the help on testing. Just to confirm, what's the best way to confirm that the assets were correctly signed? That's I think the only holdup, otherwise we're looking good.

> could not verify it live because I do not have an SSL.com account.

As it happens neither do I... the account I had access to belonged to Axo, haha. If this does break anything someone who uses that feature will let me know and it can be fixed then.

blyedev:
> I've run a build and it appears it worked! Thanks for the help on testing. Just to confirm, what's the best way to confirm that the assets were correctly signed?

I used [jsign](https://ebourg.github.io/jsign/) v8.0.0. The v7.5 on brew does not have the verify signature capability. I also had some issues compiling it but gpt sol easily fixed it. Otherwise if you have a windows machine you should be able to do so via [powershell](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.security/get-authenticodesignature?view=powershell-7.6)'s later versions.

> That's I think the only holdup, otherwise we're looking good.

Amazing, I'll push the changes you suggested. Thank you for all your efforts!

book/src/supplychain-security/signing/windows.md
```diff
@@ -4,6 +4,8 @@
 
 dist can automatically codesign Windows EXEs, PowerShell installers, and [MSIs](../../installers/msi.md) using SSL.com's [eSigner cloud signing service](https://www.ssl.com/esigner/) or Azure Artifact Signing.
 
+> Windows artifact signing currently supports only the `x86_64-pc-windows-msvc` target. Artifacts for `aarch64-pc-windows-msvc` are not signed.
```
mistydemeo: Rathe than "not signed", currently the build will fail altogether if the user specifies `aarch64-pc-windows-msvc`. Now that you mention it, probably cleaner to either error out before the build happens (at `dist init` time?) or skip signing at build time. Not crucial though.

blyedev: You're right thanks for the catch, I mixed it up with my monkey patch on rpx that fails open. Will fix!

blyedev: I pushed a draft of this fix, still have to check if works. Will do so this week.

cargo-dist/src/tasks.rs
```diff
@@ -951,6 +952,33 @@ pub(crate) struct DistGraphBuilder<'pkg_graph> {
...
+    if config.builds.ssldotcom_windows_sign.is_none()
+        && config.builds.azure_windows_sign.is_none()
+    {
+        return Ok(());
+    }
+
+    let targets = targets
+        .into_iter()
+        .filter(|target| {
+            target.is_windows() && target.as_explicit_ref() != TARGET_X64_WINDOWS
```
mistydemeo: This blocks `x86_64-pc-windows-gnu`, which is actually a valid signing target. For now, let's swap this to just check the architecture:

```suggestion
            target.is_windows() && !target.is_x86_64()
```

mistydemeo (commented):
Thanks! Gave this a test, and this looks good barring a couple minor issues:

* The Windows target check is actually stricter than it needs to be and is rejecting a target that gets a valid signed binary.
* `cargo fmt` needed

cargo-dist/src/tasks.rs
```diff
@@ -53,6 +53,7 @@ use std::collections::BTreeMap;
 use crate::backend::installer::{ExecutableZipFragment, HomebrewImpl};
 use crate::platform::targets::{
     TARGET_ARM64_LINUX_GNU, TARGET_ARM64_MAC, TARGET_X64_LINUX_GNU, TARGET_X64_MAC,
+    TARGET_X64_WINDOWS,
```
mistydemeo: Can safely pull this back out - sorry I didn't include that in the previous comment!
```suggestion
```

blyedev: Done, sorry it took me a second to get to this PR again.

mistydemeo (approved):

PR #2398, opened 2026-05-18
chore(npm): update installer dependencies

marandaneto:
## Problem

The npm installer shrinkwrap pinned transitive dependencies with known CVEs, including axios, follow-redirects, minimatch, and brace-expansion.

## Changes

- Bump the npm installer axios range to ^1.16.0.
- Regenerate npm-shrinkwrap.json so the pinned dependency tree uses patched versions.
- Update npm installer snapshots.

Patched pinned versions include axios 1.16.1, follow-redirects 1.16.0, minimatch 10.2.5, and brace-expansion 5.0.6.

## Verification

- npm audit --omit=dev
- cargo test -p cargo-dist --test integration-tests axolotlsay_simple_only -- --exact
- git diff --check

### DIFF
--- a/cargo-dist/templates/installer/npm-shrinkwrap.json
+++ b/cargo-dist/templates/installer/npm-shrinkwrap.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "hasInstallScript": true,
       "dependencies": {
-        "axios": "^1.13.5",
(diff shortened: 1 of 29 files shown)


mistydemeo:
Thanks! We ended up resolving this by removing axios as a dependency.

PR #2400, opened 2026-05-18
npm: #2350 without older npm fallbacks

mistydemeo:
As discussed in #2350, node 14.14 is already old enough we probably don't need the backport to get it back to node 14. I'm personally fine with marking node 14.14 as the new oldest version. This cherry-picks the first commit and makes a few other tweaks along with setting the minimum node as 14.14.

Closest #2350, #2359


mistydemeo:
cc @rafaeelaudibert, @WarningImHack3r

rafaeelaudibert (approved):
If we're happy with >=14.14 on a minor release, then :shipit: 

Thanks for your help getting this across the finish line!

rafaeelaudibert:
@mistydemeo let me know when a release is cut with this change, we're eager to run this :)

mistydemeo:
Will do!

PR #2405, opened 2026-05-22
fix: downgrade pnpm in CI

mistydemeo:
This test doesn't run with 11.2.2 due to a new error:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @mistydemeo/cargodisttest@file:axolotlsay-npm-package.tar.gz
```

There's nothing about it in the release notes, so I think this is a bug. https://github.com/pnpm/pnpm/releases/tag/v11.2.2


mistydemeo:
Looks like it was actually pnpm 11.0, not 11.2.2.

PR #2423, opened 2026-06-04
Install env scripts in receipts dir instead of bin dir

Gankra:
This attempts to address one of the original cardinal sins of cargo-dist: that in flat installs we create a sourceable script called `env` in the bin dir, where it can shadow `/usr/bin/env`.

This sin was originally forced by us not having any other directory we *could* put these files in, along with historical momentum of cargo calling the file `env`.

Since then we have introduced receipts, which means we have another dir where we can shove other random non-executable stuff.

* Renames `env` to `env.sh` because come on it's a shell script by any other name, even though it's only ever `source`d
* Makes us write `env.sh` and `env.fish` to the receipt dir
* Adds migration logic to attempt to remove the old `env` and `env.fish` files from the bin dir, if they look like they were made by us
* Adds migration logic to attempt to edit profiles to point to the new env files.

As always I am *profoundly* unconfident in my shell scripting (codex did most of it, but I don't even trust my reviews of shell code).


mistydemeo:
> that in flat installs we create a sourceable script called env in the bin dir, where it can shadow /usr/bin/env.

ftr, we install `env` as non-executable. Shells will typically ignore a non-executable file when evaluating the `PATH`, so a non-executable `env` earlier in the `PATH` will not take precedence over `/usr/bin/env`. I tested just now with `/bin/bash`, Homebrew-installed `bash` 5.2, `fish`, `zsh` and `dash`, and confirmed none of them are picking the cargo-dist installed `env` script over `/usr/bin/env`.

Just to check, is this in response to a specific issue that someone ran into?

Gankra:
Yes the specific issue is an annoying corner case that's "the user made their life hard, but also, because they were making kind of reasonable assumptions": they did some equivalent of of `chmod +x ~/bin/*`

ellie:
hey! just wanted to see if there's anything we can do to help with this. not entirely sure how it ended up as executable on our users systems, but it would be great if we can get it moved regardless! happy to help

mistydemeo:
Sorry for the delay reviewing this! I've simply been *very* busy lately, but hope to get to it soon.

ellie:
totally get it! btw if you opened a github sponsors for the org or yourself, we would totally sponsor you. dist is awesome and very much appreciated

cargo-dist/templates/installer/installer.sh.j2
```diff
@@ -808,6 +847,146 @@ add_install_dir_to_ci_path() {
...
+    local _legacy_env_script_path="$2"
+    local _legacy_env_script_path_expr="$3"
+    local _legacy_fish_env_script_path="$4"
+    local _legacy_fish_env_script_path_expr="$5"
+    local _env_script_path="$6"
+    local _env_script_path_expr="$7"
+    local _fish_env_script_path="$8"
+    local _fish_env_script_path_expr="$9"
+
+    if is_legacy_env_script_sh "$_install_dir_expr" "$_legacy_env_script_path"; then
```
mistydemeo, line 864: Nit: this confused me on first read because we didn't actually change the content of the shell scripts at all. This isn't really checking if it's a legacy env script, it's asking if the script at the specified path is a dist env script at all. Let's be clearer about what these are doing.

mistydemeo (commented):
Gave this a test, and confirmed it's working with both bash and fish.

Left one comment on something I'd like to clarify just a little so we remember the context in the future.

I did notice in testing that I ended up with repeated copies of the `.` source statement for the new path in my `.bashrc`. Only the sh version, not fish. I believe there's a specific sequence of events in which, if you have old *and* new versions at once in the same file, this makes it possible to end up with duplicate identical statements post-migration. (I didn't find the specific repro steps but I believe it involves installing old/new versions of the software in a particular order.) That's probably harmless, though we could dedupe while rewriting.

mistydemeo (approved):

PR #2479, opened 2026-08-17
Revert "ci: use stable versions for swatinem/build-cache"

mistydemeo:
This broke the integration tests.

This reverts commit d839dfc917c432c0102efd401d293ef006ec2516.


mistydemeo:
~~Fixed by changing the cache key.~~ This didn't fix it!

PR #2502, opened 2026-09-05
feat(config): added support for workspace.repository

mkatychev:
This feature is meant to handle the error one gets when a submodule monorepo has inconsistent package repository URLs:

```sh-session
$ dist build
  × Github CI support requires your crates to agree on the URL of your repository
  ╰─▶   ⚠ your workspace has inconsistent values for 'repository', refusing to select one:
        │   ~/rust/nushell-ci/nushell/Cargo.toml:
        │     https://github.com/nushell/nushell
        │   ~/rust/nushell-ci/nushell/crates/nu_plugin_formats/Cargo.toml:
        │     https://github.com/nushell/nushell/tree/main/crates/nu_plugin_formats
```

### DIFF
--- a/cargo-dist/src/host.rs
+++ b/cargo-dist/src/host.rs
@@ -211,7 +211,7 @@ pub(crate) fn select_hosting(
         .collect::<Vec<_>>();
 
     let raw_repository_url = if let Some(repo_url) = workspace_repository {
-        // Use workspace-level override if provided
+        // Use workspace override if present
         axoproject::RepositoryUrl::from_string(repo_url)
     } else {
         match workspaces.repository_url(Some(&package_list)) {
--- a/cargo-dist/src/host.rs
+++ b/cargo-dist/src/host.rs
@@ -85,7 +85,7 @@ impl<'a> DistGraphBuilder<'a> {
         } else {
             Some(hosting)
         };
-        self.inner.hosting = select_hosting(self.workspaces, announcing, hosting, Some(&ci))?;
+        self.inner.hosting = select_hosting(self.workspaces, announcing, hosting, Some(&ci), self.inner.config.repository.as_deref())?;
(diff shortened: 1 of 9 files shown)


cargo-dist/src/host.rs
```diff
@@ -189,6 +189,7 @@ pub(crate) fn select_hosting(
     announcing: &AnnouncementTag,
     hosting: Option<Vec<HostingStyle>>,
     ci: Option<&[CiStyle]>,
+    workspace_repository: Option<&str>,
```
mkatychev, line 192: I wasn't sure if this should be in the `path` field of `LocalRepo` so I made it a separate parameter:
https://github.com/axodotdev/cargo-dist/blob/783bc51dcd103628bda5aa5cf16c224601125160/axoproject/src/local_repo.rs#L10-L17

mistydemeo:
Oh I see, so you're using the `repository` field on child crates to refer to their path within the repo rather than just the repo itself? I hadn't considered using it that way but that feels reasonable. The Cargo book doesn't mandate a format here so this seems valid.

I don't mind letting `workspace.repository` serve as an override for this. I'll try to do a full review of this PR soon.

mkatychev:
@mistydemeo I also found that `packages` (and `version`) had to be under workspace rather than `dist` to work correctly:
https://github.com/mkatychev/nushell-ci/blob/cbf58bcf0be408634ba16805c7bf9e6bcb859d79/dist-workspace.toml#L23-L29

I can move this commit to a separate PR if needed:
https://github.com/axodotdev/cargo-dist/pull/2502/commits/d1cec6252168f2f30afd7b3f91756fd7962bf8b9

P.S. the fields were actually introduced in `0.30.3` not `0.29.0`: c948fc7 613c43a

PR #2507, opened 2026-09-12
ci: use self-repository syntax for jobs

### LINKED ISSUE #2456: Use GitHub's new "self-repository" syntax for reusable workflow calls
TL;DR: GitHub now supports `uses: $/...` as an alternative to `uses: ./...` for both composite actions and reusable workflows. The main advantage of this is that it strongly enforces that the call goes to the same repo, rather than to whatever workflow or action happens to be on the runner at that path at runtime. It also allows downstreams to enable GitHub's "require pinning" policy, whereas the older `./` syntax explicitly bypasses it.

I _think_ this should be a pretty simple change in `dist` itself: anywhere that a reusable workflow gets called with `./` should be substitutable to `$/` with no breakage 🙂 

Context: https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/

mistydemeo:
Fixes #2456.

cc @woodruffw


mistydemeo:
Confirmed working: https://github.com/mistydemeo/cargodisttest/actions/runs/34717301300/job/103617251757

woodruffw:
Thanks a ton @mistydemeo!

mistydemeo:
I have a few other things to prep, but should have a release with this out soon. Let me know if there's anything else you'd like in the next release.

PR #2511, opened 2026-09-20
Add arm64 Windows target to docs

marcoieni:
It's here since [0.29.0](https://axodotdev.github.io/cargo-dist/changelog/v0.29.0/).


ChanTsune (approved):

mistydemeo (approved):
Good catch, thanks.

